PR #364, opened 2015-01-06
(RE-3677) Add gemspec to packaging

haus:
In order to sanely reference packaging from other projects using bundler,
a gemspec is essential. It allows users to pin to a specific version or
range of versions, like ~> 0.1 instead of just the 0.1.0 tag. This will
also allow us to easily ship our packaging tools to rubygems if we want
to.


MosesMendoza:
it finally happened.. nice

PR #365, opened 2015-01-07
(maint) Use gsub instead of sub in gemspec

haus:
sub only replaces the first instance of a match, but for git describe,
we need to replace all instances of - to get a valid gem version. This
commit makes that change.


geoffnichols:
:+1:, pending Travis.

shrug:
also :+1:  pending travis

ScottGarman:
:+1:

PR #366, opened 2015-01-08
(RE-3702) Ensure uber_build taks uses retry_on_fail where appropriate.

waynr:


haus:
Seems sane overall. I wonder if retry_on_fail would fit better in Pkg::Util::Execution?

waynr:
@haus 
Based on the description at the top of the file containing `Pkg::Util::Execution` I didn't think it would be but I'm fine moving it if that's your preference.

haus:
@waynr I could be convinced either way. `retry_on_fail` doesn't always execute a shell call, but it does wrap execution of some sort (either in ruby or shell), so it seems like a reasonable fit to me.
I'd be fine if you wanted to update the description for Pkg::Util::Execution.

waynr:
@haus 
Okay, I've moved `retry_on_fail` to the `Pkg::Util::Execution` module.

haus:
I'd like to shave a bit on the name and placing of the parameter i guess. It probably shouldn't be called, timeout, because it isn't used as a timeout. It's more of an initial delay. So in it's current form, I would call it delayed_start or something.

I wonder if it makes sense to have the sleep per loop invocation or outside the loop. I could see cases where having the loop wait a bit before retrying would be helpful.

waynr:
@haus 
My intent was definitely for the timeout to occur within the loop I've just been juggling too many things this morning and neglected to actually think about what I was doing here.

haus:
@waynr cool. how do you feel about renaming timeout to delay to better reflect what it is doing in the loop?

waynr:
@haus 
That sounds good, I went ahead and renamed it.

tasks/jenkins_dynamic.rake
```diff
@@ -77,7 +77,10 @@ namespace :pl do
         #
         name = "#{Pkg::Config.project}-packaging-#{Pkg::Config.build_date}-#{Pkg::Config.ref}"
         packaging_job_url = "http://#{Pkg::Config.jenkins_build_host}/job/#{name}"
-        packaging_build_hash = Pkg::Util::Jenkins.poll_jenkins_job(packaging_job_url)
+
+        Pkg::Util::Execution.retry_on_fail(:times => 10, :delay => 1) do
```
haus: Will this correctly evaluate to a hash and not to 2 args?

tasks/00_utils.rake
```diff
@@ -188,26 +188,6 @@ def update_rpm_repo(dir)
 end
 alias :create_rpm_repo :update_rpm_repo
 
-# Loop a block up to the number of attempts given, exiting when we receive success
```
haus: Since this is being removed, it should also be deprecated in case other libraries are using it (enterprise-dist for example). There are examples in the bottom of this file.

tasks/jenkins_dynamic.rake
haus: Looks like it does the right thing on ruby 2 at least.

tasks/00_utils.rake
waynr: Whoops, forgot about this yesterday afternoon. Latest update on the patch has the deprecation!

waynr:
@haus 
I created a ticket for this patch just in case that helps it get merged a little quicker :)

haus:
:+1:

rji:
Looks good to me.

PR #367, opened 2015-01-09
Revert "(RE-3702) Ensure uber_build taks uses retry_on_fail where approp...

rji:
We're seeing uber_build fail in the following manner:

```
undefined local variable or method `packaging_build_hash' for main:Object
```

This reverts puppetlabs/packaging#366.


haus:
@waynr I think this is due to block scope in ruby. You probably want to initialize the various hashes in jenkins_dynamic outside the block, so they will be available later.

waynr:
There is an expression of horror on my face.

PR #368, opened 2015-01-10
(RE-3702) Fix block scope bug in uber_build task.

waynr:
@rji @haus


haus:
:+1:
@rji WEEEEEEEE

rji:
:+1: This patch seems to do the right thing now on Ruby 1.9.3, 2.0.0, and 2.1.1.

PR #369, opened 2015-01-15
(RE-3710) Update the msi name

melissaanne:
With the push for all in one agent packages, we're renaming our puppet
msi. This rename is so that the msi will be more in sync with the rest
of our puppet agent packages. This work is needed to build the windows
installer now that https://tickets.puppetlabs.com/browse/PUP-3844 has
been merged.


melissaanne:
Note, if we ever build with the old automation, this will break.

haus:
@melissaanne so for puppet 3.7.4, 3.7.5, ...?

melissaanne:
Totally broken. I guess I could dynamically assign the msi name...

haus:
@melissaanne if puppet version < 4...else...?

melissaanne:
@haus but that wont work until we get an actual 4.x tag...

haus:
Add pkg name to build_defaults hash?

melissaanne:
@haus https://github.com/puppetlabs/packaging/pull/370

PR #374, opened 2015-02-09
Generate MSI filename using `git describe`

joshcooper:
Previously,  MSIs were generated using `git describe` with the short sha removed, e.g. `3.7.4-1303`, which is not data that is currently passed through the AIO pipeline.

This commit changes the MSI to use the `git describe`, e.g. `3.7.4-1303-ge96f55b`. This  will simplify the process for installing MSIs during AIO acceptance testing.

### DIFF
--- a/templates/msi.xml.erb
+++ b/templates/msi.xml.erb
@@ -129,7 +129,7 @@ File.open(&quot;#{ENV[&apos;WORKSPACE&apos;]}/ondemand.yaml&quot;, &apos;w&apos;
       <command>call c:\puppetwinbuilder\setup_env.bat
 
 set PKG_NAME=<%= Pkg::Config.msi_name || 'puppet' %>
-set AGENT_VERSION_STRING=<%= Pkg::Config.version %>
+set AGENT_VERSION_STRING=<%= Pkg::Util::Version.git_describe_version %>
 set ARCH=%ARCH%
 set SUFFIX=-%ARCH%
 if &quot;%ARCH%&quot; == &quot;x86&quot; (


haus:
:+1:

ferventcoder:
:+1:

templates/msi.xml.erb
```diff
@@ -129,7 +129,7 @@ File.open(&quot;#{ENV[&apos;WORKSPACE&apos;]}/ondemand.yaml&quot;, &apos;w&apos;
       <command>call c:\puppetwinbuilder\setup_env.bat
 
 set PKG_NAME=<%= Pkg::Config.msi_name || 'puppet' %>
-set AGENT_VERSION_STRING=<%= Pkg::Config.version %>
+set AGENT_VERSION_STRING=<%= Pkg::Util::Version.git_describe_version %>
```
joshcooper: Hmm this isn't right, [`Puppet::Util::Version.git_describe_version`](https://github.com/puppetlabs/packaging/blob/master/lib/packaging/util/version.rb#L81) returns an array of values. Perhaps we want [`get_dash_version`](https://github.com/puppetlabs/packaging/blob/master/lib/packaging/util/version.rb#L108) instead? /cc @melissa @haus

melissa: It seems I mislead you. get_dash_version should work, but `@version` is getting set at https://github.com/puppetlabs/packaging/blob/master/lib/packaging/config.rb#L174 so we might just be able to use `@version`, but I feel like I'd need confirmation on that

melissa: some experimentation:
```
irb(main):005:0> puts Pkg::Config.version
3.7.4-1273
=> nil
irb(main):006:0> Pkg::Util::Version.get_dash_version
=> "3.7.4-1273"
irb(main):007:0> Pkg::Util::Version.get_dot_version
=> "3.7.4.1273"
irb(main):008:0> Pkg::Util::Version.git_describe_version
=> ["3.7.4", "1273"]
irb(main):009:0> Pkg::Util::Version.git_describe
=> "3.7.4-1273-gc08076d"
```

It looks like you're actually going to want `Pkg::Util::Version.git_describe`

joshcooper: @melissa I'll give it a go. Also confirmed the PR wasn't working:
`set AGENT_VERSION_STRING=["3.7.4", "1310"]`

joshcooper: It looks like my github PR from earlier created a branch in upstream, and when I deleted the branch it closed the PR... Submitting a new PR.

PR #375, opened 2015-02-09
Include short sha in puppet-agent MSI filename

joshcooper:
Previously, MSIs were generated using `git describe` with the short sha
removed, e.g. `3.7.4-1303`. This version identifier is not passed
through the AIO pipeline, and is not necessarily unique enough to
identify the puppet-agent package to be installed.

This commit changes the xml template for creating jenkins jobs so that
the resulting puppet-agent MSI contains the short sha, e.g.
`3.7.4-1303-ge96f55b`. This  will simplify the process for installing
MSIs during AIO acceptance testing.

### DIFF
--- a/templates/msi.xml.erb
+++ b/templates/msi.xml.erb
@@ -129,7 +129,7 @@ File.open(&quot;#{ENV[&apos;WORKSPACE&apos;]}/ondemand.yaml&quot;, &apos;w&apos;
       <command>call c:\puppetwinbuilder\setup_env.bat
 
 set PKG_NAME=<%= Pkg::Config.msi_name || 'puppet' %>
-set AGENT_VERSION_STRING=<%= Pkg::Config.version %>
+set AGENT_VERSION_STRING=<%= Pkg::Config.version %>-<%= Pkg::Config.short_ref %>
 set ARCH=%ARCH%
 set SUFFIX=-%ARCH%
 if &quot;%ARCH%&quot; == &quot;x86&quot; (


joshcooper:
Supercedes PR #374

templates/msi.xml.erb
```diff
@@ -129,7 +129,7 @@ File.open(&quot;#{ENV[&apos;WORKSPACE&apos;]}/ondemand.yaml&quot;, &apos;w&apos;
       <command>call c:\puppetwinbuilder\setup_env.bat
 
 set PKG_NAME=<%= Pkg::Config.msi_name || 'puppet' %>
-set AGENT_VERSION_STRING=<%= Pkg::Config.version %>
+set AGENT_VERSION_STRING=<%= Pkg::Config.version %>-<%= Pkg::Config.short_ref %>
```
ferventcoder: Well that is certainly a way to handle it.

So we are not using number of commits past a tag then?

ferventcoder: I'm assuming that short_ref is just the sha. 

`3.7.4-gc08076d`

and not something like what would come back from `Pkg::Util::Version.git_describe` as @melissa mentioned in https://github.com/puppetlabs/packaging/pull/374#discussion_r24366680 ? 

`3.7.4-1273-gc08076d`

ferventcoder: I can totally understand if  `Pkg::Config.short_ref` has that information available, I would just state that the variable is poorly named then. :)

haus:
```
rake -s pl:print_build_param[:short_ref] 2> /dev/null
97e8dd6
```

```
rake -s pl:print_build_param[:version] 2> /dev/null
3.7.3-1425
```

```
irb(main):007:0> Pkg::Util::Version.git_describe
=> "3.7.3-1425-g97e8dd6"
```

ferventcoder: Ah! Well then version... is... strange to me. :)

ferventcoder: :+1:

joshcooper:
Continuing discussion from https://github.com/puppetlabs/packaging/pull/374#discussion_r24366680

> It looks like you're actually going to want Pkg::Util::Version.git_describe

@melissa thanks for researching the various options. If we go the `Pkg::Util::Version.git_describe` route, then the resulting string will be the raw result from `git describe`, and won't contain the special logic that happens in `git_describe_version` for release candidates. If we don't care, then `git_describe` sounds good.

If we do care, then we could do:

```
set AGENT_VERSION_STRING=<%= Pkg::Config.version %>-<%= Pkg::Config.short_ref %>
```

And it appears to work correctly for x86 and x64:

http://jenkins-release.delivery.puppetlabs.net/job/puppet-msi-2015-02-09-13-39-58-306b3a7/ARCH=x64,label=win-build/1/console generated http://builds.puppetlabs.lan/puppet/306b3a701983777e045d601cedcf6e5a5b7ce9ac/artifacts/windows/puppet-agent-3.7.4-1310-306b3a7-x64.msi

and 

http://jenkins-release.delivery.puppetlabs.net/job/puppet-msi-2015-02-09-13-39-58-306b3a7/ARCH=x86,label=win-build/1/console generated http://builds.puppetlabs.lan/puppet/306b3a701983777e045d601cedcf6e5a5b7ce9ac/artifacts/windows/puppet-agent-3.7.4-1310-306b3a7-x86.msi

Ultimately, I just want the MSI's to match how we're doing things on other platforms (where it makes sense). And if I understand vanagon correctly, it looks like rpm packages are using `Pkg::Config.version` in the erb template: https://github.com/puppetlabs/vanagon/blob/master/templates/rpm/project.spec.erb#L14?

ferventcoder:
@joshcooper works for me. :)

joshcooper:
Originally I was thinking AIO for Windows was down the road aways. But it is closer than I was expecting, see https://github.com/puppetlabs/puppet-agent/pull/32. I'm going to close this PR, and we can focus on getting puppet-agent generating MSIs with the correct name.

PR #376, opened 2015-02-13
(RE-3836) Update packaging automation to allow promotion of projects built with vanagon

geoffnichols:
With the advent of vanagon-built AIO packages, some of the assumptions
in our package promotion automation don't apply to projects built with vanagon.

To support promotion of AIO packages, this changeset does the following: 

* Adds a "vanagon_project" parameter available, so we can set 
"vanagon_project: TRUE" in a project's build_defaults.yaml.

* Since there is not a tarball created during an AIO build, 
we don't need to try to sign it.

* Since we can no longer assume that packages promoted
into PE will use the 'pe-' prefix, we remove the 'pe-'
prefix from the 'pe:ship_rpm' and 'pe:ship_deb' tasks.

* Updates pe_ship tasks to support use of
'rpm_targets' and 'deb_targets' in place of 'cows' and 'final_mocks'
for Vanagon projects.


tasks/pe_ship.rake
```diff
@@ -100,23 +100,23 @@ if Pkg::Config.build_pe
           #
           # I am not proud of this. MM - 1/3/2014.
```
shrug: This is my favorite comment ever.

tasks/deb.rake
```diff
@@ -101,6 +101,24 @@ task :build_deb, :deb_command, :cow do |t, args|
...
 end
 
+def cow_to_codename_arch(cow)
+  codename, arch = /^base-(.*)-(.*)\.cow$/.match(cow).captures
+  [codename, arch]
+end
+
+def deb_build_targets
+  if Pkg::Config.vanagon_project
+    targets = Pkg::Config.deb_targets.split(' ')
```
haus: The assignment isn't needed here as targets is never used.

tasks/deb.rake
```diff
@@ -101,6 +101,24 @@ task :build_deb, :deb_command, :cow do |t, args|
   puts "Finished building in: #{bench}"
 end
 
+def cow_to_codename_arch(cow)
+  codename, arch = /^base-(.*)-(.*)\.cow$/.match(cow).captures
```
haus: You should be able to just have `/^base-(.*)-(.*)\.cow$/.match(cow).captures` here and get rid of the rest. It will implicitly return the array.

tasks/deb.rake
```diff
@@ -101,6 +101,24 @@ task :build_deb, :deb_command, :cow do |t, args|
...
+  [codename, arch]
+end
+
+def deb_build_targets
+  if Pkg::Config.vanagon_project
+    targets = Pkg::Config.deb_targets.split(' ')
+  else
+    cows = Pkg::Config.cows.split(' ')
```
haus: I *think* this could be rewritten as:

```
Pkg::Config.cows.split(' ').map do |cow|
  codename, arch = cow_to_codename_arch(cow)
  "{codename}-#{arch}"
end
```

tasks/mock.rake
```diff
@@ -134,6 +134,20 @@ def mock_el_ver(mock_config)
...
+def rpm_family_and_version
+  if Pkg::Config.vanagon_project
+    family_and_version = Pkg::Config.rpm_targets.split(' ').map do |target|
+      rpm_el_family, rpm_el_version, arch = target.split('-')
+      "#{rpm_el_family}-#{rpm_el_version}"
+    end
+  else
+    family_and_version = Pkg::Config.final_mocks.split.map { |mock| "#{mock_el_family(mock)}-#{mock_el_ver(mock) }" }
+  end
+  family_and_version
```
haus: family_and_version is already returned in each of the if branches above, it isn't required here (and is inconsistent with other methods :smile: )

haus:
Looks totally functional. I had a few style and simplification suggestions, but no blockers.

geoffnichols:
Updated and smoke tested promotion of code-management into my branch of enterprise-dist.

tasks/mock.rake
```diff
@@ -134,6 +134,19 @@ def mock_el_ver(mock_config)
   version
 end
 
+# Return the RPM family and version for a Vanagon or Packaging repo built project.
+def rpm_family_and_version
+  if Pkg::Config.vanagon_project
+    family_and_version = Pkg::Config.rpm_targets.split(' ').map do |target|
```
haus: Nitpick. no need for assignment here or in the else of this.

haus:
I'm :+1: here even with my nit-pick.

tasks/mock.rake
geoffnichols: Apologies, I missed that on my update. I'll fix it.

geoffnichols:
Updated to remove unneeded assignment from ```rpm_family_and_version``` method.

haus:
:+1: still. this is going to be so amazing.

PR #377, opened 2015-02-13
(maint) Update sign_repos for deb to handle failure

haus:
Previously, if there were no deb repos to sign, the nightly repo ship
would explode, because we try to call each on a possibly nil object.
This commit updates the method to helpfully let us know that there were
no deb repos to sign if dists is nil.


justinstoller:
Seems like a sane conditional. Should there be an equivalent wrapper around rpms?

haus:
rpms don't have this problem for various reasons.

shrug:
:+1:  pending travis

haus:
travis loves me.

haus:
@shrug travis loves me

shrug:
@haus We all do

PR #378, opened 2015-02-13
(RE-4008) Make repo tasks more fail friendly

haus:
Previously, if rpms and debs hadn't both been built, various repo
related methods would explode in failure when trying to wget, retreive
or ship repo/repo configs. This breaks a lot of automation when
developing a new project when only one platform may be being built. This
commit rectifies this by warning and returning early from these methods
instead of failing completely.

### DIFF
--- a/lib/packaging/deb/repo.rb
+++ b/lib/packaging/deb/repo.rb
@@ -27,10 +27,11 @@ module Pkg::Deb::Repo
 
       # First test if the directory even exists
       #
-      wget_results = Pkg::Util::Execution.ex("#{wget} --spider -r -l 1 --no-parent #{repo_base} 2>&1")
-
-      unless Pkg::Util::Execution.success?
-        fail "No debian repos available for #{Pkg::Config.project} at #{Pkg::Config.ref}."
+      begin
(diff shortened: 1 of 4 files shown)


stahnma:
:+1: pending travis.

haus:
@stahnma travis loves me

MosesMendoza:
This is one of those tough calls to make. Retrieving the repo configs is a required part of several acceptance pipelines. As such, if we can't retrieve the configs during testing, it's most helpful to fail because of it, rather than let the tests continue and have them fail cryptically later. I actually think this PR re-introduces the original behavior of packaging, which we changed to make failure more helpful. With this change, it might be helpful to communicate to QE that packaging isn't going to fail anymore, so that pipelines that use this code can write their own error handling for the case that the repo configs don't exist. I personally would opt for failing hard being configurable (e.g., a "strict" mode) which we could turn on during testing, but leave off by default for development.

haus:
@MosesMendoza good point. I opened https://tickets.puppetlabs.com/browse/RE-4009 to track adding some sort of flag to gate this behavior.

PR #379, opened 2015-02-13
(RE-4008) Update conditional to also check for non-existence

haus:
Previously if the directory did not exist, it could not be considered
empty. That is fine for the implementation of empty_dir?, but isn't good
for the ship_repo_configs method. This commit updates the bailout
section to also return early if the directory does not exist at all.


shrug:
:+1:

ScottGarman:
:+1:  assuming Travis goes green

PR #381, opened 2015-02-19
(maint) On ship don't set symlinks immutable

Iristyle:
- Windows packages are being built in the AIO pipeline for installation
   during acceptance runs.  However, the pipeline was passing SHA and
   ARCH, but not the `git describe` used to tag the file version.

   Other platforms use repo configs and are unaffected by this, but
   to be able to find the packages by convention, it was decided that
   symlinks should be used so that an MSI could always be found at

   `puppet-agent/<SHA>/artifacts/windows/puppet-agent-<ARCH>.msi`

   This symlink is generated at build time and points to a file that
   contains the complete version string generated from `git describe`.

   However, this generated a new problem with the :ship rake task,
   since it is now trying to `chattr +i` on symlinks.  The error that
   occurs is:

   `chattr: Operation not supported while reading flags on <PATH>`

   To fix this, filter the file set passed to remote_set_immutable to
   exclude symlinks.

   More details on the addition of symlinks can be found in:
   https://github.com/puppetlabs/puppet/commit/0976d62

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -170,7 +170,7 @@ namespace :pl do
       end
 
       # If we just shipped a tagged version, we want to make it immutable
-      files = Dir.glob("#{local_dir}/**/*").select { |f| File.file?(f) }.map do |file|
+      files = Dir.glob("#{local_dir}/**/*").select { |f| File.file?(f) && !File.symlink?(f) }.map do |file|
         "#{artifact_dir}/#{file.sub(/^#{local_dir}\//, '')}"
       end
       remote_set_immutable(Pkg::Config.distribution_server, files)


Iristyle:
@haus / @melissa please verify for accuracy prior to merge.

You can see the relevant stack trace for a build failure at:
https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Integrated%20Acceptance/job/platform_aio-component_pkg-build_master/BUILD_TARGET=win-x64,SLAVE_LABEL=beaker/117/consoleFull

/cc @joshcooper

joshcooper:
@Iristyle I talked this over with @haus and @melissa and we think just creating a hard link instead will do the trick (as we use hardlinks that are latter `chattr`'ed elsewhere). I've opened https://github.com/puppetlabs/puppet-agent/pull/63, and will close this.

PR #382, opened 2015-02-20
(RE-3836) Move rpm and deb _build_targets methods to Pkg::Config

geoffnichols:
This commit moves 'deb_build_targets' and 'rpm_build_targets' methods
to the Pkg::Config class to so they can be accessed from
enterprise-dist tasks.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -273,6 +273,42 @@ module Pkg
           end
         end
       end
+
+
+      # This method is duplicated from enterprise-dist so we can access it here.
+      def cow_to_codename_arch(cow)
+        /^base-(.*)-(.*)\.cow$/.match(cow).captures
+      end
+
+      # This method is duplicated from enterprise-dist so we can access it here.
+      def mock_to_dist_version_arch(mock)
+        # We care about matching against two patterns here:
+        # pupent-3.4-el5-i386 <= old style with PE_VER baked into the mock name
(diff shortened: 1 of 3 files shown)


lib/packaging/config.rb
```diff
@@ -273,6 +273,28 @@ def issue_deprecations
           end
         end
       end
+
+      def deb_build_targets
+        if self.vanagon_project
+          self.deb_targets.split(' ')
+        else
+          self.cows.split(' ').map do |cow|
+            codename, arch = cow_to_codename_arch(cow)
```
haus: This is gonna get weird. cow_to_codename_arch is defined in deb.rake, so it won't be available here.

lib/packaging/config.rb
```diff
@@ -273,6 +273,28 @@ def issue_deprecations
...
+          end
+        end
+      end
+
+      def rpm_build_targets
+        if self.vanagon_project
+          self.rpm_targets.split(' ')
+        else
+          self.final_mocks.split(' ').map do |mock|
+            platform, version, arch = mock_to_dist_version_arch(mock)
```
haus: Similar to above, this is defined in rpm.rake.

lib/packaging/config.rb
geoffnichols: Ah, looks like I added cow_to_codename_arch in 4bd8c50. I'll move it into this class.

stahnma:
eww :+1:

PR #383, opened 2015-02-26
(re-4028) Fix bad regex in pe_ship that breaks Cumulus

shrug:
Currently, the regex used to ship PE components breaks on the
CumulusLinux-2.2 codename due to an optimistic split('-')

This updates with a regex (the same one used before the above change
was implemented) that matches correctly.


shrug:
Broekn in https://github.com/puppetlabs/packaging/commit/4bd8c50a6907e546f23409e1890fe16522e20547

haus:
:+1:

stahnma:
LGTM, pending Travis.

PR #384, opened 2015-03-04
(maint) Fix signing for sles 10 and 12 packages

shrug:
This enables signing for sles-10 and sles-12 packages, aligning with sles-11 and EL.

### DIFF
--- a/tasks/pe_sign.rake
+++ b/tasks/pe_sign.rake
@@ -10,14 +10,14 @@ if Pkg::Config.build_pe
       # Find x86_64 noarch rpms that have been created as hard links and remove them
       rm_r Dir["pkg/pe/rpm/*-*-x86_64/*.noarch.rpm"]
       # We'll sign the remaining noarch
-      sign_dists = 'el4', 'el5', 'el6', 'el7', 'sles11'
+      sign_dists = ['el4', 'el5', 'el6', 'el7', 'sles10', 'sles11', 'sles12']
       ['i386', 'x86_64'].each do |arch|
         sign_dists.each do |dist|
           family = dist[/[a-z]+/]
           version = dist[/[0-9]+/]
           rpm_stagedir        = "pkg/pe/rpm/#{family}-#{version}-#{arch}/*.rpm"
           srpm_stagedir       = "pkg/pe/rpm/#{family}-#{version}-srpms/*.rpm"
-          if family == 'el' and version >= '6'
+          if (family == 'el' and version >= '6') || (family == 'sles' and version >= '12')
             modern_rpms += FileList[rpm_stagedir] + FileList[srpm_stagedir]
           else
             old_rpms += FileList[rpm_stagedir] + FileList[srpm_stagedir]


tasks/pe_sign.rake
```diff
@@ -8,16 +8,16 @@ if Pkg::Config.build_pe
       old_rpms = FileList.new
       modern_rpms = FileList.new
       # Find x86_64 noarch rpms that have been created as hard links and remove them
-      rm_r Dir["pkg/pe/rpm/*-*-x86_64/*.noarch.rpm"]
+      rm_r Dir["pkg/**/rpm/*-*-x86_64/*.noarch.rpm"]
       # We'll sign the remaining noarch
-      sign_dists = 'el4', 'el5', 'el6', 'el7', 'sles11'
+      sign_dists = 'el4', 'el5', 'el6', 'el7', 'sles11', 'sles12'
```
stahnma: What about SLES10?

tasks/pe_sign.rake
```diff
@@ -41,7 +41,7 @@ if Pkg::Config.build_pe
     desc "Sign all debian changes files staged in pkg/pe"
     task :sign_deb_changes do
       load_keychain if Pkg::Util::Tool.find_tool('keychain')
-      sign_deb_changes("pkg/pe/deb/*/*.changes") unless Dir["pkg/pe/deb/*/*.changes"].empty?
+      sign_deb_changes("pkg/**/deb/*/*.changes") unless Dir["pkg/pe/deb/*/*.changes"].empty?
```
haus: Missing a glob change in the unless.

tasks/pe_sign.rake
stahnma: Should we add Fedora here too, since Vanagon can have that?

haus: No. this is pe specific. just for promotions.

haus: sles10 seems like a good thing to add though.

shrug: No to fedora, this is the pe_sign task that gets called by the promote job. signing foss packages will come elsewehere.

I'll add sles10

haus:
@shrug have you tried it?

shrug:
Totally doesn't work. Because I'm dumb.

The PE/FOSS pathing is entirely different, not just the same structure under pe/

haus:
@shrug ah yes, of course

shrug:
This seems to work. It will need to be updated when/if we ship vanagon source packages or vanagon noarch packages

stahnma:
I guess I'm +1

haus:
don't merge this, it may be being superseded by a PR to enterprise-dist for the vanagon fixes. the sles10 and sles12 fixes should still stick around.

shrug:
Updated! Now just adds singing for sles-10 and sles-12, the other work was moved to enterprise-dist

tasks/pe_sign.rake
```diff
@@ -10,14 +10,14 @@ if Pkg::Config.build_pe
       # Find x86_64 noarch rpms that have been created as hard links and remove them
       rm_r Dir["pkg/pe/rpm/*-*-x86_64/*.noarch.rpm"]
       # We'll sign the remaining noarch
-      sign_dists = 'el4', 'el5', 'el6', 'el7', 'sles11'
+      sign_dists = 'el4', 'el5', 'el6', 'el7', 'sles10', 'sles11', 'sles12'
```
haus: nitpick: can we please add array brackets here to make it explicit what we're doing?

shrug: Yes, definitely

haus:
:+1: for real

PR #385, opened 2015-03-06
(RE-4045) Add support for alternate repositories

haus:
Previously the packaging repo was limited to shipping to main/production
or devel, based on the versioning scheme of the project being shipped.
This commit extends that to allow the rpeo to be overriden for apt or
yum repos using a param in the yaml file. apt_repo_name and
yum_repo_name are both available to override the defaults.

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -3,6 +3,7 @@
 module Pkg::Params
   BUILD_PARAMS = [:apt_host,
                   :apt_repo_path,
+                  :apt_repo_name,
                   :apt_repo_url,
                   :author,
                   :benchmark,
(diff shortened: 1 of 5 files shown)


haus:
2 example builds. one with apt_repo_name and yum_repo_name set, one without...

With aardwolf set: http://builds.puppetlabs.lan/facter/7b8f16ce120e2996ce095fe1fcfd59adfa16ebb8

Without anything set: http://builds.puppetlabs.lan/facter/f5011f7e2635baa383f021547d64f5cd1c77ab72

haus:
Style issues. just pushed up a fix. specs passed.

stahnma:
:+1:

PR #386, opened 2015-03-06
(RE-2255) Check for cows and mocks before doing stuff.

waynr:


haus:
This looks sane enough. My only question here is do we want to have the rake tasks warn and continue when there are no cows/mocks, or do we want the jenkins job to *not invoke the particular task* when the cows/mocks are empty?

waynr:
@haus I don't have any opinion on that matter (except that I have already implemented it this way and i obviously would prefer not to redo it a different way if there is no advantage).

I did briefly consider the possibility of creating a separate repo job for deb and rpm--but that seems like more work than it's worth.

haus:
My main concern is just if you call a rake task to do a thing, it should probably fail if it can't do it. But maybe I'm crazy.

haus:
:+1: once the style fixes are in

waynr:
@haus updated to move the logic to the ERB templates. This did what I want it to do.

shrug:
:+1:  pending travis

PR #387, opened 2015-03-23
(RE-4139) Update ship taks to use repo_command if set 

haus:
If the user has used the new Pkg::Config.apt_repo_command or
Pkg::Config.yum_repo_command, we want to prefer those commands over the
legacy defaults. This PR adds logic to check if the new param is
set, and if so to invoke it, and otherwise to fall back to the old
behavior. The freight rake task is renamed to update_apt_repo, and the
uber_ship is updated to use it instead of freight.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -18,20 +18,45 @@ namespace :pl do
     # to various target yum and apt repositories based on their specific type
     # e.g., final vs devel vs PE vs FOSS packages
 
-    desc "Update remote rpm repodata on #{Pkg::Config.yum_host}"
+    desc "Update remote yum repository on '#{Pkg::Config.yum_host}'"
     task :update_yum_repo do
-      STDOUT.puts "Really run remote repo update on #{Pkg::Config.yum_host}? [y,n]"
+      yum_whitelist = {
+        :yum_repo_name => "__REPO_NAME__",
+        :yum_repo_path => "__REPO_PATH__",
+        :yum_host      => "__REPO_HOST__",
+        :gpg_key       => "__GPG_KEY__",
+      }
+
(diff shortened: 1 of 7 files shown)


haus:
IT BLENDS!

haus:
I tested this by adding the following sections to a build_defaults file.

```
apt_host: weth.delivery.puppetlabs.net
apt_repo_path: /opt/tools/freight/apt/
apt_repo_command: |
  keychain -k mine;
  eval $(/usr/bin/keychain -q --agents gpg --eval __GPG_KEY__);
  export GPG_TTY=$(tty);
  sudo -E freight-cache -c /etc/freight.conf.d/community.conf;
  keychain -k mine;
```

tasks/ship.rake
```diff
@@ -18,20 +18,45 @@ namespace :pl do
     # to various target yum and apt repositories based on their specific type
     # e.g., final vs devel vs PE vs FOSS packages
 
-    desc "Update remote rpm repodata on #{Pkg::Config.yum_host}"
+    desc "Update remote yum repository on '#{Pkg::Config.yum_host}'"
     task :update_yum_repo do
-      STDOUT.puts "Really run remote repo update on #{Pkg::Config.yum_host}? [y,n]"
+      yum_whitelist = {
+        :yum_repo_name => "__REPO_NAME__",
+        :yum_repo_path => "__REPO_PATH__",
```
melissa: I can't remember, would there be a reason to want to have `:yum_repo_url`, like we do for apt?

haus: No idea. It doesn't even look like we ever use `:apt_repo_url`, so I didn't bother duplicating it in params.rb.

melissa:
:+1: I'm in. Pending the one question I had for you.

PR #389, opened 2015-04-03
(RE-4261) Add build_tar config parameter

haus:
For vanagon projects, tarballs will not be produced. This breaks the
uber_ship, which expects a tar to be there. This commit addresses this
by adding a build_tar parameter which defaults to true, which can be
used to toggle the tar ship/build. The TAR environment variable can also
be used to set this on the command line.

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -16,6 +16,7 @@ module Pkg::Params
                   :build_ips,
                   :build_msi,
                   :build_pe,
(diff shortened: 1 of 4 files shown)


haus:
I've validated that this does the right thing against puppet and the puppetlabs-release-pc1 packages.

stahnma:
:+1:

stahnma:
Rob also reviewed.

PR #391, opened 2015-04-10
(maint) Only sign tar if the tar has been built

haus:
A previous commit made building the tar optional, but neglected to make
signing the tar optional as well. This commit updates the sign_all task
to only sign the tarball if Pkg::Config.build_tar is true.


melissa:
:+1:

shrug:
:+1:  pending travis

PR #393, opened 2015-04-13
(RE-3293) Treat SNAPSHOT builds like RCs for versioning

haus:
Previously, ezbake builds would have a SNAPSHOT build that was treated
like any other. This had the consequence of making the 'final' release
for a given version series as less than the SNAPSHOT builds, which means
it would be unpromotable (also not immediately upgradable from the
snapshot version). This commit updates the versioning in packaging to
treat SNAPSHOT builds as if they were RC builds. This means that
SNAPSHOT versions will have a '0.1' prepended to them (0.7.0.SNAPSHOT...
becomes 0.7.0-0.1SNAPSHOT...), which makes the final version higher than
the snapshot version for packaging purposes.

### DIFF
--- a/lib/packaging/util/version.rb
+++ b/lib/packaging/util/version.rb
@@ -146,6 +146,9 @@ module Pkg::Util::Version
         # Grab the rc number
         rc_num = dash.match(/rc(\d+)/)[1]
         ver = dash.sub(/-?rc[0-9]+/, "-0.#{Pkg::Config.release}rc#{rc_num}").gsub(/(rc[0-9]+)-(\d+)?-?/, '\1.\2')
+      elsif dash.include?("SNAPSHOT")
+        # Insert -0.#{release} between the version and the SNAPSHOT string
+        ver = dash.sub(/^(.*)\.(SNAPSHOT\..*)$/, "\\1-0.#{Pkg::Config.release}\\2")
       else
         ver = dash.gsub('-', '.') + "-#{Pkg::Config.release}"
       end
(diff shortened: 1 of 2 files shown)


haus:
Test build using a SNAPSHOT tag: http://jenkins-release.delivery.puppetlabs.net/job/pe-puppetdb-packaging-2015-04-13-13-13-10-1.3.4.SNAPSHOT.2015.04.13.T1309/

lib/packaging/util/version.rb
```diff
@@ -215,9 +218,16 @@ def is_final?
...
     # '0.7.0-rc1-63'
     # '0.7.0-rc1-63-dirty'
+    # '0.7.0.SNAPSHOT.2015.03.25T0146'
     def is_rc?
-      return TRUE if get_dash_version =~ /^\d+\.\d+\.\d+-*rc\d+/
-      return FALSE
+      case get_dash_version
+      when /^\d+\.\d+\.\d+-*rc\d+/
+        TRUE
+      when /^\d+\.\d+\.\d+\.SNAPSHOT\.\d{4}\.\d{2}\.\d{2}T\d{4}/
```
shrug: My mind shudders at what it would look like of this regex was merged with the one above

shrug:
:+1:  if Travis agrees

lib/packaging/util/version.rb
haus: I didn't even want to go there. I could just do something like `/SNAPSHOT/` if it seems overly specific.

shrug: Nope, I'm good with keeping the cases separate, it's clear and easy to read. 

I think the specificity is reasonable? It may come back to bite us at some point, but I kind of like being fairly specific about these formats.

stahnma:
:+1:

PR #394, opened 2015-04-13
(maint) Allow shipping of deb and rpm packages in a vanagon project even...

melissa:
... if 'cows' and 'final_mocks' aren't set


haus:
@melissa is there anything else that needs updating?

melissa:
@haus I think so. The rest of the changes I made were specific to building packages

haus:
:+1:

PR #395, opened 2015-04-13
(RE-4374) Create x86_64 dir during rpm signing

haus:
When building a vanagon noarch package, the x86_64 directory won't
always be created. When hardlinking, this will mean that there won't be
hardlinks of the noarch packages in the x86_64 directory. This commit
avoids that problem by ensuring that the x86_64 directory is created
before the hardlink is attempted.


shrug:
:+1:

PR #399, opened 2015-05-19
(maint) Add build_for parameter to distinguish server and agent packages

shrug:
On PE going forward, we need to be able to be able to distinguish between
server and agent packages, as different checks may be applied. In particular
we currently ensure that noarch packages exist in both the i386 and x86_64
repos for promotion. This assumption breaks with future versions of PE, where
we need to only ship x86_64 server packages.

This adds a "build_for" parameter (defaulting to 'agent', which will behave as
it does currently) to allow the automation to be able to distinguish between agent
and server packages, and apply logic accordingly.

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -12,6 +12,7 @@ module Pkg::Params
                   :build_defaults,
                   :build_dmg,
                   :build_doc,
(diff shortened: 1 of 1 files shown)


shrug:
Aaand I'm dumb, this is completely unecessary

PR #400, opened 2015-05-20
(maint) Add instructions for smoke testing gems

Iristyle:
- Until automation is added around gems, manually smoke test them

### DIFF
--- a/tasks/tickets.rake
+++ b/tasks/tickets.rake
@@ -159,6 +159,7 @@ Procedure may vary by project and point in the release cycle. Ask around.
 In general this should happen on a variety of platforms, i.e. one or two each of kind of package we create (i.e., gem, dmg, msi, deb, rpm, etc).
 
 For Puppet, our acceptance suite now tests service scripts, and on debian, a passenger master.  Manual smoke testing can therefore be limited to other package formats than deb and rpm.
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Couldn't this be (doc) and skip ci?

tasks/tickets.rake
```diff
@@ -172,6 +173,7 @@ IMPORTANT: Please edit the description of this ticket and remove "Example:" belo
 Example:
 Smoke test platforms:
   * pick some platforms such as
+  * gem - select one Linux for the universal gem, Windows with x64 platform-specific gem, and Windows with x86 platform-specific gem
```
haus: no darwin?

joshcooper: Ah yeah, there's a `universal-darwin` gem of facter, because it has a dependency on `CFPropertyList`, e.g. http://builds.puppetlabs.lan/facter/2.4.4/artifacts/ And since puppet no longer has `CFPropertyList` dependencies there's not a `universal-darwin` puppet gem.

MikaelSmith: We also don't acceptance test Apple packages as far as I can tell.

haus:
@Iristyle ping

tasks/tickets.rake
joshcooper: we do now :tada:

PR #401, opened 2015-05-27
(RE-4666) All repsimple calls should pass --verbose

haus:
In debugging failures, having more verbose output will be useful. This
commit adds the verbose flag to the repsimple calls to enable that.


shrug:
:+1:

PR #402, opened 2015-05-27
(RE-4666) Print output from repsimple call

haus:
Previously the ship task would collect the output from repsimple and
scan it for errors, but it would not be printed to the screen. This
commit ensures that both stdout and stderr from the command are printed
to the screen for debugging purposes.


shrug:
:+1:

PR #403, opened 2015-05-28
(RE-4669) Fail promotion on reprepro errors

haus:
Previously the promotion would continue rolling even on errors creating
repos using reprepro. This commit addresses that by checking the output
of the repsimple call for errors and failing if any are found. This will
be coupled with a change to repsimple itself to exit non-zero on
reprepro errors.


melissa:
:+1:

PR #405, opened 2015-06-01
(RE-4595) Add AIX rpms to the list that packaging repo signs

stahnma:
When we added nxos and SLES support for rpm signing, AIX was missed.
This fixes that by adding AIX to the older_rpms signing patterns.


haus:
gross

haus:
:+1:

shrug:
:+1:

melissa:
Were we signing AIX packages before? If not, we could always just create a new list of unsigned packages, and then it won't fail

stahnma:
Seems like we should sign them.

PR #406, opened 2015-06-02
(RE-4681) Enable shipping of builds using version instead of ref

haus:
For some builds it will be preferable to have autosigned builds shipped
to directories according to their version, and not their ref. This makes
them easier to identify based on an already installed package and will
enable internal ships of puppet-agent to proceed. This commit adds a
new parameter to the deploy_signed_repo task. If versioning is set to
ref, the current behavior is maintained. If it is set to version, then
the repo_configs are updated and the repos are shipped to a directory
named after the dotted version. The deploy_nightly_repos shim is also
updated here to pass ref to maintain current behavior.

This PR also includes a substantial refactoring of the tasks in the nightly_repos.rake file.

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -10,8 +10,9 @@ namespace :pl do
     # This is to enable the work in CPR-52 to support nightly repos. For this
     # work we'll have signed repos for each package of a build.
     #
-    task :remote_sign_nightly_repos => "pl:fetch" do
-      target = "nightly_repos/"
+    task :remote_sign_repos, [:target_prefix] => "pl:fetch" do |t, args|
+      target_prefix = args.target_prefix or fail ":target_prefix is a required argument for #{t}"
+      target = "#{target_prefix}_repos/"
       signing_server = Pkg::Config.signing_server
       # Sign the repos please
       Pkg::Util::File.empty_dir?("repos") and fail "There were no repos found in repos/. Maybe something in the pipeline failed?"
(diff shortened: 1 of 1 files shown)


tasks/nightly_repos.rake
```diff
@@ -10,8 +10,9 @@ namespace :pl do
     # This is to enable the work in CPR-52 to support nightly repos. For this
     # work we'll have signed repos for each package of a build.
     #
-    task :remote_sign_nightly_repos => "pl:fetch" do
-      target = "nightly_repos/"
+    task :remote_sign_repos, [:target_prefix] => "pl:fetch" do |t, args|
+      target_prefix = args.target_prefix or fail ":target_prefix is a required argument for #{t}"
+      target = "#{target_prefix}_repos/"
```
melissa: What else are we going to ship here? `nightly_repo`, yeah, but what else?

haus: agent builds. really these are all signed_repos, but i didn't want to break any existing assumptions.

tasks/nightly_repos.rake
```diff
@@ -20,51 +21,59 @@ namespace :pl do
...
+        Pkg::Util::Net.rsync_to("#{target_prefix}_repos/", Pkg::Config.distribution_server, target_dir)
       end
     end
 
-    task :deploy_nightly_repos, [:target_host, :target_basedir] => ["clean", "pl:fetch"] do |t, args|
+    task :deploy_signed_repos, [:target_host, :target_basedir, :target_prefix, :versioning] => ["clean", "pl:fetch"] do |t, args|
       target_host = args.target_host or fail ":target_host is a required argument to #{t}"
       target_basedir = args.target_basedir or fail ":target_basedir is a required argument to #{t}"
+      target_prefix = args.target_prefix or fail ":target_prefix is a required argument for #{t}"
+      versioning = args.versioning or fail ":target_prefix is a required argument for #{t}"
```
melissa: copypasta

haus: yay!

melissa:
:+1: pending copypasta fix

haus:
@melissa pushed a fix

ScottGarman:
:+1: I can merge when Travis is done.

PR #409, opened 2015-06-08
(RE-4684) Add automation to ship repos to s3

haus:
This PR includes some refactoring to the deployed repos tasks and also 
adds two new tasks to nightly_repos.rake. The first is a
task that ships an already signed repo to a given s3 bucket and the
second is a wrapper task that first signs repos for a given project and
SHA and then ships it to s3.
A new method is also added to Pkg::Util::Net to enable syncing out of repos to s3.


haus:
just pushed a test fix

ScottGarman:
:+1:

PR #412, opened 2015-06-17
(RE-2542) shipping staging automation for osx

heathseals:
This commit updates the automation to allow shipping of dmgs for vanagon projects ( vanagon_project set to true ) and when build_dmg is not set. It also updates the source directory for the rsync to be pkg/apple/ instead of pkg/apple/*.dmg.  This adds support for apple projects that have a repo defined, such as PC1.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -110,10 +110,14 @@ namespace :pl do
 
   desc "ship apple dmg to #{Pkg::Config.yum_host}"
   task :ship_dmg => 'pl:fetch' do
-    Pkg::Util::Execution.retry_on_fail(:times => 3) do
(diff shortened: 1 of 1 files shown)


stahnma:
:+1:

haus:
@heathseals what happens if you uber_ship or ship_dmgs when there are none?

heathseals:
@haus just tested and nothing appears to happen if there are no dmgs ( no failures, but also no indication that the apple pkg directory was empty ).  does it need to be verbose?

haus:
@heathseals verbose would be cool. mostly i wanted to make sure that if we need to ship puppet-agent 1.2.0 without osx packages that the ship wouldn't explode.

heathseals:
@haus updated.

PR #413, opened 2015-06-19
(RE-2541) Update signing automation for OSX

heathseals:
This commit adds signing automation for Vanagon projects.  It takes
unsigned artifacts created by Vanagon, ships them to a signing sever,
unpacks the dmg, signs the packages contained within the dmg, repacks
the dmg, and brings the newly signed artifacts back.


stahnma:
I assume the signing server in this case must be an OS X machine? (I'm not very familiar with the workflows here)

heathseals:
@stahnma join the club!  but yeah, we have one dedicated signing server running osx.

tasks/sign.rake
```diff
@@ -53,6 +53,38 @@ def sign_ips(fmri, repo_uri)
   %x(pkgsign -s #{repo_uri}  -k #{Pkg::Config.privatekey_pem} -c #{Pkg::Config.certificate_pem} -i #{Pkg::Config.ips_inter_cert} #{fmri})
 end
 
+def sign_osx
```
haus: This method belongs in lib somewhere. Possibly lib/packaging/osx.rb or something like that.

tasks/sign.rake
```diff
@@ -148,6 +180,11 @@ namespace :pl do
     end
   end
 
+  desc "Sign OSX packages"
+  task :sign_osx do
```
haus: This should probably depend on the pl:fetch task to ensure that if we have signing data in build-data, that it gets loaded.

haus: Although the other tasks don't depend on pl:fetch. @MosesMendoza do you know why that is?

heathseals:
@haus updated.  i went ahead and added the fetch task as it may make sense to have some of those vars added in the build-data instead of environment vars.

lib/packaging/osx.rb
```diff
@@ -0,0 +1,32 @@
+module Pkg::OSX
+  def self.sign_osx
```
haus: most of the modules in packaging have a line after module that looks like

`class << self`

To avoid needing to prefix all of the method names with self.

haus:
@heathseals have you tested this?

heathseals:
@haus i've tested locally using my own signing cert -- it works.  As long as the proper environment variables are set, it should work fine on remote signing hosts as well.

stahnma:
:+1:

haus:
I'll try to give this a test against our signing machine later today.

heathseals:
@haus just took a look at our signing host and it appears to be using slightly different environment variables.  calling it a day, will take a look at how best to handle that later.

haus:
@heathseals sounds good. i'll still try it out and i'll leave any ideas i come up with.

heathseals:
@haus this runs as expected on the remote signer ( as long as the environment vars are setup properly ).  in enterprise-dist signing,  they certname / keychain password are pulled in via env vars on the remote host and the keychain and signing server are set via constants.  i assume the constants should be moved into build-data?

haus:
@heathseals i think you're also going to want to add the sign_osx task to the sign_all task in tasks/sign.rake. I'm trying that out right now and will have a PR for your PR if it works out.

lib/packaging/osx.rb
```diff
@@ -0,0 +1,34 @@
...
+      signing_keychain    = Pkg::Config.signing_keychain
+      signing_keychain_pw = Pkg::Config.signing_keychain_pw
+      signing_cert        = Pkg::Config.signing_cert
+      signing_server      = Pkg::Config.signing_server
+      signing_ssh_key     = Pkg::Config.signing_ssh_key
+
+      ssh_host_string = "-i #{signing_ssh_key} #{ENV['USER']}@#{signing_server}"
+      rsync_host_string = "-e 'ssh -i #{signing_ssh_key}' #{ENV['USER']}@#{signing_server}"
+
+      work_dir  = "/tmp/#{random_string(8)}"
```
haus: random_string got removed in commit f8edae58d0d084b028cc1b0552a1421a8466336d, as nothing used it. rand_string is still available in tasks/00_utils.rake.

lib/packaging/osx.rb
```diff
@@ -0,0 +1,34 @@
+module Pkg::OSX
+  class << self
+    def sign_osx
+      signing_keychain    = Pkg::Config.signing_keychain
+      signing_keychain_pw = Pkg::Config.signing_keychain_pw
+      signing_cert        = Pkg::Config.signing_cert
+      signing_server      = Pkg::Config.signing_server
+      signing_ssh_key     = Pkg::Config.signing_ssh_key
+
+      ssh_host_string = "-i #{signing_ssh_key} #{ENV['USER']}@#{signing_server}"
```
haus: If running these commands as myself, and not via a jenkins job, which would happen in an uber_ship, I won't always have an ssh key on the remote host to use (this is in the context of the sign_all, which runs on saturn itself). It would probably be good to support using ssh-agent here, and making the signing_ssh_key optional.

haus:
opened a PR to your PR at https://github.com/heathseals/packaging/pull/1

haus:
@heathseals we should also document internally what is required to make this work (host specific setup on the signing server for example).

haus:
But, with the attached PR and the right stuff on the signing server, this worked fine in the uber_ship.

mckern:
Is there a test package signed with this that I could drop into a Pooler VM and test?

haus:
@mckern yea, i will hipchat you details

mckern:
Dig it.

heathseals:
@haus.  updated.

haus:
@heathseals did you see my comment about making the ssh_key optional?

heathseals:
@haus oops.  missed that.  i'll take a look.

heathseals:
@haus updated.  the signing_ssh_key is now optional, though i'm not really happy with it.  i'm certain there's a more elegant way to handle it.

haus:
@heathseals seems sane enough for now

haus:
I'm rolling another test. Now would be an excellent time for other eyes to be on the PR.

lib/packaging/osx.rb
```diff
@@ -0,0 +1,36 @@
...
+      rsync_host_string = "-e 'ssh #{use_identity}' #{ENV['USER']}@#{signing_server}"
+
+      work_dir  = "/tmp/#{rand_string}"
+      mount     = File.join(work_dir, "mount")
+      signed    = File.join(work_dir, "signed")
+      output    = File.join("pkg", "apple", "#{Pkg::Config.yum_repo_name}")
+      Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, "mkdir -p #{work_dir}/{mount,signed}")
+      dmgs = Dir.glob("pkg/apple/**/*.dmg")
+      Pkg::Util::Net.rsync_to(dmgs.join(" "), rsync_host_string, work_dir)
+      Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, %Q[for dmg in #{dmgs.map { |d| File.basename(d, ".dmg") }.join(" ")}; do
```
haus: I think we could clean this up some and make it into its own helper method. I don't know if that needs to happen for the initial work or not.

lib/packaging/osx.rb
```diff
@@ -0,0 +1,36 @@
...
+      use_identity = "-i #{signing_ssh_key}" unless signing_ssh_key.nil?
+
+      ssh_host_string = "#{use_identity} #{ENV['USER']}@#{signing_server}"
+      rsync_host_string = "-e 'ssh #{use_identity}' #{ENV['USER']}@#{signing_server}"
+
+      work_dir  = "/tmp/#{rand_string}"
+      mount     = File.join(work_dir, "mount")
+      signed    = File.join(work_dir, "signed")
+      output    = File.join("pkg", "apple", "#{Pkg::Config.yum_repo_name}")
+      Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, "mkdir -p #{work_dir}/{mount,signed}")
```
melissa: This might be a little easier as `mkdir -p #{mount} #{signed}`

melissa:
:+1: there are a few little nitpicks left, but I think overall functionality looks reasonable. I like Haus' idea of breaking up some of the logic a little into more manageable pieces, but overall I don't have any serious concerns

lib/packaging/osx.rb
```diff
@@ -0,0 +1,36 @@
...
+      dmgs = Dir.glob("pkg/apple/**/*.dmg")
+      Pkg::Util::Net.rsync_to(dmgs.join(" "), rsync_host_string, work_dir)
+      Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, %Q[for dmg in #{dmgs.map { |d| File.basename(d, ".dmg") }.join(" ")}; do
+        /usr/bin/hdiutil attach #{work_dir}/$dmg.dmg -mountpoint #{mount} -nobrowse -quiet ;
+        /usr/bin/security -v unlock-keychain -p "$SIGNING_KEYCHAIN_PW" "$SIGNING_KEYCHAIN" ;
```
haus: This is hardcoded to use the env vars on the remote system and is totally ignoring the `Pkg::Config` values. I think we should pick one.

haus:
@heathseals SIGNING_KEYCHAIN, SIGNING_KEYCHAIN_PW and SIGNING_CERT are all still using local ENV vars loaded on the remote system, bypassing the Pkg::Config values entirely. That should get worked out before we merge this in.

Using the Pkg::Config values will be more flexible, and still allows us to set a Pkg::Config value to `$SIGNING_KEYCHAIN` for use in our infra.

heathseals:
@haus yeah, overlooked those.   updated.   i set defaults for those ENV vars we most likely expect to be loaded on the remote system.  is this ok or should we avoid setting defaults?

lib/packaging/osx.rb
heathseals: @melissa updated.

melissa:
:+1: still

lib/packaging/osx.rb
```diff
@@ -0,0 +1,36 @@
+module Pkg::OSX
+  class << self
+    def sign_osx
+      signing_keychain    = Pkg::Config.signing_keychain || "$SIGNING_KEYCHAIN"
```
haus: Defaults should go in the DEFAULTS array in `lib/packaging/config/params.rb` instead of here.

heathseals:
@haus updated.  moved the env var defaults into the config params.

haus:
cool. i'm in.

haus:
@heathseals so one last thing i guess.

signing server is already used in packaging to refer to the signing server that can rpm sign/gpg sign packages. this signing server is specific to osx, and we can't use 2 values for the same setting in packaging. for puppet-agent we would need to pick one. this should get broken out to a new setting, which is awful but necessary.

heathseals:
@haus do we want to prefix all the settings with osx_ or something similar?

haus:
@heathseals that works for me, and makes sense here

heathseals:
@haus updated.

haus:
:+1: for reals

PR #414, opened 2015-06-19
(RE-4880) Sign rpm repos during signed repo creation

haus:
Previously our rpm repos were unsigned. However, this breaks sles repo
usage that expects the repomd files to be signed. This commit begins
signing all repomd.xml files in a new sign_repos method, which is called
in the nightly_repos.rake tasks. Non-sles platforms happily ignore
signed repomd.xml files, so this is safe to apply everywhere.

This PR also includes moving the gpg and keychain methods into Pkg::Util::Gpg.

### DIFF
--- a/lib/packaging/util/gpg.rb
+++ b/lib/packaging/util/gpg.rb
@@ -1,0 +1,49 @@
+module Pkg::Util::Gpg
+  class << self
+
+    def keychain
+      if @keychain.nil?
+        @keychain = Pkg::Util::Tool.find_tool('keychain')
+      else
+        @keychain
+      end
+    end
+
+    def load_keychain
+      unless @keychain_loaded
+        unless ENV['RPM_GPG_AGENT']
+          kill_keychain
+          start_keychain
+        end
+        @keychain_loaded = TRUE
+      end
+    end
+
+    def kill_keychain
+      if keychain
+        Pkg::Util::Execution.ex("#{keychain} -k mine")
+      end
+    end
+
+    def start_keychain
+      if keychain
+        keychain_output = Pkg::Util::Execution.ex("#{keychain} -q --agents gpg --eval #{Pkg::Config.gpg_key}").chomp
+        new_env = keychain_output.match(/GPG_AGENT_INFO=([^;]*)/)
+        ENV["GPG_AGENT_INFO"] = new_env[1]
+      else
+        fail "Keychain is not installed, it is required to autosign using gpg."
+      end
+    end
+
+    def sign_file(file)
+      gpg ||= Pkg::Util::Tool.find_tool('gpg')
+
+      if gpg
+        use_tty = "--no-tty --use-agent" if ENV['RPM_GPG_AGENT']
+        Pkg::Util::Execution.ex("#{gpg} #{use_tty} --armor --detach-sign -u #{Pkg::Config.gpg_key} #{file}")
+      else
+        fail "No gpg available. Cannot sign #{file}."
+      end
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


tasks/nightly_repos.rake
```diff
@@ -31,6 +31,7 @@ namespace :pl do
     task :sign_repos => "pl:fetch" do
       Pkg::Util::RakeUtils.invoke_task("pl:sign_rpms", "repos")
       Pkg::Rpm::Repo.create_repos('repos')
+      Pkg::Rpm::Repo.sign_repos('repos')
```
stahnma: any reason not to sign other repos?

haus: I don't know what you mean. What other repos have signing in packaging today?

haus: When we have OSX ready, we can certainly start signing those, but that work hasn't landed yet.

haus:
Example of a signed rpm repo: http://agent-downloads.delivery.puppetlabs.net/facter/2.4.4.3/repos/el/6/products/i386/repodata/

shrug:
:+1:

ScottGarman:
:+1:

lib/packaging/util/gpg.rb
```diff
@@ -0,0 +1,45 @@
+module Pkg::Util::Gpg
+  class << self
+
+    def load_keychain
+      unless @keychain_loaded
+        unless ENV['RPM_GPG_AGENT']
```
mckern: If these don't happen, this method just returns `nil`. Is that desirable?

haus: Seems fine to me. That is how it has been used so far. I didn't want to change that interface in this PR.

lib/packaging/util/gpg.rb
```diff
@@ -0,0 +1,45 @@
...
+    def kill_keychain
+      keychain ||= Pkg::Util::Tool.find_tool('keychain')
+
+      if keychain
+        Pkg::Util::Execution.ex("#{keychain} -k mine")
+      end
+    end
+
+    def start_keychain
+      keychain ||= Pkg::Util::Tool.find_tool('keychain')
```
mckern: Can we move `#keychain` into a method of its own and let it handle resolving the path to the system `keychain`?

lib/packaging/util/gpg.rb
mckern: Fair enough.

lib/packaging/util/gpg.rb
haus: Sure. I tried it as both constants and instance variables and got very sad in doing so, but a method seems reasonable.

haus: same for gpg? or is the single use of gpg fine for now?

mckern: Single use is probably fine but I'd like to move towards that pattern for any place where we're looking for a path to a binary.

lib/packaging/util/gpg.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      if keychain
+        keychain_output = Pkg::Util::Execution.ex("#{keychain} -q --agents gpg --eval #{Pkg::Config.gpg_key}").chomp
+        new_env = keychain_output.match(/(GPG_AGENT_INFO)=([^;]*)/)
+        ENV[new_env[1]] = new_env[2]
```
mckern: This is functionally approximating the process of running `eval` on the output from `keychain`, right?

haus: yup.

mckern: I want to say "ick" but I honestly couldn't think of a more reliable way to do this. It's hacks all the way down.

haus: yea, the matching could be a little less awkward (why are we capturing a known string and then using it?), but it works.

mckern:
ej9qg1t

haus:
@mckern updated

mckern:
:+1: pending Travis

PR #415, opened 2015-06-24
(maint) Use correct dist macro definition in mock_defines

geoffnichols:
This updates the dist macro. As explained in RE-4897:

> Previously we incorrectly specified the dist macro. All --defines passed
> along the command line are accessed using % in the spec file, but
> defined without it on the command line. This commit removes the % so
> that the dist macro can be correctly defined for platforms that don't
> already have it available (el4, el5, sles10).


haus:
:+1:

shrug:
:+1:  when travis is happy

PR #417, opened 2015-06-25
(RE-4907) Update nightly_repos to avoid ref == version problems

haus:
When version == ref, the repo_configs will be named the same and moving
them onto themself won't work, and will fail the job. This commit avoids
the entire rewrite conditional in the case that the ref is equal to the
version, as it is not needed then.


stahnma:
:+1:

PR #419, opened 2015-06-30
(RE-4915) Remove use of the 're' user (which was removed from Jira)

kylog:


melissa:
:+1:

melissa:
Actually, I realized I should just merge this, so, this is in!

joshcooper:
nice!

PR #421, opened 2015-07-15
(RE-4956) Add osx signing to nightly repos tasks

haus:
This PR adds osx signing automation to the nightly repos task. To do that it adds a parameter to the sign method of Pkg::OSX.


shrug:
:+1:

PR #422, opened 2015-07-20
(RE-5122) Teach packaging how to uber_ship solaris 10 packages

haus:
This PR has the basics in place to all us to both ship and sign svr4 packages to a designated location. To do this we also need to add two new Pkg::Config parameters for svr4_host and svr4_path.


ScottGarman:
:+1:

tasks/ship.rake
```diff
@@ -137,6 +146,7 @@ namespace :pl do
       Rake::Task["pl:ship_debs"].invoke if Pkg::Config.cows || Pkg::Config.vanagon_project
       Rake::Task["pl:ship_dmg"].execute if Pkg::Config.build_dmg || Pkg::Config.vanagon_project
       Rake::Task["pl:ship_tar"].execute if Pkg::Config.build_tar
+      Rake::Task["pl:ship_svr4"].execute if Pkg::Config.vanagon_project
```
melissa: should we add `ship_ips` to this? Or are we just focused on solaris 10 right now?

haus: I'm focussed on solaris 10 right now.

PR #423, opened 2015-07-21
(CPR-157) Allow packaging repo to build deb repos with more arches

stahnma:
Previously we would only look for i386/amd64 architecture for packaging.
This commit adds all sorts of other arches. If we have no packages of
that arch, nothing fails.


haus:
:+1:

PR #424, opened 2015-07-22
(maint) Remove all as an arch

stahnma:
Apparently all is some sort of meta-arch and doesn't need to specified,
and if you do specify it, things blow up. Who knew?


shrug:
:+1:

PR #426, opened 2015-07-23
(packaging)(RE-5134) Copy files for latest instead of using symlinks

shrug:


tasks/nightly_repos.rake
```diff
@@ -139,7 +139,7 @@ namespace :pl do
         end
 
         # Make a latest symlink for the project
-        FileUtils.ln_s(File.join("..", local_target, "repos"), File.join(Pkg::Config.project + "-latest", "repos"))
+        cp_pr(File.join("..", local_target, "repos"), File.join(Pkg::Config.project + "-latest"))
```
haus: I really hate using methods from 00_utils.rake, but this change seems sane.

haus:
:+1: Let's migrate that method soon, or just use a raw fileutils call.

tasks/nightly_repos.rake
shrug: Yep, they should all be updated but this is consistent with the other calls in the method

PR #428, opened 2015-08-14
(maint) Add cisco-wrlinux and huaweios packages to rpm tasks

ScottGarman:


mckern:
Can the two OSes be split into two separate commits? I know it's a nitpick but the granularity will be helpful if we need to cherry pick or revert for some unforeseen reason.

ScottGarman:
@mckern splilt up into two commits as requested

mckern:
:+1:

PR #429, opened 2015-08-14
Ticket/master/re 5105

melissa:


lib/packaging/util.rb
```diff
@@ -33,11 +33,23 @@ def self.in_project_root(&blk)
...
 
-  def self.get_var(var)
-    check_var(var, ENV[var])
+  def self.get_var(var, required = true)
+    self.check_var(var, ENV[var], required)
     ENV[var]
   end
 
+  def self.check_var(varname, var = nil, required = true)
+    if var.nil?
```
haus: Can you reverse this?

```
if var
  var
else
  ...stuff
end
```

lib/packaging/util/jira.rb
```diff
@@ -21,22 +21,82 @@ def self.jira_project_name(projects, project)
...
+    def self.get_auth_vars
+      vars = {}
+      vars[:site]             = Pkg::Util.get_var("SITE", false) || 'https://jira1-test.ops.puppetlabs.net'
+      vars[:username]         = Pkg::Util.get_var("JIRA_USER")
+      vars[:password]         = Pkg::Util::Jira.get_password(vars[:site], vars[:username])
+      vars[:base64_encoding]  = Pkg::Util::Jira.get_base64_encoding(vars[:username], vars[:password])
+      vars
+    end
+
+    def self.get_password(site, user)
```
haus: This feels like is should be a more generic utility. Something like get_input with an option of disabling echo.

lib/packaging/util/jira.rb
```diff
@@ -21,22 +21,82 @@ def self.jira_project_name(projects, project)
...
+      puts "\nOkay trying to log in to #{site} as #{user} ..."
+      password
+    end
+
+    def self.get_base64_encoding(username, password)
+      require 'base64'
+      Base64.encode64("#{username}:#{password}").strip
+    end
+
+    def self.link_issues(inwardIssue, outwardIssue, site, authentication, type = 'Blocks')
```
haus: Pkg::Util::Net has a method called curl_form_data. It seems like it might fit in nicely here.

lib/packaging/util/jira.rb
```diff
@@ -67,12 +127,19 @@ def project_name(project)
       self.class.jira_project_name(@client.Project.all, project)
     end
 
-    def create_issue(summary, description, project, parent, assignee)
+    def create_issue(summary, description, project, parent, assignee = nil, story_points = nil, type = nil, components = nil, epic_parent = nil)
+      require 'jira'
```
haus: We should probably collect all of the `require` statements at the top. It is far more readable, and this is the jira class. We can always rescue the load error with a stern warning or add jira-ruby to the Gemfile.

lib/packaging/util.rb
```diff
@@ -33,11 +33,23 @@ def self.in_project_root(&blk)
...
   end
 
-  def self.get_var(var)
-    check_var(var, ENV[var])
+  def self.get_var(var, required = true)
+    self.check_var(var, ENV[var], required)
     ENV[var]
   end
 
+  def self.check_var(varname, var = nil, required = true)
```
MosesMendoza: Why is the `var` argument optional? If it is meant to be optional, how would I not supply it but supply `required` as false?

lib/packaging/util/jira.rb
```diff
@@ -21,22 +21,82 @@ def self.jira_project_name(projects, project)
       projects.find { |p| p.key == project }.name
     end
 
-    def self.jira_issue_fields(summary, description, project, parent, assignee)
+    def self.jira_issue_fields(summary, description, project, parent, assignee, story_points = nil, type = nil, components = nil, epic_parent = nil)
```
stahnma: Don't we want named parameters here and not just a bunch of defaults? I thought named params in ruby were like thing: value. https://robots.thoughtbot.com/ruby-2-keyword-arguments

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,431 @@
+# This rake task creates tickets in jira for a release.
+#
+def get_platform_ticket_vars
+  vars = {}
+
+  # configuration
+  # Which tickets are we going to need for this platform?
+  vars[:pe_only]      = Pkg::Util.get_var("PE_ONLY", false) || false
+  vars[:agent_only]   = Pkg::Util.get_var("AGENT_ONLY", false) || false
```
stahnma: Can we default this to true? We have like 10x the number of agents than we do servers.

lib/packaging/util/jira.rb
haus: We don't want that because it means we need to have ruby 2 or higher everywhere, which we don't on say our deb builders i don't think. Also, keyword args with defaults is ruby 2.1, keyword args is just ruby 2.

lib/packaging/util.rb
melissa: The `var` has always been this way. It sounds like it would make more sense to pull that optionality out though... I'll do that.

lib/packaging/util/jira.rb
melissa: I agree with that, but I'm not sure the best way to do that. We want to communicate to the user about what it is they're entering and why. Since this adds in a lot of complexity to the method, I feel like it makes sense to leave it as it is until we do get something that would use something similar?

tasks/agent-tickets.rake
melissa: I figured the more tickets the better if this wasn't set right. By defaulting to false, we'll get both agent and master tickets (once master tickets are a thing), and then we can just close the master tickets rather than trying to make them by hand if we accidentally forget to create them. But I can default it to true if you prefer.

lib/packaging/util/jira.rb
stahnma: So how does ruby handle default arguments for more than one parameter? Normally in most languages you can only specify a default value for the last param. In this case if I want to send N[-2] then I have to explicitly pass N[-4], N[-3] as well?

lib/packaging/util/jira.rb
haus: Leave that part up to the user of the method and not to the method itself?

lib/packaging/util/jira.rb
haus: you would need to explicitly pass nil for story_points and type if you wanted to specify something non-nil for components.

lib/packaging/util/jira.rb
melissa: ooh, I didn't even think of that, thanks!

lib/packaging/util/jira.rb
melissa: If there's a better way to pass these in, that's totally fine. I was also thinking about just doing args* type thing, and just assuming the order, but that would make passing some variables in and not others just as complicated as it is in this way.

haus: a common pre-ruby2 method is to use an args or options hash so that you can easily check for elements and set defaults as needed.

tasks/agent-tickets.rake
melissa: Do you still want me to default to true given my rational for defaulting to false here?

lib/packaging/util.rb
```diff
@@ -33,11 +34,24 @@ def self.in_project_root(&blk)
...
+  def self.get_var(var, default = nil)
+    self.check_var(var, ENV[var], default)
     ENV[var]
   end
 
+  def self.check_var(varname, var, default = nil)
+    if var
+      var
+    else
+      if default.nil?
```
haus: can we reverse this? if variable else end is much more readable than if variable.nil? else end

lib/packaging/util.rb
```diff
@@ -47,4 +61,13 @@ def self.require_library_or_fail(library, library_name = nil)
     end
   end
 
+  def self.get_input(noecho = true)
```
haus: you don't need the intermediate variable input here at all

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,425 @@
...
+  # enter password only to find out they typo'd one of the above
+  vars.merge(Pkg::Util::Jira.get_auth_vars)
+end
+
+def validate_platform_ticket_vars(jira, vars)
+end
+
+
+def create_platform_tickets(jira, vars)
+  description = {}
```
haus: Most of these descriptions are only used once. It looks like test_targets is the only one used more than once. It would be more readable for these to be inlined into the ticket hash, except perhaps the one that is used several times.

lib/packaging/util.rb
melissa: No, since there's still the possibility that default will be set to `false`, in which case this would do the wrong thing

haus: So a var can default to false but can't default to nil? Can you include some YARD style doc for this method, so that this is clear?

lib/packaging/util.rb
```diff
@@ -33,11 +34,37 @@ def self.in_project_root(&blk)
...
+  def self.get_var(var, default = nil)
+    self.check_var(var, ENV[var], default)
     ENV[var]
   end
 
+  # Utility to check if a variable is set
+  #
+  # @param varname [String] the name of the variable to be checked
+  # @param var [String] the contents of the variable to be checked
+  # @param default [String] what the var should default to if var is not set
```
haus: Besides String, default can also take `true`, `false`, and `nil`.

lib/packaging/util.rb
```diff
@@ -33,11 +34,37 @@ def self.in_project_root(&blk)
     result
   end
 
-  def self.get_var(var)
-    check_var(var, ENV[var])
+  # Utility to get the contents of an Environment variable
+  #
+  # @param var [String] The variable to return
+  # @param default [String] What the variable is set to if it is not already set
```
haus: for the type of the yard doc we want [String, true, false, nil] probably to show that it can accept more than just String.

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,425 @@
...
+  # What information do we need about this platform?
+  vars[:platform_tag] = Pkg::Util.get_var("PLATFORM_TAG")
+  vars[:codename]     = Pkg::Util.get_var("CODENAME", '')
+
+  # Jira authentication - do this after validating other params, so user doesn't need to
+  # enter password only to find out they typo'd one of the above
+  vars.merge(Pkg::Util::Jira.get_auth_vars)
+end
+
+def validate_platform_ticket_vars(jira, vars)
```
haus: Will there be a method body here or should this go away?

lib/packaging/util/jira.rb
haus: This still feels weird. Why not add this debugging/status output to the Pkg::Util::Jira.initialize method?

melissa:
This is only failing with Ruby 1.8,7, because it can't load 'io/console'

haus:
Do we still have 1.8.7 running anywhere?

haus:
I *think* osx was the lone holdout, but now osx-builder is gone.

lib/packaging/util.rb
mckern: `elsif`?

lib/packaging/util.rb
```diff
@@ -33,11 +34,39 @@ def self.in_project_root(&blk)
...
+  def self.get_var(var, default = nil)
+    self.check_var(var, ENV[var], default)
     ENV[var]
   end
 
+  # Utility to check if a variable is set
+  #
+  # @param varname [String] the name of the variable to be checked
+  # @param var [String] the contents of the variable to be checked
+  # @param default [String, Boolean, nil] what the var should default to if var is not set
```
haus: boolean isn't a ruby type

melissa: So this has to be `[String, true, false, nil]`?

mckern: @haus Boolean is valid in this context in yardoc.

mckern: @melissa Boolean is a keyword for Yard, which denotes that the values `true` or `false` will be returned becaue the alternative (which you described) is really verbose.

haus: fun. TIL

mckern: I know this was a pre-existing method, but the array of possible objects that it can return is sort of staggering.

melissa: I know, I was just wondering what the best thing to put as the return value would be

mckern: I think we might actually be able to lose this method entirely. It's only used by `get_var`?

lib/packaging/util.rb
```diff
@@ -33,11 +34,26 @@ def self.in_project_root(&blk)
     result
   end
 
+  # Utility to get the contents of an Environment variable
+  #
+  # @param var [String] The name of the environment variable to return
+  # @return [String, Boolean, Hash, Array, nil] The contents of ENV[var]
   def self.get_var(var)
```
haus: @melissa most of the other modules we have do something like

```
module Pkg::$thing
  class << self
```

to avoid needing to prefix all of the methods with self.

mckern: I have mixed feelings about that. I've definitely done it, but `self.` as a prefix makes intent explicit.

lib/packaging/util.rb
```diff
@@ -33,11 +34,26 @@ def self.in_project_root(&blk)
...
   end
 
+  # Utility to check if a variable is set
+  #
+  # @param varname [String] the name of the variable to be checked
+  # @param var [String, Boolean, Hash, Array, nil] the contents of the variable to be checked
+  # @return [String, Boolean, Hash, Array, nil] the contents of var
+  # @raise [RuntimeError] raises an exception if the variable is not set and is required
+  def self.check_var(varname, var)
+    var.nil? and fail "Requires #{varname} be set!"
```
mckern: `&&` is prefered to `and`. I know it's old code. We can fix that here though. Alternatively, consider:

```ruby
def self.check_var(varname, var)
  fail "Requires #{varname} be set!" if var.nil?
  var
end
```

Preserves the same weird-ass behavior, until/unless I can find time to fix that weird-ass behavior in all Rake tasks.

lib/packaging/util.rb
```diff
@@ -47,4 +63,15 @@ def self.require_library_or_fail(library, library_name = nil)
     end
   end
 
+  # Utility to retrieve command line input
+  # @param noecho [Boolean, nil] if we are retrieving command line input with or without privacy. This is mainly
+  #   for sensitive information like passwords.
+  def self.get_input(noecho)
+    if noecho
+      input = STDIN.noecho(&:gets).chomp
```
mckern: Prefer use of `$stdin` and `$stderr` to `STDIN` and `STDERR`. There's reasons why and I'm happy to explain them, but you probably don't want to know.

mckern: Suggestion:

```ruby
def self.get_input(echo = true)
  system 'stty -echo' if echo
  $stdin.gets.chomp
ensure
  system 'stty echo'
end
```

mckern: Revised:

```ruby
def self.get_input(echo = true)
  fail "Cannot get input on a noninteractive terminal" unless $stdin.tty?

  system 'stty -echo' unless echo
  $stdin.gets.chomp!
ensure
  system 'stty echo'
end
```

lib/packaging/util/jira.rb
```diff
@@ -21,31 +25,96 @@ def self.jira_project_name(projects, project)
...
+      if options_hash[:story_points]
+        fields['customfield_10002'] = options_hash[:story_points].to_i
+      end
+      if options_hash[:components]
+        fields['components'] = options_hash[:components]
+      end
+
+      if options_hash[:type]
+        fields['issuetype'] = { 'name' => options_hash[:type] }
+        if options_hash[:type] == 'Epic'
```
mckern: I'd move this `if options_hash[:type] == 'Epic'` block out of and below the `if options_hash[:type]` block.

lib/packaging/util/jira.rb
```diff
@@ -21,31 +25,96 @@ def self.jira_project_name(projects, project)
       projects.find { |p| p.key == project }.name
     end
 
-    def self.jira_issue_fields(summary, description, project, parent, assignee)
+    def self.jira_issue_fields(options_hash)
```
mckern: This entire method is complicated enough that I'm tempted to suggest we should break it out into a subclass and abstract the way Issues get constructed. Wow.

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+    },
+    {
+      :index        => '22',
+      :project      => 'QENG',
+      :summary      => "Add #{vars[:platform_tag]} agent to PE agent integration jobs",
+      :description  => description[:pe_integration],
+      :blocked_by   => ['21'],
+      :story_points => '1',
+      :components   => ['CI', 'Scrum Team - Integration'],
+    },
```
melissa: @justinstoller I noticed you were updating the jenkins tickets to be only 1 story point, and to include the components 'CI' and 'Scrum Team'. Are there any other updates like that that should happen in the script?

justinstoller: Nope!

lib/packaging/util.rb
melissa: I think I'm going to stick with what I currently have

lib/packaging/util/jira.rb
melissa: I've definitely like to work with you on the best way of doing that

mckern: Out of scope for this PR but we'll look at doing that soonish.

lib/packaging/util/jira.rb
```diff
@@ -21,31 +25,108 @@ def self.jira_project_name(projects, project)
...
+      # be linked to the main epic. Otherwise, we need to set it to have an epic_parent. This can
+      # either be an epic linked to the main epic or the main epic itself.
+      if options_hash[:parent]
+        fail "A ticket with a parent must be classified as a Sub-ticket\n\n#{options_hash}" unless (options_hash[:type] == 'Sub-task' || !options_hash[:type])
+        fields['issuetype'] = { 'name' => "Sub-task" }
+        fields['parent'] = { 'key' => options_hash[:parent] }
+      elsif options_hash[:epic_parent]
+        fail "This ticket cannot be a Sub-task of an epic\n\n#{options_hash}" if options_hash[:type] == 'Sub-task'
+        fields['customfield_10006'] = options_hash[:epic_parent]
       end
```
mckern: Is there any way to move the failure conditions above the ~40 lines of option setting so that the method fails a little faster? Or to move it as close as possible to the conditions/data it needs to determine if it should fail?

lib/packaging/util/jira.rb
```diff
@@ -21,31 +25,108 @@ def self.jira_project_name(projects, project)
...
 
       fields
     end
 
+    def self.get_auth_vars
+      vars = {}
+      vars[:site]     = ENV['SITE'] || 'https://tickets.puppetlabs.com'
```
mckern: `SITE` is probably a too-generic name for this environment variable. Also, I know there's no way around in right now but yuck, hardcoded URLs.

mckern: `SITE` is probably a too-generic name for this environment variable. Also, I know there's no way around in right now but yuck, hardcoded URLs.

lib/packaging/util/jira.rb
```diff
@@ -21,31 +25,108 @@ def self.jira_project_name(projects, project)
...
+      vars[:username] = Pkg::Util.get_var("JIRA_USER")
+      vars
+    end
+
+    def self.get_base64_encoding(username, password)
+      Base64.encode64("#{username}:#{password}").strip
```
mckern: Is the `strip` for whitespace on the `Base64.encode64` or on the `"#{username}:#{password}"` string?
Also, prefixing a method with `get_` implies it's connecting to something and retrieving some sort of data. Probably better to just name this `base64_encode`?

lib/packaging/util/jira.rb
melissa: I'm a little confused by the last option you stipulate here, Ryan. I was under the impression these failure cases were as close as possible to the conditions/data it needs to determine if it should fail. Can you give me an example of what you mean by that?

Also, what is the concern behind this comment? We are only parsing out the given data into a has that can be interpreted by jira-ruby. We aren't creating or updating a ticket yet, just manipulating data. Is speed a concern? Readability?

lib/packaging/util/jira.rb
mckern: Alternatively, since you only call it once, just call it directly instead of using a method.

lib/packaging/util/jira.rb
mckern: Strictly readability. The interpreter will functionally fail at the same speed, but a human reading it will take longer to get the idea into their brain.

lib/packaging/util/jira.rb
melissa: The strip is to remove the `\n` from the results of the base64 call. `bWVsaXNzYTplcm5lc3RvNTQ=\n` vs. `bWVsaXNzYTplcm5lc3RvNTQ=`

mckern: Oh, OK. Consider `strict_encode64` then:

```ruby
>> Base64.encode64('thisthing')
=> "dGhpc3RoaW5n\n"
>> Base64.strict_encode64('thisthing')
=> "dGhpc3RoaW5n"
```

lib/packaging/util/jira.rb
```diff
@@ -67,12 +148,16 @@ def project_name(project)
       self.class.jira_project_name(@client.Project.all, project)
```
mckern: What happened if these methods were called without `self.class` prefixing them?

lib/packaging/util/jira.rb
```diff
@@ -67,12 +148,16 @@ def project_name(project)
...
-      fields = self.class.jira_issue_fields(summary, description, project,
-                                            parent, assignee)
+    def create_issue(options_hash)
+      fields = self.class.jira_issue_fields(options_hash)
 
       issue = @client.Issue.build
-      issue.save!({ 'fields' => fields })
+      begin
+        issue.save!({ 'fields' => fields })
+      rescue Exception => e
```
mckern: This will rescue all exceptions, print an error, and then return `true` (instead of reraising or returning `false`). Rescuing all exceptions is kind of dicey, but not the scariest thing in our code base, but at a minimum this should probably return `false` if that happens.

Also, what are you feelings about the implicit `begin` in a method definition? It's purely stylistic, but it's the more common convention:

```ruby
def create_issue(options_hash)
  fields = self.class.jira_issue_fields(options_hash)

  issue = @client.Issue.build
  issue.save!({ 'fields' => fields })

rescue Exception => e
  puts "Cannot create Jira Ticket with fields #{fields}"
  puts e
end
```

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+  description[:pe_repo] = <<-DOC
+pe_repo needs to be udpated to support installing the agent
+DOC
+
+  description[:pre_suites] = <<-DOC
+Get git tests working. AIO test failures can be found in the pipeline, and ticketed from there.
+DOC
+
+  description[:module_support] = <<-DOC
+guys, we're supporting this.
```
mckern: Suggestions to use instead of `guys`: Folks? People? Listen?

mckern: Also, I love you use of `DOC` instead of `EOL` or `HERE`

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+      :summary      => "Intention to support #{vars[:platform_tag]} as an agent platform",
+      :description  => description[:module_support],
+    },
+    {
+      :index        => '24',
+      :project      => 'DOC',
+      :summary      => "Intention to support #{vars[:platform_tag]} as an agent platform",
+      :description  => description[:docs_support],
+    },
+  ]
```
mckern: This should just be a giant JSON or YAML blob, that we parse after it's defined. Much more readable.

tumblr_inline_nh7x46u6i31suuq6s

haus: a configfile?

tasks/agent-tickets.rake
stahnma: We should probably use something other than guys. Folks or people.

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+stand up build and test infrastructure for the new platform. Once that work has been completed, users may
+see the new platform available through nightly builds. Otherwise, it will become available whenever the
+next release happens.
+
+This epic is only to add support for this platform in the agent stack. The server stack work for this
+platform will happen in a separate epic.
+DOC
+
+  # Values for the main ticket
+  parent_project  = vars[:pe_only] == "true" ? 'RE' : 'CPR'
```
mckern: You used the ternary operator!

giphy

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+  parent_assignee = vars[:username]
+
+  jira.user?(parent_assignee)
+  jira.project?(parent_project)
```
mckern: Do these methods `raise` if they fail, or do they just return false?

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+
+  main_ticket = {
+    :summary => summary,
+    :description => description[:top_level_ticket],
+    :project => parent_project,
+    :assignee => parent_assignee,
+    :type => 'Epic',
+  }
+
+  parent_key, _ = jira.create_issue(main_ticket)
```
mckern: What gets assigned to `_` here?

melissa: the issue number, which Kylo was using, but I'm not. Should I just take it out?

tasks/agent-tickets.rake
melissa: https://github.com/melissa/packaging/blob/ticket/master/re-5105/lib/packaging/util/jira.rb#L135-L145

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+
+    # Define the relationship of all tickets that this ticket is blocked by
+    # so we can pass in the key of the blocking ticket
+    blocked_by = []
+    if subticket[:blocked_by]
+      subticket[:blocked_by].each do |linked_ticket|
+        blocked_by << subticket_hash[linked_ticket]['key']
+      end
+    end
+
```
mckern: Instead of preinitializing an empty `Array`, consider instead `map`?

```ruby
blocked_by = subticket[:blocked_by].map { |linked_ticket| subticket_hash[linked_ticket]['key'] }
```

It should return an empty `Array` if there's no blocking tickets.

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+    # Define the relationship of all tickets this ticket blocks so we can
+    # pass in the key of the blocked ticket
+    blocks = []
+    if subticket[:blocks]
+      subticket[:blocks].each do |linked_ticket|
+        blocks << subticket_hash[linked_ticket]['key']
```
mckern: See: `map`

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,414 @@
...
+
+
+  ## SUPPORTING TICKETS
+
+  # We want to keep track of the tickets we've already created so that we can define
+  # relationships between them in a human-readable way.
+  subticket_hash = {}
+
+  # Create subtasks for each step of the release process
+  subtickets.each do |subticket|
```
mckern: Looking at how many times the various keys in the `subticket` iterator are called, it's probably worth breaking that out into a `Class` with `attr_accessors` later on. Out of scope for this PR, but it's a definite code smell.

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,426 @@
...
+      :project      => 'HI',
+      :summary      => "Get pre-suites working for #{vars[:platform_tag]} for Hiera",
+      :description  => description[:test_targets],
+      :story_points => '2',
+      :blocked_by   => ['16'],
+    },
+    {
+      :index        => '20',
+      :project      => 'QENG',
+      :summary      => "Update agent jenkins jobs to include #{vars[:platform_tag]}",
```
justinstoller: Actually, I do have a request. Just doing some scanning over our sprint board and the difference between the two QENG summaries makes it a little confusing to group the work. Can we make them similar?

I don't have a strong preference, but maybe something like:
`Add #{vars[:platform_tag]} to (puppet-agent|PE Integration) pipelines`

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,414 @@
...
+
+  task :agent_tickets do
+    vars = get_platform_ticket_vars
+    jira = Pkg::Util::Jira.new(vars[:username], vars[:site])
+
+    vars[:base64_encoding] = Pkg::Util.base64_encode("#{vars[:username]}:#{jira.client.options[:password]}")
+
+    puts "Creating new platform tickets based on:"
+    require 'pp'
+    pp vars.select { |k, v| k != :password }
```
mckern: if you're not using `v`, the convention is to use `_` instead.

tasks/tickets.rake
```diff
@@ -1,16 +1,7 @@
...
-  require 'io/console'
-  puts "Logging in to #{site} as #{user}"
-  print "Password please: "
-  password = STDIN.noecho(&:gets).chomp
-  puts "\nOkay trying to log in to #{site} as #{user} ..."
-  password
-end
-
-def get_vars
+def get_release_ticket_vars
```
mckern: What's actually being done here? It looks like it's just initializing an empty `Hash` and then merging it with another `Hash`?

tasks/tickets.rake
```diff
@@ -43,7 +30,7 @@ def validate_vars(jira, vars)
   jira.user? vars[:tester]
 end
```
mckern: Same assumption as earlier, that the `jira` "question mark" instance methods actually `raise` instead of returning a boolean?

lib/packaging/util/jira.rb
```diff
@@ -21,31 +24,109 @@ def self.jira_project_name(projects, project)
...
+        'outwardIssue'  => { 'key'  => outwardIssue },
+      }
+
+      uri = "#{site}:443/rest/api/2/issueLink"
+      form_data = ['-D-',
+                   '-X POST',
+                   "--data '#{data.to_json}'",
+                   "-H 'Authorization: Basic #{authentication}'",
+                   "-H 'Content-Type: application/json'"]
+      options = { :quiet => true }
```
kylog: Out of scope for this PR right now, but have you looked into whether we could PR jira-ruby so that is has this capability?

tasks/tickets.rake
```diff
@@ -392,14 +379,17 @@ DOC
...
-                                     project,
-                                     nil,     # no parent id
-                                     assignee)
+  parent_key, parent_id = jira.create_issue(main_ticket_hash)
 
-  puts "Main release ticket: #{key} (#{assignee}) - #{summary}"
+  puts "Main release ticket: #{parent_key} (#{assignee}) - #{summary}"
 
   # Create subtasks for each step of the release process
   subticket_idx = 1
```
mckern: YESSSS, HARDCODE ALL THE VALUES

tasks/agent-tickets.rake
kylog: +1 on folks or people rather than guys.

But also: what does this ticket mean? What are the criteria for being done? This is fuzzy to me.

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,414 @@
...
+Get git tests working. AIO test failures can be found in the pipeline, and ticketed from there.
+DOC
+
+  description[:module_support] = <<-DOC
+guys, we're supporting this.
+DOC
+
+  description[:docs_support] = <<-DOC
+guys, we're supporting this.
+DOC
```
kylog: Same question here: What are the criteria for being done with this ticket?

mckern:
So. After reading this all through, there's one thing I'd like throw out there:

This will work; there's some obvious comments but no general concerns with the code. I'm OK merging it because I don't want to hold @melissa's work hostage. But this code should really be extracted from the Packaging Repo and integrated w/ the Jira library so we can write smaller stand-alone scripts around it -- this is basically pushing Rake to and beyond what I would consider the limits of reasonable use.

tasks/agent-tickets.rake
justinstoller: If this is a new supported platform for PE the module team needs to evaluate whether or not currently supported modules should be updated to support it and ticket accordingly. ie add it to stdlib's pipeline, a new ubuntu or debian should go into the apt module pipeline, etc

tasks/agent-tickets.rake
```diff
@@ -0,0 +1,414 @@
...
+      :description  => description[:pooler_image],
+      :story_points => '5',
+    },
+    {
+      :index        => '5',
+      :project      => 'RE',
+      :summary      => "Add #{vars[:platform_tag]} platform definition to pl-build-tools-vanagon",
+      :description  => description[:pl_build_tools_configuration],
+      :story_points => '2',
+      :blocked_by   => ['4'],
```
kylog: Seeing this use of `blocked_by` makes me think this is ripe for bugs. E.g. I add a new element and then I need to update all the following indices *and* any blocking relationships. I am 100% sure that I will get this wrong the first time I try to do such surgery.

Could we make `blocked_by` use a short_name or something human mnemonic like that? Then we don't need to bump indices for unrelated tickets when adding a ticket, and blocked by relationships don't need adjusting.

kylog:
This looks great! Given that this is already in use, and that it touches common code with tickets.rake (which I need to PR), I'm inclined to do the minimum so this passes travis, merge it, and then file remaining comments as issues we can follow up on.

Thoughts?

kylog:
And re travis: how about we drop 1.8.7 and fix the one rubocop nit?

lib/packaging/util/jira.rb
melissa: I tried, and I couldn't come up with anything that works :(

lib/packaging/util/jira.rb
melissa: By soonish, do you mean in another year?

tasks/agent-tickets.rake
melissa: Do we need to wait until pe_repo has been updated to update the pe jenkins jobs? Do we want to make sure the new platform moves through the foss pipelines smoothly before adding it to the pe jenkins pipelines?

justinstoller: I believe the answer is 'yes' to both.

/cc @doug-rosser for an ack

doug-rosser: Whatisthisidonteven?!?!

justinstoller: I think that's a yes?

melissa: So, are you asking to combine those two tickets into a single ticket? Or are you asking to have one ticket definition that is used to create both tickets?

lib/packaging/util/jira.rb
mckern: Maybe two!

tasks/agent-tickets.rake
doug-rosser: I need to start drinking coffee again. Yes, pe_repo should be (must be?) updated before adding new stuffs to pe jenkins integration pipelines. Ideally foss pipelines go green before adding to pe jenkins pipelines.

doug-rosser: I'd rather have two tickets, as my Integration stuffs is pretty orthogonal to the agent workings. I don't care how they're implemented here.

stahnma:
Agree with @kylog. Let's get working code merged and open new issues on the style-points/abstraction other recommendations.

tasks/agent-tickets.rake
justinstoller: I want two tickets, but I want the summary of both tickets to follow the same sentence structure so while scanning our board I can easily understand which are CI tickets for a single platform.

I'm not particular about whether it is, "Add X Platform to Y Jenkins pipelines" or "Update Y Jenkins pipelines with X Platfrom", just as long as its consistent across the QENG tickets.

It's, also, not a reason to hold up this PR. It's just a nice to have I noticed when actually trying to organize the tickets into our next sprint.

tasks/agent-tickets.rake
melissa: `NoMethodError: undefined method `map' for nil:NilClass`

mckern: Ugh. That makes sense if there's no assumption that `subticket[:blocked_by]` exists. I think I hadn't noticed that `subticket` was an iterator, and not a specific object or method. yuck. What you're doing is fine and it'll work until we can refactor all of this into classes.

tasks/agent-tickets.rake
mckern: ignore me -grand galactic inquisitor

PR #430, opened 2015-09-08
(maint) Remove Ruby 1.8

melissa:
We no longer support Ruby 1.8 in the packaging repo. Previously, this
was kept around because we were using some builders that only had Ruby
1.8 available for use. However, we have since gotten rid of those
builders, and now have the luxury of having at least Ruby 1.9 in all
situations where we will use the automation in the packaging repo.


mckern:
:+1: on this

PR #432, opened 2015-09-09
(MAINT) Let Travis use modern infrastructure

mckern:
Without `sudo: false` in .travis.yml, the tests will run on legacy
travis infrastructure; this means that tests take FOREVER to run.
Since we shouldn't need root access to test this repo, we can speed
up our runs by disabling root access. Eg. `sudo: false`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 script: "bundle exec $CHECK"
+sudo: false
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


stahnma:
Your commit message doesn't make sense. I think you might have a word.

mckern:
Fixified.

stahnma:
:+1:

PR #434, opened 2015-09-09
(maint) IPS shipping fixes

haus:
This PR includes a few fixes to allow shipping of vanagon projects to our ips repository.


tasks/ship.rake
```diff
@@ -79,25 +79,35 @@ namespace :pl do
...
+    tmpdir.chomp!
+
+    Pkg::Util::Net.rsync_to(source_dir, Pkg::Config.ips_host, tmpdir)
+
+    remote_cmd = %(for pkg in #{tmpdir}/*.p5p; do
+    sudo pkgrecv -s $pkg -d #{Pkg::Config.ips_path} '*';
+    done)
+
+    Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.ips_host, remote_cmd)
+    Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.ips_host, "sudo pkgrepo refresh -s #{Pkg::Config.ips_path}")
```
stahnma: Do we have documented or automatically setup somewhere that we're using sudo?

stahnma:
:+1:

tasks/ship.rake
haus: We're sshing in as the jenkins user or ourselves, so not using sudo isn't an option (for svcadm, it might work for pkgrepo). I don't know where we would document this.

PR #435, opened 2015-09-15
(maint) Swap logic around which architecture to sign

melissa:
We had this problem where, if a package was noarch, we assumed we were
building it for both i386 and x86_64 architectures. However, there are
some platforms that we only build one architecture for. In the case of
these platforms that we built 64-bit packages for, we're not actually
signing them. So, that's a problem. This commit flips the logic of the
problem. I've decided 64-bit packages are more important than 32-bit
packages.


haus:
seems like a fine workaround. can you clean up the commit message a little? (add - to 32-bit, packags => packages)

haus:
:+1:

PR #436, opened 2015-09-15
(maint) Be more clear about version queries

HAIL9000:
Prior to this commit, the ticket for Jira tidy up suggested that
the "introduced in" query be made against version "X.Y.Y" and the
"fixed in" query against "X.Y.Z". Since both should be made against
the same version (the version being released), this description
was a little misleading and has caused some confusion.

Update the rake task so that they are both for version "X.Y.Z".


haus:
X.X.X

haus:
:+1:

PR #437, opened 2015-09-21
(RE-5524) Add whitelisting support to nightly_repos

haus:


haus:
need to do some more testing and add windows to the platforms.

haus:
i've added windows and am rolling some tests. please review.

haus:
my tests have passed.

haus:
build_defaults.yaml would have something like the following added:

```
foss_platforms:
  - el-6-i386
  - ubuntu-14.04-amd64
  - windows-2012-x64
pe_platforms:
  - el-4-i386
  - ubuntu-10.04-amd64
```

haus:
I think in the ideal world we would move the platforms module into its own gem. This would require that both enterprise-dist and package be runnable via ruby scripts and not rake. That may be a long journey.

lib/packaging/config/platforms.rb
```diff
@@ -0,0 +1,59 @@
+module Pkg::Platforms
+  # Each element in this hash
+  PLATFORM_INFO = {
+    'el' => {
+      '4' => { :architectures => ['i386', 'x86_64'], :repo => false, :package_format => 'rpm', :signature_format => 'v3', },
```
stahnma: I feel like this info may be other places :cry:

stahnma:
:+1: , but man, I want to fix up these platform hashes.

lib/packaging/config/platforms.rb
haus: if only our tools could rely on external libraries

haus:
@stahnma yea i think it's worth prioritizing the minimal effort to get packaging totally usable from a bundle install. then we can roll libraries of things like this, that something like beaker could consume as well.

joshcooper:
We'll need to update this when we add new foss platforms. How about we update the ticket generator to remind us?

haus:
@joshcooper Good idea.

geoffnichols:
Created RE-5631 for @joshcooper's recommendation.

PR #439, opened 2015-09-23
(RE-5613) Don't fail if no swix packages to sign

geoffnichols:
Prior to this commit, the `pl:sign_swix` and
`pl:sign_all` tasks fail if there are no swix
packages to sign. Current enterprise-dist
package promotion workflow removes packages
that won't be promoted into enterprise-dist
before invoking packaging's `pl:sign_all`
task. This means swix packages are removed
before the signing attempt, which causes
package promotion to fail, since there are
no swix packages to sign.

This commit adjusts the `pl:sign_swix` task
so it does not fail if there are no swix packages
to sign. (This is similar to the current
behavior of `sign_svr4` and `sign_osx`.)


stahnma:
:+1:

geoffnichols:
Updated commit to add ticket number.

PR #441, opened 2015-09-23
(maint) Platform codenames can't be symbols

haus:
In the initial work I neglected to update the cumulus codename to be a
string and not a symbol in the platform hash. This breaks the whitelist
automation if cumulus is included.


melissa:
:+1: pending travis

PR #443, opened 2015-09-30
(RE-5577) Fix osx signing paths

shrug:
Currently, when signing osx pkgs, when the automation
retrieves the DMGs it's dropping them in the PC1 directory,
meaning when we ship we ship signed packages in the wrong
location, and unsigned packages in the correct location.

This updates the retrieval to put the packages in the correct
location when they get retrieved.


shrug:
The other option is to send over the complete directory structure, sign them in place, then rsync it back

shrug:
I have not yet tested this

stahnma:
Please test.

shrug:
OK, tested this seems to be working as expected. The pkg's in the dmg's are now signed in the right place.

haus:
:+1:

PR #444, opened 2015-10-01
(maint) Only attempt to ship swix packages if they are there

melissa:


tasks/ship.rake
```diff
@@ -138,6 +138,18 @@ namespace :pl do
...
+  task :ship_swix => 'pl:fetch' do
+    packages = Dir['pkg/eos/**/*.swix']
+    if packages.empty?
+      STDOUT.puts "There aren't any swix packages in pkg/eos. Maybe something went wrong?"
+    else
+      Pkg::Util::Execution.retry_on_fail(:times => 3) do
+        Pkg::Util::Net.rsync_to("pkg/eos/", Pkg::Config.tar_host, Pkg::Config.swix_path)
+      end
+    end
+  end if Pkg::Config.build_swix || Pkg::Config.vanagon_project
```
haus: We shouldn't need the task guarded by the if as there are already guards in the task.

haus:
:+1:

PR #446, opened 2015-10-05
(RE-5550) Latest was not set correctly

mhaskel:
We don't want to symlink latest into the PE version since lastest
deployed is not deployed for all PEs, and linking the version doesn't
have the correct repo_configs, so copy latest over to the version of PE
we're shipping for.

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -210,18 +210,15 @@ namespace :pl do
     #
     # @param target_host the remote host where the packages are being shipped
     #        ex: agent-downloads.delivery.puppetlabs.net
-    # @param pa_source the path all versions of puppet-agent are deployed to
-    #        ex: /opt/puppet-agent/puppet-agent
-    # @param pe_target the path puppet-agent is deployed to within a PE version
-    #        ex: /opt/puppet-agent/2015.2/puppet-agent
+    # @param remote_dir the base path to deploy packages to
+    #        ex: /opt/puppet-agent
     # @param versioning whether the puppet-agent version is a version string or
     #        a github ref. Valid values are 'version' and 'ref'
-    # @param pe_version the PE-version to deploy to. Must match the version
-    #        from pe_target.  ex: 2015.2
-    task :link_signed_repos, [:target_host, :pa_source, :pe_target, :versioning, :pe_version] => ["pl:fetch"] do |t, args|
+    # @param pe_version the PE-version to deploy to.
+    #        ex: 2015.2
+    task :link_signed_repos, [:target_host, :remote_dir, :versioning, :pe_version] => ["pl:fetch"] do |t, args|
       target_host = args.target_host or fail ":target_host is a required argument for #{t}"
-      pa_source = args.pa_source or fail ":pa_source is a required argument for #{t}"
-      pe_target = args.pe_target or fail ":pe_target is a required argument for #{t}"
+      remote_dir = args.remote_dir or fail ":remote_dir is a required argument for #{t}"
       versioning = args.versioning or fail ":versioning is a required argument for #{t}"
       pe_version = args.pe_version or fail ":pe_version is a required argument for #{t}"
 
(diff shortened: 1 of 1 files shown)


mhaskel:
On a test run we ended up with the following in latest/repo_configs under the PE version we were deploying:

```
root@weth:/opt/puppet-agent/2012.1# ll puppet-agent-latest/repo_configs/*
puppet-agent-latest/repo_configs/deb:
total 4.0K
-rw-r--r-- 1 jenkins jenkins 138 Oct  5 14:21 pl-puppet-agent-latest-trusty.list

puppet-agent-latest/repo_configs/rpm:
total 4.0K
-rw-r--r-- 1 jenkins jenkins 232 Oct  5 14:21 pl-puppet-agent-latest-el-7-x86_64.repo
```

haus:
@mhaskel what's the contents of those files?

mhaskel:
Adding the sed steps resulted in the following in the repo_configs under 2012.1
```
root@weth:/opt/puppet-agent/2012.1/puppet-agent-latest# cat repo_configs/rpm/pl-puppet-agent-latest-el-7-x86_64.repo
[pl-puppet-agent-latest]
name=PL Repo for puppet-agent at commit latest
baseurl=http://weth.delivery.puppetlabs.net/2012.1/puppet-agent-latest/repos/el/7/PC1/x86_64/
enabled=1
gpgcheck=1
gpgkey=http://weth.delivery.puppetlabs.net/4BD6EC30
```

```
root@weth:/opt/puppet-agent/2012.1/puppet-agent-latest# cat repo_configs/deb/pl-puppet-agent-latest-trusty.list
# Packages for puppet-agent built from ref latest
deb http://weth.delivery.puppetlabs.net/2012.1/puppet-agent-latest/repos/apt/trusty trusty PC1
```

tasks/nightly_repos.rake
```diff
@@ -207,24 +207,28 @@ namespace :pl do
...
+      local_pa = File.join(pa_source, version_string)
+      local_pe = File.join(pe_target, version_string)
+      local_pa_latest = "#{pa_source}-latest"
+      local_pe_latest = "#{pe_target}-latest"
+
       Pkg::Util::Net.remote_ssh_cmd(target_host, "mkdir -p '#{pe_target}'")
-      Pkg::Util::Net.remote_ssh_cmd(target_host, "ln -sf '#{local_pa}' '#{local_pe_latest}'")
+      Pkg::Util::Net.remote_ssh_cmd(target_host, "mkdir -p '#{local_pe_latest}'")
+      Pkg::Util::Net.remote_ssh_cmd(target_host, "cp -r #{local_pa_latest}/* #{local_pe_latest}")
+      Pkg::Util::Net.remote_ssh_cmd(target_host, "sed -i 's/\\/#{File.basename(local_pa_latest)}/\\/#{pe_version}\\/#{File.basename(local_pa_latest)}/' #{local_pe_latest}/repo_configs/*/*")
```
haus: you can use `|` here to separate the sed sections and then you shouldn't need to escape the slashes.

tasks/nightly_repos.rake
```diff
@@ -207,24 +207,28 @@ namespace :pl do
     # We want to keep the puppet-agent repos at a higher level and them link
     # them into the correct version of PE. This is a private method and is
     # called from the internal_puppet-agent-ship jenkins job
-    task :link_signed_repos, [:target_host, :pa_source, :pe_target, :versioning] => ["pl:fetch"] do |t, args|
```
haus: Can you add some comments about what the task arguments are for? The names don't make it clear to me.

mhaskel:
@haus sed updated for marginally less terrible, will add comments now

tasks/nightly_repos.rake
```diff
@@ -207,24 +207,39 @@ namespace :pl do
...
+    #
+    # @param target_host the remote host where the packages are being shipped
+    #        ex: agent-downloads.delivery.puppetlabs.net
+    # @param pa_source the path all versions of puppet-agent are deployed to
```
haus: Suggestion: if there were an argument called remote_dir that was `/opt/puppet-agent` in this example, then pe_target, and pa_source could be derived. `File.join(remote_dir, Pkg::Config.project)` and `File.join(remote_dir, Pkg::Config.project, pe_version)`.

It would bring the argument count down to 4.

mhaskel:
@haus how about now?

haus:
@mhaskel this looks good to me. does it still work?

mhaskel:
@haus yep! Ran the jobs and made sure everything was still in the right place

haus:
:+1:

PR #448, opened 2015-10-12
(maint) Let's actually ship ips packages during the uber_ship

haus:
Previously the uber_ship neglected to include update_ips_repo, which is
good because any builds that didn't have solaris 11 packages would have
failed in a large ball of flames. This commit adds update_ips_repo to
the uber_ship task and also updates the task itself to not fail but just
to helpfully warn so that builds without solaris 11 packages won't fail.


mhaskel:
:+1:

shrug:
:+1:

PR #449, opened 2015-10-12
(maint) Move update_ips_repo to the remote namespace

mhaskel:
This improves consistency with the other update_*_repo tasks that
already live under remote.


shrug:
:+1:

PR #452, opened 2015-10-19
(maint) Add nil check for foss_platforms in nightly_repos.rake

haus:
Previously the checks were insufficient in the nightly repos ship job
for projects that have not defined foss_platforms to whitelist nightly
repo ships. If foss_platforms were nil, the empty? check would explode
in a empty? not defined for nilclass error. This is bad. This commit
avoids that problem by ensuring that foss_platforms is set and non-nil
before running .empty? on it.


stahnma:
:+1:

PR #453, opened 2015-10-19
(maint) Add debug flag to Pkg::Util::Execution.ex

haus:
It is currently fairly difficult to get details of why a command has
failed or what it returned by default in the packaging repo. This commit
aims to make that a little easier but adding a parameter to the ex
method to expose what command was executed and what it returned.
This PR also includes a commit which passes the verbose parameter to the ex method when rsync_to
and rsync_from methods are invoked. This should give a better picture of
what is happening when debugging failures.


shrug:
:+1:

PR #454, opened 2015-10-20
(maint) The task should always exist

underscorgan:
There were some old conditionals that made the update_ips_repo task only
exist for vanagon packages or when explicitly building for IPS. We fail
gracefully within the task if there are no solaris packages, so this is
unnecessary.


haus:
@underscorgan while you're there...https://github.com/puppetlabs/packaging/pull/454/files#diff-02259d8a37d5ca4f20dfe072004f5b82R112 ?

underscorgan:
@haus updated!

haus:
:+1:

PR #455, opened 2015-10-20
(CPR-243) Add ability to sign for Solaris 11

underscorgan:
This will allow us to sign Solaris 11 packages from rake tasks in
the packaging repo. Signing for Sol11 is a complicated process, so we
rsync the files up to the signing server, create a new repo there, sign
the repo, pull the packaging files back out, and rsync them back to our
build box. This adds the Pkg::IPS library to be able to sign packages,
and updates the rake task to use the new functionality.


tasks/sign.rake
```diff
@@ -130,12 +121,9 @@ namespace :pl do
...
   desc "Sign ips package, uses PL certificates by default, update privatekey_pem, certificate_pem, and ips_inter_cert in project_data.yaml to override."
-  task :sign_ips, :repo_uri, :fmri do |t, args|
-    repo_uri  = args.repo_uri
-    fmri      = args.fmri
-    puts "Signing ips packages..."
-    sign_ips(fmri, repo_uri)
-  end if Pkg::Config.build_ips
+  task :sign_ips do
+    Pkg::IPS.sign unless Dir['pkg/solaris/11/**/*.p5p'].empty?
+  end if Pkg::Config.build_ips || Pkg::Config.vanagon_project
```
haus: Let's always define the task. That may be less confusing.

underscorgan:
Ok, with the updated looping I was able to verify that I can install the signed packages on sol11 (and got it to fail with unsigned packages, so that seems like good evidence that they are signed?)

haus:
@underscorgan i'm :+1: here. can we ticket reaching out to oracle for cache flushing advice to clean this up?

haus:
@underscorgan there is a bit more wiring here, but this is a good step 1. the rest of the bits can be a follow up PR.

PR #456, opened 2015-10-21
(maint) Updates to call the IPS signing tasks

underscorgan:


haus:
@underscorgan this looks reasonable. We'll want to work out the jenkins ssh key before merging, so things don't break.

tasks/ship.rake
```diff
@@ -70,15 +70,25 @@ namespace :pl do
...
 
-  desc "Ship svr4 packages to #{Pkg::Config.yum_host}"
+  desc "Ship svr4 packages to #{Pkg::Config.svr4_host}"
   task :ship_svr4 do
     Pkg::Util::Execution.retry_on_fail(:times => 3) do
       if File.directory?("pkg/solaris/10")
         Pkg::Util::Net.rsync_to('pkg/solaris/10', Pkg::Config.svr4_host, Pkg::Config.svr4_path)
       end
     end
+
```
haus: Looks like the ends here are out of whack. The end below ship_p5p should probably move up to end the ship_svr4 task.

tasks/ship.rake
```diff
@@ -172,6 +185,7 @@ namespace :pl do
       Rake::Task["pl:ship_swix"].execute if Pkg::Config.vanagon_project
       Rake::Task["pl:ship_tar"].execute if Pkg::Config.build_tar
       Rake::Task["pl:ship_svr4"].execute if Pkg::Config.vanagon_project
+      Rake::Task["pl:ship_ips"].execute if Pkg::Config.build_ips || Pkg::Config.vanagon_project
```
haus: So because of how the uber_ship works, this is going to end up running update_ips_repo twice. We may want to break that chaining above.

haus: (update_ips_repo is directly invoked here: https://github.com/puppetlabs/packaging/blob/master/tasks/jenkins.rake#L253, which is the usual uber_ship entry point)

underscorgan: Ah, ok. So maybe get rid of `ship_ips` all together and just have `ship_p5p`?

haus:
@underscorgan that sounds good to me

underscorgan:
This was tested with an internal puppet-agent ship, signed packaged were shipped and verified!

haus:
:+1:

PR #458, opened 2015-10-29
Revert " (RE-4595) Add AIX rpms to the list that packaging repo signs"

melissa:
This reverts commit 896ea5f035fb9a4dc76774d87d6692bdfdb1e874.

Currently, signing on AIX is broken, and it prevents us from signing any
other packages

Conflicts:
	tasks/sign.rake


haus:
:+1:

PR #470, opened 2015-11-17
(maint) Remove deploy_apt_repos from the uber_ship

underscorgan:
The current rsyncing ends up copying the entire repository over every
time freight is run. This takes a very long time and causes outages
since the link target is updated before the new timestamped directory is
populated. We need to deploy manually until this can be fixed.

### DIFF
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -250,12 +250,15 @@ namespace :pl do
 
     desc "Retrieve packages built by jenkins, sign, and ship all!"
     task :uber_ship => "pl:fetch" do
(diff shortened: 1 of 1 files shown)


mckern:
:+1:

mckern:
@underscorgan @stahnma did a ticket get opened up for this while I was out? I can open one up and get it into the sprint tonight if not.

mckern:
@stahnma RE-5998 was opened earlier and relates to this; should we open a separate ticket for this (or the downtime) specifically? We can get that in in the AM before the sprint if so.

PR #471, opened 2015-11-19
(RE-5681) Add the ability to ship nuget packages

melissa:


tasks/ship.rake
```diff
@@ -180,6 +180,17 @@ namespace :pl do
     end
   end
 
+  desc "ship Windows nuget packages to #{Pkg::Config.nuget_host}"
+  task :ship_nuget => 'pl:fetch' do
+    packages = Dir['pkg/windows/**/*.nupkg']
+    if packages.empty?
+      STDOUT.puts "There aren't any nuget packages in pkg/windows. Maybe something went wrong?"
```
ferventcoder: Are we fine with not bubbling up an error?

ferventcoder:
Probably needs some MOAR documentation. Especially since we are doing it in a very particular way (aka Nexus) and directly posting packages where they would end up if we did the same using choco push to a source.

ferventcoder:
Was this closed on purpose?

melissa:
Yes, meant to be closed.

tasks/ship.rake
melissa: We're not always going to have nuget packages to ship, so yes, this is intended behaviour

PR #472, opened 2015-11-19
(RE-5998) Updates to debian shipping

underscorgan:
We don't want to ship the entirety of the freight output since it creates an entire copy of the repository on each run. This updates the rsync command to only copy the actual changes.

Additionally this adds support to initially copy to a staging directory on the APT_HOST to avoid network quirks causing issues with the apt repository.


underscorgan:
I'm in the process of testing this between weth and deb-builder-5, but please review!

underscorgan:
Successfully tested between weth and deb-builder-5. Looks to only be rsyncing what it should be rsyncing.

stahnma:
:+1: Hold on to yer butts

PR #473, opened 2015-11-20
(RE-5681) Ship Nuget Packages to Repo

jcoconnor:
Add new target to packaging scripts to support nuget (nupkg) style
packages for windows.

Note - this supersedes #471

### DIFF
--- a/lib/packaging/nuget.rb
+++ b/lib/packaging/nuget.rb
@@ -1,0 +1,36 @@
+module Pkg::Nuget
+  class << self
+    def ship(packages)
+      #
+      # Support shipping of Nuget style packages to a nexus based nuget feed
+      # Using curl to submit the packages rather than windows based choco/mono.
+      # This approach gives more flexibility and fits in with the current Puppet
+      # release automation practices using linux/mac systems.
+
+      # Sanity checks
+      fail "NUGET_HOST is not defined" if Pkg::Config.nuget_host.empty?
+      fail "NUGET_REPO is not defined" if Pkg::Config.nuget_repo_path.to_s.empty?
+
+      # Retrieve password without revealing it
+      puts "Obtaining credentials to ship to nuget feed #{Pkg::Config.nuget_repo_path} on #{Pkg::Config.nuget_host}"
+      print "Username please: "
+      username = Pkg::Util.get_input(true)
+      print "Password please: "
+      password = Pkg::Util.get_input(false)
+      authentication = Pkg::Util.base64_encode("#{username}:#{password}")
+
+      uri = "#{Pkg::Config.nuget_host}#{Pkg::Config.nuget_repo_path}"
+      form_data = ["-H 'Authorization: Basic #{authentication}'", "-f"]
+      packages.each do |pkg|
+        puts "Working on package #{pkg}"
+        projname, version = File.basename(pkg).match(/^(.*)-([\d+\.]+)\.nupkg$/).captures
+        package_form_data = ["--upload-file #{pkg}"]
+        package_path = "#{projname}/#{version}/#{File.basename(pkg)}"
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          Pkg::Util::Net.curl_form_data("#{uri}/#{package_path}", form_data + package_form_data)
+        end
+        fail "The Package upload (curl) failed with error #{$?.exitstatus}" unless $?.success?
+      end
+    end
+  end
+end
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -190,6 +190,16 @@ namespace :pl do
     end
   end
 
+  desc "ship Windows nuget packages to #{Pkg::Config.nuget_host}"
+  task :ship_nuget => 'pl:fetch' do
+    packages = Dir['pkg/windows/**/*.nupkg']
+    if packages.empty?
+      STDOUT.puts "There aren't any nuget packages in pkg/windows. Maybe something went wrong?"
+    else
+      Pkg::Nuget.ship(packages)
+    end
+  end
+
   desc "UBER ship: ship all the things in pkg"
   task :uber_ship => 'pl:fetch' do
     if confirm_ship(FileList["pkg/**/*"])
@@ -199,6 +209,7 @@ namespace :pl do
       Rake::Task["pl:ship_debs"].invoke if Pkg::Config.cows || Pkg::Config.vanagon_project
       Rake::Task["pl:ship_dmg"].execute if Pkg::Config.build_dmg || Pkg::Config.vanagon_project
       Rake::Task["pl:ship_swix"].execute if Pkg::Config.vanagon_project
+      Rake::Task["pl:ship_nuget"].execute if Pkg::Config.vanagon_project
       Rake::Task["pl:ship_tar"].execute if Pkg::Config.build_tar
       Rake::Task["pl:ship_svr4"].execute if Pkg::Config.vanagon_project
       Rake::Task["pl:ship_p5p"].execute if Pkg::Config.build_ips || Pkg::Config.vanagon_project
@@ -281,4 +292,3 @@ namespace :pl do
     end
   end
 end
-
--- a/lib/packaging.rb
+++ b/lib/packaging.rb
@@ -12,6 +12,7 @@ module Pkg
   require 'packaging/rpm'
   require 'packaging/osx'
   require 'packaging/ips'
+  require 'packaging/nuget'
 
   # Load configuration defaults
   Pkg::Config.load_defaults
@@ -22,4 +23,3 @@ module Pkg
   Pkg::Config.issue_reassignments
   Pkg::Config.issue_deprecations
 end
-
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -84,6 +84,8 @@ module Pkg::Params
                   :msi_name,
                   :name,
                   :notify,
+                  :nuget_host,
+                  :nuget_repo_path,
                   :project,
                   :origversion,
                   :osx_build_host,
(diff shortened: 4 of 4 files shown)


tasks/ship.rake
```diff
@@ -189,6 +189,31 @@ namespace :pl do
...
 
+  if Pkg::Config.vanagon_project
+    desc "ship nuget patckages to #{Pkg::Config.nuget_host}"
+    task :ship_nupkg => 'pl:fetch' do
+      packages = Dir['pkg/windows/**/*.nupkg']
+      if packages.empty?
+        STDOUT.puts "There aren't any nuget packages in pkg/windows. Maybe something went wrong"
+      else
+        packages.each do | pkg |
+          projname , version = File.basename(pkg).match(/^(.+)\.(.+)$/).captures
```
jcoconnor: Note - this regexp still needs some work as the version is just the final filename.

tasks/ship.rake
```diff
@@ -189,6 +189,31 @@ namespace :pl do
...
+          puts "Working on file #{pkg}, Project=#{projname}, Version=#{version}"
+
+          #curl -v -H 'Authorization: Basic #{auth}' --upload-file pkg/windows/x64/pl-boost.1.58.0.1.nupkg http://nexus.delivery.puppetlabs.net:8081/content/repositories/#{repo_name}/pl-boost/1.58.0.1/pl-boost.1.58.0.1.nupkg
+          job_url = "http://#{Pkg::Config.nuget_host}:8081/content/repositories/#{Pkg::Config.nuget_repo_path}/pl-boost/1.58.0.1/pl-boost.1.58.0.1.nupkg"
+          # Also need to use authorisation flag.
```
jcoconnor: As the comment says - authorisation flags still need to be added.

ferventcoder:
Does this supersede #471 ?

jcoconnor:
Yes

tasks/ship.rake
```diff
@@ -189,6 +189,31 @@ namespace :pl do
     end
   end
 
+  if Pkg::Config.vanagon_project
```
melissa: We shouldn't need this line since we already have https://github.com/puppetlabs/packaging/pull/473/files#diff-02259d8a37d5ca4f20dfe072004f5b82R226

tasks/ship.rake
```diff
@@ -189,6 +189,31 @@ namespace :pl do
...
+    desc "ship nuget patckages to #{Pkg::Config.nuget_host}"
+    task :ship_nupkg => 'pl:fetch' do
+      packages = Dir['pkg/windows/**/*.nupkg']
+      if packages.empty?
+        STDOUT.puts "There aren't any nuget packages in pkg/windows. Maybe something went wrong"
+      else
+        packages.each do | pkg |
+          projname , version = File.basename(pkg).match(/^((?:\-|\w)+)\.((?:\d|\.)+)\.\w+$/).captures
```
melissa: `name, version = File.basename(pkg).match(/^(.*)-([\d+\.]+)\.nupkg$/).captures` was what I came up with. I'm not sure if Travis would be happier with that or not

melissa: Also, new package name format will be `pl-boost-x64-1.58.0.1.nugpkg`

shermdog:
Did you intend to remove the newlines at the end of the files?

ferventcoder:
From earlier, I mentioned we may need some documentation surrounding this to ensure that folks understand that we are shipping nuget packages in a different manner than normally intended (basically bypassing the entire push command). Mostly because we can easily use curl and post to a particular location, versus trying to bring in mono and choco just to be able to use the push command. Just a little documentation somewhere :)

jcoconnor:
@shermdog I'm looking at the newline problem - seems to be a side effect of Atom - will try and correct, but it doesn't seem to cause any rubocop failures, so not sure if its a problem.

jcoconnor:
@ferventcoder Does this imply we should be using choco and mono to do the shipping ?

melissa:
@jcoconnor re/ @ferventcoder's comment, this does not imply we should be using choco or mono to ship. We could, but we made the decision to allow users to ship from a non-windows box, and a non-specially souped up linux/mac system. The curl call allows the maximum flexibility

jcoconnor:
@melissa, @ferventcoder Updated documentation in separate commit.

ferventcoder:
@jcoconnor can you push that other commit please?

jcoconnor:
@ferventcoder Have actually squashed them at @melissa's suggestion and re-pushed - you should see the documentation there now.

lib/packaging/nuget.rb
```diff
@@ -0,0 +1,36 @@
+module Pkg::Nuget
+  class << self
+    def ship(packages)
+      #
+      # Support shipping of Nuget style packages to a nexus based nuget feed
+      # Using curl to submit the packages rather than windows based choco/mono.
+      # This approach gives more flexibility and fits in with the current Puppet
+      # release automation practices using linux/mac systems.
```
ferventcoder: I take it this is the documentation. :)

jcoconnor: That would be it - does it suffice ?

lib/packaging/nuget.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      # Retrieve password without revealing it
+      puts "Obtaining credentials to ship to nuget repo on #{Pkg::Config.nuget_host}"
+      print "Username please: "
+      username = Pkg::Util.get_input(true)
+      print "Password please: "
+      password = Pkg::Util.get_input(false)
+      authentication = Pkg::Util.base64_encode("#{username}:#{password}")
+
+      uri = "http://#{Pkg::Config.nuget_host}:8081/content/repositories/#{Pkg::Config.nuget_repo_path}"
```
melissa: We might actually want to format this differently.

`uri = "http://#{Pkg::Config.nuget_host}:#{Pkg::Config.nuget_repo_path}"`

And then in our build-data repo, or env variable, or whatever, we could set
`nuget_host: nexus.delivery.puppetlabs.net`
`nuget_repo_path: 8081/content/repositories/temp-build-tools`

melissa: Also, I'm wondering about automatically including the `http://` in the uri. It's probably safe to do that, since generally we don't include that information in out host definitions. But what happens if we have `http://http://`? Would it be safer to not specify the `http://`? Does curl work when we don't specify the protocol? Would that be bad if we didn't specify the protocol?

melissa:
:+1: pending questions above.

Tested with `NUGET_HOST=nexus.delivery.puppetlabs.net NUGET_REPO=temp-build-tools rake pl:jenkins:uber_ship`

lib/packaging/nuget.rb
jcoconnor: @melissa Suggest we test if the protocol isn't there and put in http:// if not present.

ON the nugget_repo_path: comment above - "8081" is the port number not part of the path, so if we are breaking this up, we should give it its own "port_number" parameter.

melissa:
Once this is merged, we'll want to add those variables to https://github.com/puppetlabs/build-data/blob/pl-build-tools/builder_data.yaml

lib/packaging/nuget.rb
```diff
@@ -0,0 +1,36 @@
...
+      fail "NUGET_HOST is not defined" if Pkg::Config.nuget_host.empty?
+      fail "NUGET_REPO is not defined" if Pkg::Config.nuget_repo_path.to_s.empty?
+
+      # Retrieve password without revealing it
+      puts "Obtaining credentials to ship to nuget repo on #{Pkg::Config.nuget_host}"
+      print "Username please: "
+      username = Pkg::Util.get_input(true)
+      print "Password please: "
+      password = Pkg::Util.get_input(false)
+      authentication = Pkg::Util.base64_encode("#{username}:#{password}")
```
underscorgan: I'm not super familiar with nexus/nuget, but is this the best way to do authentication? Does it support authentication via ssh key or token or something that could be stored in environment variables?

lib/packaging/nuget.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      uri = "http://#{Pkg::Config.nuget_host}:8081/content/repositories/#{Pkg::Config.nuget_repo_path}"
+      form_data = ["-H 'Authorization: Basic #{authentication}'", "-f"]
+      packages.each do | pkg |
```
underscorgan: [nitpick, nonblocking] I think we tend to not have spaces between the `|` and the variable.

lib/packaging/nuget.rb
jcoconnor: @underscorgan Agree with you on that and intend to ticket it for further work. There is a token based system available with nexus that we don't understand yet.

heathseals:
@jcoconnor does it work?  other than the prompts for creds, it looks good to me.  :+1:

lib/packaging/nuget.rb
jcoconnor: Fixed

lib/packaging/nuget.rb
jcoconnor: Created RE-6072 for the key/token based authentication.

lib/packaging/nuget.rb
```diff
@@ -0,0 +1,36 @@
...
+      # Using curl to submit the packages rather than windows based choco/mono.
+      # This approach gives more flexibility and fits in with the current Puppet
+      # release automation practices using linux/mac systems.
+
+      # Sanity checks
+      fail "NUGET_HOST is not defined" if Pkg::Config.nuget_host.empty?
+      fail "NUGET_REPO is not defined" if Pkg::Config.nuget_repo_path.to_s.empty?
+
+      # Retrieve password without revealing it
+      puts "Obtaining credentials to ship to nuget repo on #{Pkg::Config.nuget_host}"
```
melissa: Oh, can we change this to `puts "Obtaining credentials to ship to nuget feed #{Pkg::Config.nuget_repo_path} on #{Pkg::Config.nuget_host}"` for additional clarity on where these packages are going?

lib/packaging/nuget.rb
jcoconnor: @ferventcoder @shrug @mckern Could we get your opinions @melissa 's note above.
Tx - John

heathseals: having the port in the path looks kind of weird.  maybe something like this:

`uri = "#{Pkg::Config.nuget_host}/#{Pkg::Config.nuget_repo_path}"`
 and then the data :
`nuget_host: http://nexus.delivery.puppetlabs.net:8081`
`nuget_repo_path: content/repositories/temp-build-tools`

melissa: :+1: that sounds perfect

jcoconnor:
@melissa @heathseals pushed up minor change with your suggestions on the hostname

melissa:
:+1: tested with `NUGET_HOST=nexus.delivery.puppetlabs.net:8081 NUGET_REPO=/content/repositories/nuget-pl-build-tools/ rake pl:jenkins:uber_ship`

ferventcoder:
:+1:

PR #474, opened 2015-11-24
(RE-6007) Update puppet agent release task

joshcooper:
Previously some of the tickets contains both PM and DEV tasks. This
commit breaks those out so they can be assigned separately. In addition
makes the following changes:

* Notify everyone at the start of the release

* If we skip a z release, make sure any tickets with that fixVersion
  are reassigned accordingly and the version is closed in Jira

* Add cpp-pcp-client to the list of components to reconcile commits

* Clarify that git versions use the tag (4.1.0), whereas Jira
  fixVersions use the project name and version (PUP 4.1.0)

* Escape `{GPG key}` so it renders correctly

* Kick off manual pipeline after tagging puppet-agent

* bundle install & exec when executing gem smoke test

* Disable PE autopromotion when shipping

* Also notify dev@puppetlabs.com with an announcement

* Merge stable to master after the release, resolving conflicts as
  described in the README

* Assign go/no-go, reconcile commits, and closing resolved tickets to PM

* Generate JQL-based `fixed in` and `introduced in` queries for each component


tasks/puppet-agent-release-tickets.rake
```diff
@@ -1,11 +1,32 @@
...
+def build_queries(vars, label)
+  queries = {}
+
+  queries[:all]          = build_query(vars, "#{label} in ('puppet-agent #{vars[:puppet_agent_release]}', 'PUP #{vars[:puppet_release]}', 'FACT #{vars[:facter_release]}', 'HI #{vars[:hiera_release]}', 'MCO #{vars[:mcollective_release]}', 'pxp-agent #{vars[:pcp_release]}')")
+  queries[:puppet_agent] = build_query(vars, "#{label} = 'puppet-agent #{vars[:puppet_agent_release]}'")
+  queries[:puppet]       = build_query(vars, "#{label} = 'PUP #{vars[:puppet_release]}'")
+  queries[:facter]       = build_query(vars, "#{label} = 'FACT #{vars[:facter_release]}'")
+  queries[:hiera]        = build_query(vars, "#{label} = 'HI #{vars[:hiera_release]}'")
+  queries[:mcollective]  = build_query(vars, "#{label} = 'MCO #{vars[:mcollective_release]}'")
+  queries[:pcp]          = build_query(vars, "#{label} = 'pxp-agent #{vars[:pcp_release]}'")
```
kylog: It would make more sense to me if the variables here were named based on `pxp_agent` rather than `pcp`. I just did one of these tickets (for puppet-agent 1.3.2) and it was confusing to me what was meant by "PCP_RELEASE".

I.e. this line would become:
```
queries[:pxp_agent]    = build_query(vars, "#{label} = 'pxp-agent #{vars[:pxp_agent_release]}'")
```

and there would be complementary changes below.

tasks/puppet-agent-release-tickets.rake
```diff
@@ -39,41 +68,33 @@ end
...
-  * Any tickets which are not resolved should be brought to the attention of the appropriate scrum team so they can be resolved for the release.
+3) Create a public pair of queries for inclusion in the release notes/announcement. These allow easy tracking as new bugs come in for a particular version and allow everyone to see the list of changes slated for the next release (Paste their URLs into the "Prepare long form release notes and short form release story" ticket).
 
+  {{affectedVersion in ("puppet-agent #{vars[:puppet_agent_release]}", "PUP #{vars[:puppet_release]}", "FACT #{vars[:facter_release]}", "HI #{vars[:hiera_release]}", "MCO #{vars[:mcollective_release]}", "pxp-agent #{vars[:pcp_release]}")}}, Save as "Introduced in puppet-agent #{vars[:puppet_agent_release]}", click Details, add permission for Everyone
+   {{fixVersion in ("puppet-agent #{vars[:puppet_agent_release]}", "PUP #{vars[:puppet_release]}", "FACT #{vars[:facter_release]}", "HI #{vars[:hiera_release]}", "MCO #{vars[:mcollective_release]}", "pxp-agent #{vars[:pcp_release]}")}}, Save as "Fixed in puppet-agent #{vars[:puppet_agent_release]}", click Details, add permission for Everyone
```
kylog: Perhaps add a comment that the fixVersion one may need to be edited down to remove components that haven't changed. 

I.e. it just occurred to me that with a release (like 1.3.2 or 1.3.1) that includes several unchanged components, those components should show up in the affectedVersion filter but not the fixVersion filter. And I don't know if it's really worth parameterizing that on the input to this ticket task.

tasks/puppet-agent-release-tickets.rake
```diff
@@ -39,41 +68,33 @@ end
...
+1) Ensure that all tickets targeted at this release for all components are resolved.
+  * Use the following filter to view all tickets for all components for this release: #{vars[:fixed_in][:all]}
+  * Any tickets which are not resolved should be brought to the attention of the appropriate scrum team so they can be resolved for the release.
+  * Verify tickets have release notes. Ping devs to update tickets with missing release notes.
+
+2) Use the [ticketmatch|https://github.com/puppetlabs/ticketmatch] script to ensure all tickets referenced in the commit log have a bug targeted at the release, and ensure all tickets targeted at the release have a corresponding commit. This needs to be done for each component, including puppet-agent itself. This process may reveal tickets that need to be resolved, so this process may need to be repeated a few times.
 
   * cd ~/work
   * git clone https://github.com/puppetlabs/ticketmatch
```
kylog: You may need to `gem install highline` at this point.

tasks/puppet-agent-release-tickets.rake
```diff
@@ -1,11 +1,32 @@
...
+  queries[:mcollective]  = build_query(vars, "#{label} = 'MCO #{vars[:mcollective_release]}'")
+  queries[:pcp]          = build_query(vars, "#{label} = 'pxp-agent #{vars[:pcp_release]}'")
+
+  queries
+end
+
 def get_agent_release_ticket_vars
   vars = {}
 
+  vars[:site]            = Pkg::Util.get_var("JIRA_INSTANCE") || "https://tickets.puppetlabs.com"
```
melissa: This should already be queried for all ticket creation instances at https://github.com/puppetlabs/packaging/blob/master/lib/packaging/util/jira.rb#L87

joshcooper: Good question. I need `vars[:site]` to be populated before calling `build_queries`, so that the links refer to the correct site. I could move the call to `get_auth_vars` earlier, but I'm confused by the comment in `Jira authentication - do this after validating other params` on line 55, since AFAIK calling `get_auth_vars` doesn't connect to jira. I think authentication doesn't actually happen until we try to validate the vars and call:

```
  jira.project vars[:project]
```

So I'm thinking I can move the call to `get_auth_vars` up and delete the comment.

tasks/puppet-agent-release-tickets.rake
joshcooper: Yep, agree that PCP here really means pxp-agent. I'll update that.

Related question, should we be tracking cpp-pcp-client as a separate component? On one hand, it is a library used by pxp-agent, so isn't user-facing. On the other hand, it is a distinct component of puppet-agent, with its own fix versions, e.g. `cpp-pcp-client 1.0.1`. So for example, if you want to know what's fixed in the release, or what needs to be fixed before we can release, it would be useful to have those appear in the query.

kylog: Good point about cpp-pcp-client. I agree it should show up in the various jira queries, etc, so I'd say we should add that as a version to specify when the tickets are created.

As an alternative, since it's not promoted, we could potentially *derive* the version bump from the puppet-agent repo, but that would be more work and may not warrant the effort. If we end up with more libraries like this (internal, non-promoted), it might be worth revisiting this approach then.

tasks/puppet-agent-release-tickets.rake
joshcooper: Added in https://github.com/joshcooper/packaging/commit/607124c93d680d992bf0c8ed95680b3319c235f6

tasks/puppet-agent-release-tickets.rake
joshcooper: Added in https://github.com/puppetlabs/packaging/commit/5fc6690981b96d2ba739cfea0c95a78c4038fdeb

tasks/puppet-agent-release-tickets.rake
joshcooper: Added in https://github.com/joshcooper/packaging/commit/e82f9e8d33692edf2e5a091251245f6f64c09e7c

tasks/puppet-agent-release-tickets.rake
joshcooper: Updated in https://github.com/joshcooper/packaging/commit/de8427ccbf4e59498437872d395db78180713fd2

joshcooper:
The latest result can be viewed in https://jira1-test.ops.puppetlabs.net/browse/PA-199, generated using

```
$ rake pl:puppet_agent_release_tickets BUILDER=melissa DEVELOPER=kylo WRITER=nick.fagerlund OWNER=eric.sorenson TESTER=john.duarte PROJECT_MANAGER=steven.barlow PUPPET_AGENT_RELEASE=1.3.1 PUPPET_RELEASE=4.3.1 FACTER_RELEASE=3.1.2 HIERA_RELEASE=3.0.5 MCOLLECTIVE_RELEASE=2.8.6 PXP_AGENT_RELEASE=1.0.1 CPP_PCP_CLIENT_RELEASE=1.0.1 DATE=2015-12-31 JIRA_USER=josh JIRA_INSTANCE=https://jira1-test.ops.puppetlabs.net
```

underscorgan:
:+1: 

is this good to merge now?

kylog:
:+1:

PR #475, opened 2015-11-25
(RE-5619) Allow shipping to PE feature repos

shrug:
For PE, we want to allow the promotion and use of arbitrary
test/dev packages into feature branches of PE, that don't pollute
the mainline PE branches and repos.

This adds logic to enable shipping to the feature repos for a PE branch,
and during a normal ship/promotion, links the shipped package into the
feature repos (so that mainline packages are immediately available to the
feature branches).

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -344,6 +344,13 @@ module Pkg
           end
         end
       end
+
+      def yum_target_path(feature_branch = false)
+        if feature_branch || Pkg::Config.pe_feature_branch
+          return "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/feature/repos/"
+        end
+        "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/repos/"
+      end
     end
   end
 end
--- a/tasks/pe_ship.rake
+++ b/tasks/pe_ship.rake
@@ -2,13 +2,17 @@ if Pkg::Config.build_pe
   namespace :pe do
     desc "ship PE rpms to #{Pkg::Config.yum_host}"
     task :ship_rpms => "pl:fetch" do
+      puts "Shipping packages to #{Pkg::Config.yum_target_path}"
       Pkg::Util::File.empty_dir?("pkg/pe/rpm") and fail "The 'pkg/pe/rpm' directory has no packages. Did you run rake pe:deb?"
-      target_path = ENV['YUM_REPO'] ? ENV['YUM_REPO'] : "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/repos/"
       Pkg::Util::Execution.retry_on_fail(:times => 3) do
-        Pkg::Util::Net.rsync_to('pkg/pe/rpm/', Pkg::Config.yum_host, target_path)
+        Pkg::Util::Net.rsync_to('pkg/pe/rpm/', Pkg::Config.yum_host, Pkg::Config.yum_target_path)
       end
       if Pkg::Config.team == 'release'
-        Rake::Task["pe:remote:update_yum_repo"].invoke
+
+        # If this is not a feature branch, we need to link the shipped packages into the feature repos,
+        # then update their metadata as well.
+        Pkg::Util::RakeUtils.invoke_task("pe:remote:link_shipped_rpms_to_feature_repo") unless Pkg::Config.pe_feature_branch
+        Pkg::Util::RakeUtils.invoke_task("pe:remote:update_yum_repo")
       end
     end
 
@@ -129,17 +133,23 @@ if Pkg::Config.build_pe
     namespace :remote do
       desc "Update remote rpm repodata for PE on #{Pkg::Config.yum_host}"
       task :update_yum_repo => "pl:fetch" do
-        repo_base_path = File.join(Pkg::Config.yum_repo_path, Pkg::Config.pe_version, "repos")
 
-        # This entire command is going to be passed across SSH, but it's unwieldy on a
-        # single line. By breaking it into a series of concatenated strings, we can maintain
-        # a semblance of formatting and structure (nevermind readability).
-        command  = %(for dir in #{repo_base_path}/{#{rpm_family_and_version.join(",")}}-*; do)
-        command += %(  sudo createrepo --checksum=sha --quiet --database --update $dir; )
-        command += %(done; )
-        command += %(sync)
+        # Paths to the repos. If this is *not* a feature branch, we need to also
+        # Update the feature branch repos in addition to the primary repos.
+        repo_base_paths = Array[Pkg::Config.yum_target_path]
+        repo_base_paths << Pkg::Config.yum_target_path(true) unless Pkg::Config.pe_feature_branch
 
-        Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.yum_host, command)
+        repo_base_paths.each do |repo_base_path|
+          # This entire command is going to be passed across SSH, but it's unwieldy on a
+          # single line. By breaking it into a series of concatenated strings, we can maintain
+          # a semblance of formatting and structure (nevermind readability).
+          command  = %(for dir in #{repo_base_path}/{#{rpm_family_and_version.join(",")}}-*; do)
+          command += %(  sudo createrepo --checksum=sha --quiet --database --update $dir; )
+          command += %(done; )
+          command += %(sync)
+
+          Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.yum_host, command)
+        end
       end
 
       #   the repsimple application is a small wrapper around reprepro, the purpose of
@@ -187,6 +197,21 @@ if Pkg::Config.build_pe
         puts "Cleaning up apt repo 'incoming' dir on #{Pkg::Config.apt_host}"
         Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "rm -r #{incoming_dir}")
 
+      end
+
+      # Throw another tire on the fire
+      desc "Remotely link shipped packages into feature repo on #{Pkg::Config.yum_host}"
+      task :link_shipped_rpms_to_feature_repo => "pl:fetch" do
+        next if Pkg::Config.pe_feature_branch
+        repo_base_path = Pkg::Config.yum_target_path
+        feature_repo_path = Pkg::Config.yum_target_path(true)
+        pkgs = FileList['pkg/pe/rpm/**/*.rpm'].select { |path| path.gsub!('pkg/pe/rpm/', '') }
+        command  = %(for pkg in #{pkgs.join(' ')}; do)
+        command += %(  sudo ln -f "#{repo_base_path}/$( dirname ${pkg} )/$( basename ${pkg} )" "#{feature_repo_path}/$( dirname ${pkg} )/" ; )
+        command += %(done; )
+        command += %(sync)
+
+        Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.yum_host, command)
       end
     end
   end
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -99,6 +99,7 @@ module Pkg::Params
                   :packaging_root,
                   :packaging_url,
                   :pbuild_conf,
+                  :pe_feature_branch,
                   :pe_name,
                   :pe_platforms,
                   :pe_version,
@@ -204,7 +205,8 @@ module Pkg::Params
               { :var => :team,                    :envvar => :TEAM },
               { :var => :update_version_file,     :envvar => :NEW_STYLE_PACKAGE },
               { :var => :yum_repo_path,           :envvar => :YUM_REPO },
-              { :var => :yum_host,                :envvar => :YUM_HOST }]
+              { :var => :yum_host,                :envvar => :YUM_HOST },
+              { :var => :pe_feature_branch,       :envvar => :PE_FEATURE_BRANCH }]
   # Default values that are supplied if the user does not supply them
   #
   # usage is the same as above
@@ -223,7 +225,8 @@ module Pkg::Params
               { :var => :ips_signing_cert,        :val => '$IPS_SIGNING_CERT' },
               { :var => :ips_inter_cert,          :val => '$IPS_INTER_CERT' },
               { :var => :ips_root_cert,           :val => '$IPS_ROOT_CERT' },
-              { :var => :ips_signing_key,         :val => '$IPS_SIGNING_KEY' }]
+              { :var => :ips_signing_key,         :val => '$IPS_SIGNING_KEY' },
+              { :var => :pe_feature_branch,       :val => false }]
 
   # These are variables which, over time, we decided to rename or replace. For
   # backwards compatibility, we assign the value of the old/deprecated


shrug:
Related to the ongoing work at https://github.com/puppetlabs/enterprise-dist/pull/2815

underscorgan:
I may be missing something here, but my main questions are around what happens if there are multiple feature branches? Won't this make all of the yum packages live in the same directory / pollute other feature branches' packages?

stahnma:
@underscorgan the plan is that is each composition happens based on what's in the json file and not just what's laying around on disk. So if there are 5 different versions of a package, the correct one gets selected per feature branch.

underscorgan:
@stahnma cool, thanks!

tasks/pe_ship.rake
```diff
@@ -3,12 +3,25 @@ if Pkg::Config.build_pe
...
       end
       if Pkg::Config.team == 'release'
         Rake::Task["pe:remote:update_yum_repo"].invoke
+
+        # If this is not a feature branch, we need to link the shipped packages into the feature repos,
+        # then update their metadata as well.
+        unless Pkg::Config.pe_feature_branch
+          Rake::Task["pe:remote:link_shipped_rpms_to_feature_repo"].invoke
+          # Update repo metadata for feature branch
+          Pkg::Config.pe_feature_branch = true
```
underscorgan: By not resetting this after L23 won't this break update_yum_repos when that's called later?

shrug: Hmm... good catch, it might. Better safe than sorry anyway

shrug: Now reset on line 24

underscorgan:
This looks reasonable :+1:

tasks/pe_ship.rake
```diff
@@ -3,12 +3,26 @@ if Pkg::Config.build_pe
     desc "ship PE rpms to #{Pkg::Config.yum_host}"
     task :ship_rpms => "pl:fetch" do
       Pkg::Util::File.empty_dir?("pkg/pe/rpm") and fail "The 'pkg/pe/rpm' directory has no packages. Did you run rake pe:deb?"
-      target_path = ENV['YUM_REPO'] ? ENV['YUM_REPO'] : "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/repos/"
+      if Pkg::Config.pe_feature_branch
+        target_path = ENV['YUM_REPO'] ? ENV['YUM_REPO'] : "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/feature/repos/"
```
mckern: Is this bit of configuration here the sort of thing that should be defined in `lib/packaging/config.rb`?

tasks/pe_ship.rake
```diff
@@ -3,12 +3,26 @@ if Pkg::Config.build_pe
...
+
+        # If this is not a feature branch, we need to link the shipped packages into the feature repos,
+        # then update their metadata as well.
+        unless Pkg::Config.pe_feature_branch
+          Rake::Task["pe:remote:link_shipped_rpms_to_feature_repo"].invoke
+          # Update repo metadata for feature branch
+          Pkg::Config.pe_feature_branch = true
+          Rake::Task["pe:remote:update_yum_repo"].invoke
+          Pkg::Config.pe_feature_branch = false
+        end
```
mckern: Isn't this equivalent to

```ruby
Rake::Task["pe:remote:link_shipped_rpms_to_feature_repo"].invoke unless Pkg::Config.pe_feature_branch
Rake::Task["pe:remote:update_yum_repo"].invoke
```

tasks/pe_ship.rake
```diff
@@ -129,7 +143,11 @@ if Pkg::Config.build_pe
     namespace :remote do
       desc "Update remote rpm repodata for PE on #{Pkg::Config.yum_host}"
       task :update_yum_repo => "pl:fetch" do
-        repo_base_path = File.join(Pkg::Config.yum_repo_path, Pkg::Config.pe_version, "repos")
+        if Pkg::Config.pe_feature_branch
+          repo_base_path = File.join(Pkg::Config.yum_repo_path, Pkg::Config.pe_version, "feature", "repos")
+        else
+          repo_base_path = File.join(Pkg::Config.yum_repo_path, Pkg::Config.pe_version, "repos")
+        end
```
mckern: This is the second time I've seen a bit of configuration linked wrapped in `if Pkg::Config.pe_feature_branch`; should we make these values into functions in the library and let the function determine the value returned for these?

tasks/pe_ship.rake
```diff
@@ -188,6 +206,23 @@ if Pkg::Config.build_pe
         Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "rm -r #{incoming_dir}")
 
       end
+
+      # Throw another tire on the fire
+      desc "Remotely link shipped packages into feature repo on #{Pkg::Config.yum_host}"
+      task :link_shipped_rpms_to_feature_repo => "pl:fetch" do
+        unless Pkg::Config.pe_feature_branch
+          next
+        end
```
mckern: Nitpick:

```ruby
next unless Pkg::Config.pe_feature_branch
```

tasks/pe_ship.rake
shrug: Probably. I was trying to avoid much general refactoring for this PR, but honestly all of these rake tasks need significant love.

tasks/pe_ship.rake
shrug: Almost. It's also flipping the config so the call to update_yum_repo updates the feature repos after linking the files

tasks/pe_ship.rake
mckern: Agreed re: general love. I'd like to see new code go in the right place when it's added but you could probably persuade me that this is OK "for now" if there's a ticket to move this code into the library and we can pull it into a sprint soonish.

tasks/pe_ship.rake
mckern:
```ruby
Rake::Task["pe:remote:link_shipped_rpms_to_feature_repo"].invoke unless Pkg::Config.pe_feature_branch
Rake::Task["pe:remote:update_yum_repo"].invoke if Pkg::Config.pe_feature_branch
```

???

shrug: Sorry, I'm unclear. 
If we're not shipping to a feature repo, then we need to link the shipped RPMs to the repo, and then tell update_yum_repo to also go run createrepo in the feature repos (it has already been run in the main repo), so flip the config flag for feature to 'true', run it, then flip it back to 'false'. update_yum_repo determines it's path from that config variable on line 146 (discussed below!)

haus: What if pe:remote:update_yum_repo were a method that operated on a directory and could be called more than once?

mckern: Configuration flipping like this is a definite code smell, which is why I'm trying to find another way to solve this problem. @haus' suggestion is an interesting one.

lib/packaging/config.rb
```diff
@@ -344,6 +344,16 @@ def rpm_build_targets
...
+
+      def yum_target_path(feature=false)
+        if ENV['YUM_REPO']
+          return ENV['YUM_REPO']
+        elsif feature == true || Pkg::Config.pe_feature_branch
+          return "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/feature/repos/"
+        else
+          return "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/repos/"
+        end
+      end
```
mckern: This is a good start to what we talked about last week. Here are some notes:

- `ENV['YUM_REPO']` is picked up automatically by the accessor defined by `lib/packaging/config/params.rb`, so you can I think you should be able to access it with `#yum_repo_path` instead of calling the environment variable directly.
- `feature` has a truth value on its own, so it doesn't need to be compared with `==`.

Here's some notes on a gently edited version of this:

```ruby
def yum_target_path(feature_branch = false)
  # Generally favor one-line "thing if" style if the conditions are short
  # and there's no strict need for cascading if/else blocks
  return Pkg::Config.yum_repo_path if Pkg::Config.yum_repo_path

  # We should probably move "is this a feature branch?" into
  # a method, but we don't have to do that right now. Also,
  # I'd suggest putting this all on one line in the "thing if" style
  # except that this line becomes really long. Wooo strings!
  if feature_branch || Pkg::Config.pe_feature_branch
    return "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/feature/repos/"
  end

  # We don't need an explicit return if this is the final/default value
  # as the last value used is implicitly returned
  "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/repos/"
end
```

tasks/pe_ship.rake
```diff
@@ -129,17 +132,23 @@ if Pkg::Config.build_pe
...
-        command  = %(for dir in #{repo_base_path}/{#{rpm_family_and_version.join(",")}}-*; do)
-        command += %(  sudo createrepo --checksum=sha --quiet --database --update $dir; )
-        command += %(done; )
-        command += %(sync)
 
-        Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.yum_host, command)
+        # Paths to the repos. If this is *not* a feature branch, we need to also
+        # Update the feature branch repos in addition to the primary repos.
+        repo_base_paths = [yum_target_path]
+        repo_base_paths << yum_target_path(feature=true) unless Pkg::Config.pe_feature_branch
```
mckern: `yum_target_path(true)` since we're not using keyword arguments

tasks/pe_ship.rake
```diff
@@ -188,6 +197,21 @@ if Pkg::Config.build_pe
         Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "rm -r #{incoming_dir}")
 
       end
+
+      # Throw another tire on the fire
+      desc "Remotely link shipped packages into feature repo on #{Pkg::Config.yum_host}"
+      task :link_shipped_rpms_to_feature_repo => "pl:fetch" do
+        next unless Pkg::Config.pe_feature_branch
+        repo_base_path = yum_target_path
+        feature_repo_path = yum_target_path(feature=true)
```
mckern: `yum_target_path(true)`

tasks/pe_ship.rake
```diff
@@ -129,17 +132,23 @@ if Pkg::Config.build_pe
...
+        # Update the feature branch repos in addition to the primary repos.
+        repo_base_paths = [yum_target_path]
+        repo_base_paths << yum_target_path(feature=true) unless Pkg::Config.pe_feature_branch
+
+        repo_base_paths.each do |repo_base_path|
+          # This entire command is going to be passed across SSH, but it's unwieldy on a
+          # single line. By breaking it into a series of concatenated strings, we can maintain
+          # a semblance of formatting and structure (nevermind readability).
+          command  = %(for dir in #{repo_base_path}/{#{rpm_family_and_version.join(",")}}-*; do)
+          command += %(  sudo createrepo --checksum=sha --quiet --database --update $dir; )
```
mckern: We're intentionally summing all yum meta with sha1?

tasks/pe_ship.rake
```diff
@@ -188,6 +197,21 @@ if Pkg::Config.build_pe
         Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "rm -r #{incoming_dir}")
 
       end
+
+      # Throw another tire on the fire
```
mckern: tire_fire

tasks/pe_ship.rake
```diff
@@ -188,6 +197,21 @@ if Pkg::Config.build_pe
...
+      desc "Remotely link shipped packages into feature repo on #{Pkg::Config.yum_host}"
+      task :link_shipped_rpms_to_feature_repo => "pl:fetch" do
+        next unless Pkg::Config.pe_feature_branch
+        repo_base_path = yum_target_path
+        feature_repo_path = yum_target_path(feature=true)
+        pkgs = FileList['pkg/pe/rpm/**/*.rpm']
```
mckern: Uuuuugh Rake's `FileList` class. I get why we're using it here, I just hates me some Rake these days.

tasks/pe_ship.rake
```diff
@@ -188,6 +197,21 @@ if Pkg::Config.build_pe
...
+
+      # Throw another tire on the fire
+      desc "Remotely link shipped packages into feature repo on #{Pkg::Config.yum_host}"
+      task :link_shipped_rpms_to_feature_repo => "pl:fetch" do
+        next if Pkg::Config.pe_feature_branch
+        repo_base_path = yum_target_path
+        feature_repo_path = yum_target_path(true)
+        pkgs = FileList['pkg/pe/rpm/**/*.rpm']
+        command  = %(for pkg in #{pkgs.join(' ')}; do)
+        command += %(  sudo ln -f #{repo_base_path}/$( dirname ${pkg} )/$( basename ${pkg} ) #{feature_repo_path}/$( dirname ${pkg} )/ ; )
```
mckern: Modest nitpick, but should the strings constructed here be wrapped in double quotes before they're sent?

```
sudo ln -f "#{repo_base_path}/$( dirname ${pkg} )/$( basename ${pkg} )" "#{feature_repo_path}/$( dirname ${pkg} )/"
```

lib/packaging/config.rb
```diff
@@ -344,6 +344,13 @@ def rpm_build_targets
           end
         end
       end
+
+      def yum_target_path(feature_branch=false)
```
mckern: It's a nitpick but I think Rubocop will flag it: spaces after assignment (`feature_branch = false`). Totes not blocking, but I don't know if it'll choke Travis.

tasks/pe_ship.rake
```diff
@@ -3,11 +3,14 @@ if Pkg::Config.build_pe
...
-      target_path = ENV['YUM_REPO'] ? ENV['YUM_REPO'] : "#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/repos/"
       Pkg::Util::Execution.retry_on_fail(:times => 3) do
-        Pkg::Util::Net.rsync_to('pkg/pe/rpm/', Pkg::Config.yum_host, target_path)
+        Pkg::Util::Net.rsync_to('pkg/pe/rpm/', Pkg::Config.yum_host, yum_target_path)
       end
       if Pkg::Config.team == 'release'
+
+        # If this is not a feature branch, we need to link the shipped packages into the feature repos,
+        # then update their metadata as well.
+        Rake::Task["pe:remote:link_shipped_rpms_to_feature_repo"].invoke unless Pkg::Config.pe_feature_branch
```
mckern: @haus mentioned another/better way to run a Rake task from within another Rake task than directly calling `#invoke` on it but I don't remember what it was. Maybe he'll remind me?

tasks/pe_ship.rake
```diff
@@ -129,17 +132,23 @@ if Pkg::Config.build_pe
...
-        # a semblance of formatting and structure (nevermind readability).
-        command  = %(for dir in #{repo_base_path}/{#{rpm_family_and_version.join(",")}}-*; do)
-        command += %(  sudo createrepo --checksum=sha --quiet --database --update $dir; )
-        command += %(done; )
-        command += %(sync)
 
-        Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.yum_host, command)
+        # Paths to the repos. If this is *not* a feature branch, we need to also
+        # Update the feature branch repos in addition to the primary repos.
+        repo_base_paths = [yum_target_path]
```
mckern: Nonblocking nitpick, but would you consider `Array[yum_target_path]` more readable?

tasks/pe_ship.rake
shrug: Yep, I didn't change that. It's probably safe to update that or let creatrepo default to sha256 now? I don't think we have any instances of old yum using these repos anymore, but that's probably out of scope for this PR

tasks/pe_ship.rake
haus: `Pkg::Util::RakeUtils.invoke_task(task_name)`

tasks/pe_ship.rake
```diff
@@ -129,17 +132,23 @@ if Pkg::Config.build_pe
...
+        repo_base_paths << yum_target_path(true) unless Pkg::Config.pe_feature_branch
```
mckern: So if this is a feature branch, then repo_base_paths should include both `#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/feature/repos/` and `#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/repos/`, not just `#{Pkg::Config.yum_repo_path}/#{Pkg::Config.pe_version}/feature/repos/`?

tasks/pe_ship.rake
mckern: Thanks

tasks/pe_ship.rake
shrug: Nope, if it's *not* a feature branch. If it is, we just update the feature repo metadata, if it's not a feature branch we need to update both.

mckern:
I think I'm done manually linting this. Have you got any test runs or output from it?

shrug:
http://jenkins-compose.delivery.puppetlabs.net/view/Promotion/job/Package-Promotion-Test/24/

underscorgan:
the changes look clean, test run also seems to be about what i'd expect.

:+1:

mckern:
online-betting

PR #476, opened 2015-12-02
(maint) Remove HuaweiOS from RPM tasks

ScottGarman:
Platform has changed from RPM to DEB.

I'm not seeing any need to specify cumulus as a DEB platform, so this PR only removes huawei from the RPM tasks. Please let me know if I need to explicitly add huawei to the DEB tasks.


shrug:
:+1: , I'd have to dig into the deb tasks to see what's required there

PR #477, opened 2015-12-04
(RE-5812) Add missing steps for new agent platforms

underscorgan:
New agent platforms need to be added to build_defaults so they can be
appropriately whitelisted for nightlies, and they need to be added to
the S3 and internal agent ship jobs so they can be used by PE.

### DIFF
--- a/tasks/agent-tickets.rake
+++ b/tasks/agent-tickets.rake
@@ -90,6 +90,22 @@ Jenkins jobs should be updated to include the new target depending on which tier
 This is for all puppet-agent jenkins pipelines.
 DOC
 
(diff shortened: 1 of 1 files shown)


underscorgan:
Tested at https://jira1-test.ops.puppetlabs.net/browse/CPR-261

shermdog:
:+1:

stahnma:
@underscorgan should I see a ticket for s3 stuff in that epic?

underscorgan:
@stahnma https://jira1-test.ops.puppetlabs.net/browse/RE-6136

PR #478, opened 2015-12-09
(maint) re-add remote:deploy_apt_repo to the uber_ship

underscorgan:
We were running this manually after we accidentally broke the world a
bunch. The past several deployments have worked using this task run
manually, and I'm as confident as ever that this change will not break
anything. So this commit adds remote:deploy_apt_repo back to the
uber_ship.


mckern:
"We were running this manually after ~~we~~ Ryan accidentally broke the world a
bunch."

mckern:
:+1:  because we haven't had a fire in a while and I'm cold.

PR #479, opened 2015-12-09
(maint) Update permissions for the deb deployment

underscorgan:
Manually mangling permissions is the worst. Except for maybe the sticky
bit. The sticky bit is definitely terrible. This stops rsync from
trying to set the group of things on apt.puppetlabs.com and removes the
sticky bit, because rsync does not listen to -s with its chmod.

### DIFF
--- a/lib/packaging/deb/repo.rb
+++ b/lib/packaging/deb/repo.rb
@@ -199,7 +199,8 @@ SignWith: #{Pkg::Config.gpg_key}"
         --update
         --verbose
         --perms
-        --chmod='Dugo-s,Dug=rwx,Do=rx,Fug=rw,Fo=r'
+        --no-group
+        --chmod='Dug=rwx,Do=rx,Fug=rw,Fo=r'
         --exclude='dists/*-*'
         --exclude='pool/*-*'
       )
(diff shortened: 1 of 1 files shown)


lib/packaging/deb/repo.rb
```diff
@@ -199,7 +199,8 @@ def repo_deployment_command(origin_path, destination_path, destination, dryrun =
         --update
         --verbose
         --perms
-        --chmod='Dugo-s,Dug=rwx,Do=rx,Fug=rw,Fo=r'
+        --no-g
```
mckern: `--no-g`?

underscorgan:
@mckern updated.

mckern:
I'm :+1: assuming that we've validated that `--no-group` does the needful.

underscorgan:
```
morgan@burji3:~$ touch /tmp/foo
morgan@burji3:~$ sudo chgrp root /tmp/foo
morgan@burji3:~$ rsync -a --no-group /tmp/foo /tmp/bar
morgan@burji3:~$ ll /tmp/bar
-bash: ll: command not found
morgan@burji3:~$ ls -la /tmp/bar
-rw-r--r-- 1 morgan allstaff 0 Dec  9 15:21 /tmp/bar
morgan@burji3:~$ ls -la /tmp/foo
-rw-r--r-- 1 morgan root 0 Dec  9 15:21 /tmp/foo
morgan@burji3:~$
```

mckern:
works for giant man

PR #480, opened 2015-12-10
(maint) Re-add the Dugo-s to the chmod

underscorgan:
While this does nothing for the sticky bit, everything else fails
without this option. So, add it back so rsync will actually sync.


melissa:
:+1: especially with the comment :)

PR #481, opened 2015-12-10
(RE-5813) Use PA as the parent project for agent platform tickets

underscorgan:
Since the PA project exists we should always use that for platform
additions. We also always want puppet_agent_configuration to be part of
the PA project, and c_toolchain and build tools should be either in PA
or RE depending on whether or not the platform is PE only.

### DIFF
--- a/tasks/agent-tickets.rake
+++ b/tasks/agent-tickets.rake
@@ -206,7 +206,7 @@ DOC
     },
     {
       :short_name   => 'puppet_agent_configuration',
-      :project      => 'RE',
+      :project      => 'PA',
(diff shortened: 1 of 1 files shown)


underscorgan:
Regular platform: https://jira1-test.ops.puppetlabs.net/browse/PA-139
PE only platform: https://jira1-test.ops.puppetlabs.net/browse/PA-147

stahnma:
:+1:

McdonaldSeanp:
:+1:

melissa:
:+1: pending travis. Maybe I should just merge this?

underscorgan:
:rage:

PR #482, opened 2015-12-11
(RE-5631) Update packaging platform hash

underscorgan:
When new platforms are being added, the platform hash in packaing should
be updated to include the new platform so it can be added to whitelists
and blacklists and the like.


underscorgan:
New ticket created https://jira1-test.ops.puppetlabs.net/browse/RE-6194 in epic https://jira1-test.ops.puppetlabs.net/browse/PA-154

stahnma:
:+1:

McdonaldSeanp:
:+1:

PR #483, opened 2015-12-15
(RE-6070) Ship RPMS to YUM_STAGING_SERVER

mckern:
We used to ship our RPM repos directly to our perimeter nodes, which in the new world is sort of not getting the job done anymore. We should be shipping them to a staging server instead. This PR is tracking the work needed to move new RPMs from our local machine to our staging server, and then from our staging server to our production infrastructure.

### DIFF
--- a/lib/packaging/rpm/repo.rb
+++ b/lib/packaging/rpm/repo.rb
@@ -44,8 +44,8 @@ module Pkg::Rpm::Repo
     # @return [String] an rsync command that can be executed on a remote host
     #   to copy local content from that host to a remote node.
     def repo_deployment_command(origin_path, destination_path, destination, dryrun = false)
-      path = Pathname.new(origin_path)
-      dest_path = Pathname.new(destination_path)
+      path = Pathname.new(origin_path).cleanpath
+      dest_path = Pathname.new(destination_path).cleanpath
 
       # You may think "rsync doesn't actually remove the sticky bit, let's
       # remove the Dugo-s from the chmod". However, that will make your rsyncs
--- a/lib/packaging/rpm/repo.rb
+++ b/lib/packaging/rpm/repo.rb
@@ -29,6 +29,69 @@ module Pkg::Rpm::Repo
       cmd << "done "
     end
 
+    # @deprecated this command will die a painful death when we are
+    #   able to sit down with Operations and refactor our distribution infra.
+    #   At a minimum, it should be refactored alongside its Debian counterpart
+    #   into something modestly more generic.
+    #   - Ryan McKern 11/2015
+    #
+    # @param origin_path [String] path for RPM repos on local filesystem
+    # @param destination_path [String] path for RPM repos on remote filesystem
+    # @param destination [String] remote host to send rsynced content to. If
+    #        nil will copy locally
+    # @param dryrun [Boolean] whether or not to use '--dry-run'
+    #
+    # @return [String] an rsync command that can be executed on a remote host
+    #   to copy local content from that host to a remote node.
+    def repo_deployment_command(origin_path, destination_path, destination, dryrun = false)
+      path = Pathname.new(origin_path)
+      dest_path = Pathname.new(destination_path)
+
+      # You may think "rsync doesn't actually remove the sticky bit, let's
+      # remove the Dugo-s from the chmod". However, that will make your rsyncs
+      # fail due to permission errors.
+      options = %w(
+        rsync
+        --archive
+        --hard-links
+        --update
+        --human-readable
+        --itemize-changes
+        --progress
+        --verbose
+        --perms
+        --chmod='Dugo-s,Dug=rwx,Do=rx,Fug=rw,Fo=r'
+        --omit-dir-times
+        --no-group
+        --no-owner
+        --delay-updates
+      )
+
+      options << '--dry-run' if dryrun
+      options << path
+
+      if destination
+        options << "#{destination}:#{dest_path.parent}"
+      else
+        options << "#{dest_path.parent}"
+      end
+
+      options.join("\s")
+    end
+
+    # @param path [String] The path to mangle permissions for
+    # @param sudo [Boolean] Whether or not the chmod command
+    #   should be wrapped by sudo
+    #
+    # @return [String] a chmod command (optionally wrapped in sudo)
+    #   that can be executed on a remote host
+    #   to mangle/reset permissions for a given directory
+    def repo_permissions_command(path, sudo = true)
+      cmd = "chmod -R g-s,g=rwX #{path}"
+      cmd = "sudo -E #{cmd}" if sudo
+      cmd
+    end
+
     def create_repos(directory = "repos")
       Dir.chdir(directory) do
         createrepo = Pkg::Util::Tool.check_tool('createrepo')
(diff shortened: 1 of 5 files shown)


mckern:
This is pretty sharply copy-pasta'ed out of the Deb code paths, with a new config value for `:yum_staging_server`. I'm trying to work out how we got `pl-createrepo` on weth and the burjis, so that I can recreate it on weth-dev and do a clean test of this; that's probably the next step for this PR.

shrug:
Does it work?

shrug:
:+1:

mckern:
FYI, please don't merge until @melissa and @underscorgan get a chance to review it

ScottGarman:
I won't merge but nothing seems totally insane here. :+1:

melissa:
Is there a depreciation warning for `yum_host`?

tasks/ship.rake
```diff
@@ -18,21 +18,21 @@ namespace :pl do
...
-    desc "Update remote yum repository on '#{Pkg::Config.yum_host}'"
+    desc "Update remote yum repository on '#{Pkg::Config.yum_staging_server}'"
     task :update_yum_repo do
       yum_whitelist = {
         :yum_repo_name => "__REPO_NAME__",
         :yum_repo_path => "__REPO_PATH__",
-        :yum_host      => "__REPO_HOST__",
-        :gpg_key       => "__GPG_KEY__",
+        :yum_staging_server => "__REPO_HOST__",
+        :gpg_key => "__GPG_KEY__",
```
melissa: You don't want the `=>` to be lined up?

tasks/ship.rake
```diff
@@ -138,6 +138,22 @@ namespace :pl do
...
     end
+
+    desc "Copy rpm repos from #{Pkg::Config.yum_staging_server} to #{Pkg::Config.yum_host}"
+    task :deploy_yum_repos do
+      puts "Really run remote rsync to deploy yum repos from #{Pkg::Config.yum_staging_server} to #{Pkg::Config.yum_host}? [y,n]"
+      if ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          Pkg::Rpm::Repo.deploy_repos(
+            Pkg::Config.yum_repo_path,
+            Pkg::Config.yum_repo_path,
```
melissa: Lines 148 and 149 look like duplicates

mckern:
@melissa I think I need to run through the config values I've added and refactor how they're deprecated (and if they should be). As for the fat arrows aligning, that doesn't matter one way or the other.

tasks/ship.rake
mckern:
```ruby
def deploy_repos(yum_path, destination_staging_path, origin_server, destination_server, dryrun = false)
```

mckern: (I concede it is crap UX/design)

lib/packaging/rpm/repo.rb
```diff
@@ -29,6 +29,53 @@ def repo_creation_command(createrepo)
...
+        --perms
+        --chmod='Dugo-s,Dug=rwx,Do=rx,Fug=rw,Fo=r'
+        --omit-dir-times
+        --no-group
+        --no-owner
+      )
+
+      options << '--dry-run' if dryrun
+      options << path
+      if !destination.nil?
```
melissa: Just a nitpick, but this might be more legible if we switch the logic of this conditional around

mckern: We both know no amount of legibility will make this function less shameful.

mckern:
@melissa the `if !` has been turned into an `unless`

lib/packaging/rpm/repo.rb
```diff
@@ -169,5 +218,35 @@ def create_repos_from_artifacts
...
+      cp_command = repo_deployment_command(destination_staging_path, yum_path, nil, dryrun)
+      # Defensive permissions setting are defensive
+      chmod_command = "sudo chmod -R g=rwX #{destination_staging_path}; sudo chmod -R g-s,g=rwX #{yum_path}"
+
+      if dryrun
+        puts "[DRYRUN] not executing #{chmod_command} on #{destination_server}"
+      else
+        Pkg::Util::Net.remote_ssh_cmd(destination_server, chmod_command)
+      end
+      Pkg::Util::Net.remote_ssh_cmd(origin_server, rsync_command)
```
melissa: Where is `rsync_command` set? I'm not seeing it anywhere

mckern:
Travis (and rubocop) just reminded me why I used the `if !` form instead of the `unless`:

```
lib/packaging/rpm/repo.rb:72:7: C: Style/UnlessElse: Never use unless with else. Rewrite these with the positive case first.
      unless destination.nil?
```

lib/packaging/rpm/repo.rb
```diff
@@ -169,5 +218,35 @@ def create_repos_from_artifacts
...
+    #
+    # @param yum_path [String] path for Deb repos on local and remote filesystem
+    # @param destination_staging_path [String] staging path for Deb repos on
+    #        remote filesystem
+    # @param origin_server [String] remote host to start the  rsync from
+    # @param destination_server [String] remote host to send rsynced content to
+    # @param dryrun [Boolean] whether or not to use '--dry-run'
+    def deploy_repos(yum_path, destination_staging_path, origin_server, destination_server, dryrun = false)
+      rsync_pkgs = repo_deployment_command(yum_path, destination_staging_path, destination_server, dryrun)
+      rsync_metadata = repo_deployment_command(yum_path, destination_staging_path, destination_server, dryrun)
```
melissa: I'm also not seeing either of these used anywhere in the method. Also, they're identical. What are the differences and what are they used for?

lib/packaging/rpm/repo.rb
haus: `if not nil?` can usually just be `if $thing`

mckern: Yeah, you'd think so, right?

melissa: Wait, really? `if destination` doesn't work? Blarg. It sounds like this is getting too complicated. I'm fine keeping the original syntax

melissa:
Pending tests, this looks good to me. What have you done to test it @mckern ?

mckern:
@melissa Travis. Gonna try it tonight.

underscorgan:
:+1: pending tests!

mckern:
Running `rake pl:ship_rpms` and `rake pl:remote:deploy_yum_repos` appears to do the needful now but I had to patch more than just the new methods I've added to make it work predictably. I still hate this, but tests pass and it works. Please review. My apologies for all the yak hair.

yak-shaving

lib/packaging/util/net.rb
```diff
@@ -31,14 +31,28 @@ def check_host(host, args = { :required => true })
...
+      cmd = "#{ssh} -t #{target} '#{command.gsub("'", "'\\\\''")}'"
+
+      # This is NOT a good way to support this functionality.
+      # It needs to be refactored into a set of methods that
+      # other methods can use to safely and deterministically
+      # support dry-run operations.
+      # But I need to ship packages RIGHT NOW.
+      # - Ryan McKern, 13/01/2016
+      if ENV['DRYRUN']
+        puts "[DRY-RUN] Executing '#{command}' on #{target}"
```
haus: looks like Executing command on target gets output twice. maybe move the puts on line 48 into an else here?

mckern: Line 48 never gets exec'ed if dry-run because of the short-circuit `return` here. Am I misunderstanding?

haus: nope. i can't read.

mckern: Well, I knew that.

tasks/ship.rake
```diff
@@ -1,14 +1,28 @@
 namespace :pl do
-  desc "Ship mocked rpms to #{Pkg::Config.yum_host}"
+  desc "Ship mocked rpms to #{Pkg::Config.yum_staging_server}"
   task :ship_rpms do
     ["aix", "cisco-wrlinux", "el", "eos", "fedora", "nxos", "sles"].each do |dist|
+      next unless File.directory?("pkg/#{dist}")
```
melissa: Do we need this if we also have lines 7 and 8?

mckern: ¯\_(ツ)_/¯

Let me think about it for a few minutes; I generally just kept the same overall logic as before when I rearranged the task.

melissa:
Yup, this looks good to me. :+1: 

Thanks @mckern !

tasks/ship.rake
melissa: This comment isn't a blocker, but it'd be nice to keep the code as clean as possible.

underscorgan: I think it should work without that check, was mostly concerned what Dir would do with a path that didn't exist, but irb seemed to work so *shrug*

mckern: Everything is gross.

tasks/ship.rake
```diff
@@ -158,13 +187,13 @@ namespace :pl do
     end
   end
 
-  desc "ship apple dmg to #{Pkg::Config.yum_host}"
+  desc "ship apple dmg to #{Pkg::Config.yum_staging_server}"
```
underscorgan: I know this is just updating existing code, but I feel like this should maybe be its own parameter/variable/etc?

Surely out of the scope of this work, but might want to ticket it at least :\

mckern: Agreed

mckern:
Updated to remove that extra `if this is a dir` check

melissa:
My earlier :+1: still stands

PR #487, opened 2016-01-06
Update platform additions to JJB and BHG processes

justinstoller:
This updates the platform additions ticketing task to remove steps now unnecessary since moving pipelines to JJB and Beaker-HostGenerator. Since this is contains several very small updates to the same file it may be easiest to review the commits individually.

I'm unaware of a good way to test these changes, perhaps you can help @melissa ?

@kylog , @joshcooper , and @er0ck these are the changes we discussed in HC.
@jpartlow this includes the similar update for the Integration team that I mentioned.

/cc @waynr and @jameswagner2015 this updates the ticketing script for platform additions to the more self-service workflow we've discussed.

### DIFF
--- a/tasks/agent-tickets.rake
+++ b/tasks/agent-tickets.rake
@@ -241,7 +241,7 @@ DOC
       :summary      => "Update platform puppet-agent jenkins pipelines to include #{vars[:platform_tag]}",
       :description  => description[:platform_jenkins],
       :story_points => '1',
-      :blocked_by   => ['build_data', 'puppet_agent_configuration', 'puppet_pre_suites', 'facter_pre_suites', 'hiera_pre_suites', 'pooler_image', 'decide_tier'],
+      :blocked_by   => ['build_data', 'puppet_agent_configuration', 'pooler_image', 'decide_tier'],
     },
     {
       :short_name   => 'internal_agent_ship',
@@ -264,7 +264,7 @@ DOC
       :project      => 'PE',
       :summary      => "Add #{vars[:platform_tag]} to pe_repo module",
       :description  => description[:pe_repo],
-      :blocked_by   => ['platform_jenkins', 'puppet_pre_suites', 'facter_pre_suites', 'hiera_pre_suites', 'internal_agent_ship', 's3_agent_ship'],
+      :blocked_by   => ['platform_jenkins', 'internal_agent_ship', 's3_agent_ship'],
     },
     {
       :short_name   => 'pe_jenkins',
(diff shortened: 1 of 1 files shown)


tasks/agent-tickets.rake
```diff
@@ -236,51 +220,6 @@ DOC
...
-    },
-    {
-      :short_name   => 'hiera_pre_suites',
-      :project      => 'HI',
-      :summary      => "Get pre-suites working for #{vars[:platform_tag]} for Hiera",
-      :description  => description[:pre_suites],
-      :story_points => '2',
-      :blocked_by   => ['hiera_test_targets'],
-    },
-    {
```
joshcooper: +1000

melissa:
To test, set `JIRA_INSTANCE=https://jira1-test.ops.puppetlabs.net` and run the rake task

kylog:
:+1:

tasks/agent-tickets.rake
```diff
@@ -330,12 +268,11 @@ DOC
...
     {
       :short_name   => 'pe_jenkins',
-      :project      => 'QENG',
+      :project      => 'PE',
       :summary      => "Update PE Integration jenkins pipelines to include #{vars[:platform_tag]}",
-      :description  => description[:pe_integration],
+      :description  => description[:pe_jenkins],
       :blocked_by   => ['pe_repo', 'platform_jenkins', 'internal_agent_ship', 's3_agent_ship'],
       :story_points => '1',
-      :components   => ['CI', 'Scrum Team - Integration'],
```
jpartlow: Can we set :scrum_team => 'Integration'?  I don't know if it can handle that field, but that would ensure that the ticket shows up in our backlog, otherwise it's just a PE ticket floating about.

melissa: @jpartlow it's probably do-able, but we'd need to validate the scrum team field with the jira api. It's possible it's a custom field. https://github.com/puppetlabs/packaging/blob/master/lib/packaging/util/jira.rb#L27 would need to be updated to handle it.

jpartlow: Thanks @melissa; I don't think it's critical, so I don't think we need to hold up this PR to figure it out.

joshcooper:
:+1:

tasks/agent-tickets.rake
justinstoller: FWIW, I looked it up and I scrum team is `customfield_11500` and the Integration team's id is `11214`.

I don't see a good way of searching for valid values for custom fields (and I've exhausted the amount of time I have for it). Hard coding those values would work-ish, but it would be hacky and I don't think you'd be able to test it (Are the ids the same between the staging and prod envs?).

I'm going to leave that info above here incase someone else wants to jump on that.

justinstoller:
For review by my esteemed colleagues, tickets for our newest platform starting at:
https://jira1-test.ops.puppetlabs.net/browse/PA-149

PR #489, opened 2016-01-19
(maint) add a bundler group for jira tasks

ScottGarman:
The jira ticket rake tasks depend on the jira-ruby gem, so this
adds a bundler gem group for jira functionality.


underscorgan:
:+1:

PR #491, opened 2016-01-20
Polish up the Gemfile

mckern:
gemcut011

This PR adds a constraint around ActiveRecord, depending on the Ruby revision (which should be safe as this is an integer which only goes up, and is sane; unlike `RUBY_VERSION`). This will let us use newer versions of `jira-ruby` without worrying about accidentally pulling in ActiveRecord 5 now that `jira-ruby` has unconstrained their ActiveRecord requirements.

While I was in there, I also updated the Ruby hash syntax (since a `Gemfile` is kinda just a big dumb Ruby hash) and mangled quotes so that they're consistent across the file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,11 +1,11 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
 group :development, :test do
-  gem 'rake', "~> 0.9.6"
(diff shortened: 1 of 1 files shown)


mckern:
Oh, I didn't realize I pushed this into `upstream:gemfile`. Blargh.

PR #493, opened 2016-01-23
(RE-6286) Add pl:fetch to tasks to get build data

shermdog:
Some tasks requires yum_staging_server or apt_signing_server.  We need to fetch that data to be able to use it.


mckern:
DEPEND ON ALL THE THINGS

:+1:

melissa:
:+1: I'm sure there's more we could add this too... It might not be a bad idea to hold off on any more changes until the release is out. Even though this change looks like it will do only good, I'm still scared

PR #496, opened 2016-01-27
(maint) DMGs don't get shipped to production

underscorgan:
Currently shipping is weird and a bit quirky. Let's try to make it a little better!


mckern:
no-touching

underscorgan:
it'll probably be less awful if I get lots of little PRs rather than jamming everything into the same PR.

McdonaldSeanp:
:+1:

PR #497, opened 2016-01-27
(MAINT) Change how rsync copies rpm & deb repos

mckern:
These new settings should hopefully lead to less conflict
and gnashing of teeth and rending of garments and lamentations
from our loved ones.

### DIFF
--- a/lib/packaging/deb/repo.rb
+++ b/lib/packaging/deb/repo.rb
@@ -194,15 +194,20 @@ SignWith: #{Pkg::Config.gpg_key}"
       # fail due to permission errors.
       options = %w(
         rsync
+        --delay-updates
+        --links
         --hard-links
         --copy-links
         --omit-dir-times
         --progress
-        --archive
+        --recursive
         --update
+        --human-readable
+        --itemize-changes
         --verbose
         --perms
         --no-group
+        --no-owner
         --chmod='Dugo-s,Dug=rwx,Do=rx,Fug=rw,Fo=r'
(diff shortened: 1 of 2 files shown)


ScottGarman:
Short of trying to ship something using this change, how else can it be tested? I'm cautiously :+1:

mckern:
I have some new test/dev servers that we can test this against. I'll get some data tonight.

lib/packaging/deb/repo.rb
```diff
@@ -194,16 +194,21 @@ def repo_deployment_command(origin_path, destination_path, destination, dryrun =
       # fail due to permission errors.
       options = %w(
         rsync
+        --ignore-existing
+        --recursive
+        --links
         --hard-links
-        --copy-links
```
underscorgan: I think we still need --copy-links

haus:
I love that the flags needed are so different.

mckern:
tumblr_n3cochlvij1sl4pv2o1_500

lib/packaging/deb/repo.rb
mckern: Oof, we do; we're still resolving those debs aren't we? OK. Will fix.

stahnma:
@mckern  so this is 24 days old. Is it ready or abandoned?

mckern:
@stahnma neither but I don't think we ever got agreement on what the filesystem on the receiving end should absolutely look like, so this didn't go anywhere. Closing for now, will reopen if we nail that down.

PR #499, opened 2016-02-02
(RE-6453) Exclude apt/yum shipping when on same host

shermdog:
Some projects such as pl-build-tools do not use a separate signing/staging server for repos.  As such we only need to update apt/yum repo locally.  This commit exludes the shipping jobs when the staging/ship host is the same.

### DIFF
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -260,6 +260,9 @@ namespace :pl do
         remote:deploy_yum_repo
         remote:update_ips_repo
       )
+      # Some projects such as pl-build-tools do not stage to a separate server - so we do to deploy
+      uber_tasks.delete("remote:deploy_apt_repo") if Pkg::Config.apt_host == Pkg::Config.apt_signing_server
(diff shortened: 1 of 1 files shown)


tasks/jenkins.rake
```diff
@@ -260,6 +260,9 @@ namespace :pl do
         remote:deploy_yum_repo
         remote:update_ips_repo
       )
+      # Some projects such as pl-build-tools do not stage to a separate server - so we do to deploy
+      uber_tasks.delete("remote:deploy_apt_repo") if (Pkg::Config.apt_host == Pkg::Config.apt_signing_server)
```
shermdog: Are they good for future multiple conditionals?

ScottGarman:
:+1:

tasks/jenkins.rake
mckern: Pre-optimizing for problems you don't have?

shermdog: True, just didn't want to leave any landmines :)  I'll fix that up right quick.

mckern:
:+1: will merge if no one else does

PR #501, opened 2016-02-09
(RE-6493) explicitly require 'set'

mckern:
Without requiring 'set', the #to_set method will not be available on Arrays.


melissa:
:+1: let's merge this in!

PR #502, opened 2016-02-10
(maint) UNRAKE MY HEARTTTTTTTT

underscorgan:
SAY YOU'LL DEPRECATE MY TASKS AGAIN

### DIFF
--- a/lib/packaging/osx.rb
+++ b/lib/packaging/osx.rb
@@ -6,7 +6,7 @@ module Pkg::OSX
       ssh_host_string = "#{use_identity} #{ENV['USER']}@#{Pkg::Config.osx_signing_server}"
       rsync_host_string = "-e 'ssh #{use_identity}' #{ENV['USER']}@#{Pkg::Config.osx_signing_server}"
 
-      work_dir  = "/tmp/#{rand_string}"
+      work_dir  = "/tmp/#{Pkg::Util.rand_string}"
       mount     = File.join(work_dir, "mount")
       signed    = File.join(work_dir, "signed")
       Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, "mkdir -p #{mount} #{signed}")
--- a/lib/packaging/util.rb
+++ b/lib/packaging/util.rb
@@ -84,4 +84,21 @@ module Pkg::Util
   def self.rand_string
     rand.to_s.split('.')[1]
   end
+
+  def self.ask_yes_or_no
+    return Pkg::Util.boolean_value(ENV['ANSWER_OVERRIDE']) unless ENV['ANSWER_OVERRIDE'].nil?
+    answer = Pkg::Util.get_input
+    return true if answer =~ /^y$|^yes$/
+    return false if answer =~ /^n$|^no$/
+    puts "Nope, try something like yes or no or y or n, etc:"
+    Pkg::Util.ask_yes_or_no
+  end
+
+  def self.confirm_ship(files)
+    puts "The following files have been built and are ready to ship:"
+    files.each { |file| puts "\t#{file}\n" unless File.directory?(file) }
+    puts "Ship these files?? [y,n]"
+    Pkg::Util.ask_yes_or_no
+  end
+
 end
--- a/lib/packaging/util.rb
+++ b/lib/packaging/util.rb
@@ -95,9 +95,9 @@ module Pkg::Util
   end
 
   def self.confirm_ship(files)
-    puts "The following files have been built and are ready to ship:"
+    STDOUT.puts "The following files have been built and are ready to ship:"
     files.each { |file| puts "\t#{file}\n" unless File.directory?(file) }
-    puts "Ship these files?? [y,n]"
+    STDOUT.puts "Ship these files?? [y,n]"
     Pkg::Util.ask_yes_or_no
   end
 
@@ -106,6 +106,6 @@ module Pkg::Util
     if new_cmd
       msg << " Please use #{new_cmd} instead."
     end
-    puts("\n#{msg}\n")
+    STDOUT.puts("\n#{msg}\n")
   end
 end
--- a/lib/packaging/util.rb
+++ b/lib/packaging/util.rb
@@ -80,4 +80,8 @@ module Pkg::Util
   ensure
     system 'stty echo'
   end
+
+  def self.rand_string
+    rand.to_s.split('.')[1]
+  end
 end
--- a/lib/packaging/util.rb
+++ b/lib/packaging/util.rb
@@ -101,4 +101,11 @@ module Pkg::Util
     Pkg::Util.ask_yes_or_no
   end
 
+  def self.deprecate(old_cmd, new_cmd = nil)
+    msg = "!! #{old_cmd} is deprecated."
+    if new_cmd
+      msg << " Please use #{new_cmd} instead."
+    end
+    puts("\n#{msg}\n")
+  end
 end
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -152,5 +152,22 @@ module Pkg::Util::Net
       remote_cmd = "for file in #{files.join(" ")}; do lsattr $file | grep -q '\\-i\\-'; if [ $? -eq 1 ]; then sudo chmod #{permissions} $file; else echo \"$file is immutable\"; fi; done"
       Pkg::Util::Net.remote_ssh_cmd(host, remote_cmd)
     end
+
+    def escape_html(uri)
+      require 'cgi'
+      CGI.escapeHTML(uri)
+    end
+
+    # Add a parameter to a given uri. If we were sane we'd use
+    # encode_www_form(params) of URI, but because we're not, because that will http
+    # encode it, which isn't what we want since we're require the encoding provided
+    # by escapeHTML of CGI, since this is being transfered in the xml of a jenkins
+    # job via curl and DEAR JEEBUS WHAT HAVE WE DONE.
+    def add_param_to_uri(uri, param)
+      require 'uri'
+      uri = URI.parse(uri)
+      uri.query = [uri.query, param].compact.join('&')
+      uri.to_s
+    end
   end
 end
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -153,6 +153,11 @@ module Pkg::Util::Net
       Pkg::Util::Net.remote_ssh_cmd(host, remote_cmd)
     end
 
+    # Remotely set the immutable bit on a list of files
+    def remote_set_immutable(host, files)
+      Pkg::Util::Net.remote_ssh_cmd(host, "sudo chattr +i #{files.join(" ")}")
+    end
+
     def escape_html(uri)
       require 'cgi'
       CGI.escapeHTML(uri)
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -174,5 +174,30 @@ module Pkg::Util::Net
       uri.query = [uri.query, param].compact.join('&')
       uri.to_s
     end
+
+    # We take a tar argument for cases where `tar` isn't best, e.g. Solaris.  We
+    # also take an optional argument of the tarball containing the git bundle to
+    # use.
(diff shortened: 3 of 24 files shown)


underscorgan:
There is still a lot of WIP stuff here, but if you feel so inclined to merge I have other rake puns I can use.

Otherwise, I'll keep pushing here as I deprecate things.

underscorgan:
This deprecates all the tasks from 00_utils and moves them to ruby libraries.

shrug:
i-love-this-132299859679

shrug:
320x242-fire-safety-trailer-trash-can-24-300x226

shermdog:
Skimming through everything seems sane to me.

mckern:
### IMMA LET YOU FINISH BUT MORGAN HAD THE BEST PR TITLE OF ALL TIME

transparent_gold_cup_trophy_png_picture

(Still reviewing the actual work)

lib/packaging/osx.rb
```diff
@@ -6,7 +6,7 @@ def sign(target_dir = 'pkg')
       ssh_host_string = "#{use_identity} #{ENV['USER']}@#{Pkg::Config.osx_signing_server}"
       rsync_host_string = "-e 'ssh #{use_identity}' #{ENV['USER']}@#{Pkg::Config.osx_signing_server}"
 
-      work_dir  = "/tmp/#{rand_string}"
+      work_dir  = "/tmp/#{Pkg::Util.rand_string}"
```
mckern: ... we were doing this instead of using `mktemp`? We deserve this pain.

lib/packaging/util.rb
```diff
@@ -80,4 +80,32 @@ def self.get_input(echo = true)
   ensure
     system 'stty echo'
```
mckern: I know you didn't do this but what-the-what were we doing here?

lib/packaging/util.rb
```diff
@@ -80,4 +80,32 @@ def self.get_input(echo = true)
...
   end
+
+  def self.rand_string
+    rand.to_s.split('.')[1]
```
mckern: dawson

lib/packaging/util.rb
```diff
@@ -80,4 +80,32 @@ def self.get_input(echo = true)
...
+  end
+
+  def self.ask_yes_or_no
+    return Pkg::Util.boolean_value(ENV['ANSWER_OVERRIDE']) unless ENV['ANSWER_OVERRIDE'].nil?
```
mckern: I'm worried that I'm the one who wrote this version of `#ask_yes_or_no` originally, but I am endlessly amused by what happens if some clown named Me sets `ANSWER_OVERRIDE=false` here.

Should we fix this?

lib/packaging/util.rb
shrug: "# @param noecho [Boolean, nil] if we are retrieving command line input with or without privacy. This is mainly
 #   for sensitive information like passwords."

lib/packaging/util.rb
underscorgan: Oh good. Yes. yes we should.

lib/packaging/util.rb
mckern: unicorn-rainbow

Thanks @shrug

lib/packaging/util.rb
mckern: So, it looks like `#boolean_value` will PROBABLY do the right thing:

```ruby

  def self.boolean_value(var)
    return TRUE if var == TRUE || ( var.is_a?(String) && ( var.downcase == 'true' || var.downcase =~ /^y$|^yes$/))
    FALSE
  end
```

But yeah, we can probably turn that into a more straight-forward code path. It's probably safe to leave well enough alone in this PR if we sanitize this later on.

lib/packaging/util.rb
```diff
@@ -80,4 +80,32 @@ def self.get_input(echo = true)
...
+  end
+
+  def self.confirm_ship(files)
+    STDOUT.puts "The following files have been built and are ready to ship:"
+    files.each { |file| puts "\t#{file}\n" unless File.directory?(file) }
+    STDOUT.puts "Ship these files?? [y,n]"
+    Pkg::Util.ask_yes_or_no
+  end
+
+  def self.deprecate(old_cmd, new_cmd = nil)
```
mckern: ARE WE DEPRECATING THE ORIGINAL DEPRECATE COMMAND?!

dancing panda

lib/packaging/util.rb
```diff
@@ -80,4 +80,32 @@ def self.get_input(echo = true)
...
+    msg = "!! #{old_cmd} is deprecated."
+    if new_cmd
+      msg << " Please use #{new_cmd} instead."
+    end
+    STDOUT.puts("\n#{msg}\n")
```
mckern: While we're in here, can we change this to use `$stdout` instead of `STDOUT`? Might as well tidy those up where we find them.

lib/packaging/util.rb
```diff
@@ -80,4 +80,32 @@ def self.get_input(echo = true)
...
+    return Pkg::Util.boolean_value(ENV['ANSWER_OVERRIDE']) unless ENV['ANSWER_OVERRIDE'].nil?
+    answer = Pkg::Util.get_input
+    return true if answer =~ /^y$|^yes$/
+    return false if answer =~ /^n$|^no$/
+    puts "Nope, try something like yes or no or y or n, etc:"
+    Pkg::Util.ask_yes_or_no
+  end
+
+  def self.confirm_ship(files)
+    STDOUT.puts "The following files have been built and are ready to ship:"
```
mckern: See later comment re: `$stdout` vs `STDOUT` and tidying them up where we find them.

lib/packaging/util.rb
underscorgan: always be deprecating!

lib/packaging/util/net.rb
```diff
@@ -152,5 +152,52 @@ def remote_set_permissions(host, permissions, files)
       remote_cmd = "for file in #{files.join(" ")}; do lsattr $file | grep -q '\\-i\\-'; if [ $? -eq 1 ]; then sudo chmod #{permissions} $file; else echo \"$file is immutable\"; fi; done"
       Pkg::Util::Net.remote_ssh_cmd(host, remote_cmd)
     end
+
+    # Remotely set the immutable bit on a list of files
+    def remote_set_immutable(host, files)
```
mckern: No one has ever regretted this ever anywhere for any reason.

lib/packaging/util/net.rb
```diff
@@ -152,5 +152,52 @@ def remote_set_permissions(host, permissions, files)
...
+    def escape_html(uri)
+      require 'cgi'
+      CGI.escapeHTML(uri)
+    end
+
+    # Add a parameter to a given uri. If we were sane we'd use
+    # encode_www_form(params) of URI, but because we're not, because that will http
+    # encode it, which isn't what we want since we're require the encoding provided
+    # by escapeHTML of CGI, since this is being transfered in the xml of a jenkins
+    # job via curl and DEAR JEEBUS WHAT HAVE WE DONE.
```
mckern: DEER JEBUS indeed.

lib/packaging/util/net.rb
```diff
@@ -152,5 +152,52 @@ def remote_set_permissions(host, permissions, files)
...
+      uri.to_s
+    end
+
+    # We take a tar argument for cases where `tar` isn't best, e.g. Solaris.  We
+    # also take an optional argument of the tarball containing the git bundle to
+    # use.
+    def remote_bootstrap(host, treeish, tar_cmd = nil, tarball = nil)
+      unless tar = tar_cmd
+        tar = 'tar'
+      end
```
mckern: ... what were we even doing here?

lib/packaging/util/net.rb
```diff
@@ -152,5 +152,52 @@ def remote_set_permissions(host, permissions, files)
...
+      tarball ||= Pkg::Util::Git.git_bundle(treeish)
+      tarball_name = File.basename(tarball).gsub('.tar.gz', '')
+      Pkg::Util::Net.rsync_to(tarball, host, '/tmp')
+      appendix = Pkg::Util.rand_string
+      Pkg::Util::Net.remote_ssh_cmd(host, "#{tar} -zxvf /tmp/#{tarball_name}.tar.gz -C /tmp/ ; git clone --recursive /tmp/#{tarball_name} /tmp/#{Pkg::Config.project}-#{appendix} ; cd /tmp/#{Pkg::Config.project}-#{appendix} ; rake package:bootstrap")
```
mckern: YOU MADE IT USE `remote_ssh_cmd` AND WE ARE FRIENDS FOREVER (if you want)

lib/packaging/util/net.rb
```diff
@@ -152,5 +152,52 @@ def remote_set_permissions(host, permissions, files)
...
+      "/tmp/#{Pkg::Config.project}-#{appendix}"
+    end
+
+    # Given a BuildInstance object and a host, send its params to the host. Return
+    # the remote path to the params.
+    def remote_buildparams(host, build)
+      params_file = build.config_to_yaml
+      params_file_name = File.basename(params_file)
+      params_dir = Pkg::Util.rand_string
```
mckern: ROFL at our continued use of `#rand_string`.

parks-and-recreation-jean-ralphio-the-worst-worst-woooorst-1372637673p

tasks/00_utils.rake
```diff
@@ -252,26 +119,98 @@ end
...
+
+def add_param_to_uri(uri, param)
+  Pkg::Util.deprecate('add_param_to_uri', 'Pkg::Util::Net.add_param_to_uri')
+  Pkg::Util::Net.add_param_to_uri(uri, param)
+end
+
+def cp_pr(src, dest, options = {})
+  Pkg::Util.deprecate('cp_pr', 'FileUtils.cp_r')
+  mandatory = { :preserve => true }
+  FileUtils.cp_r(src, dest, options.merge(mandatory))
```
mckern: I don't think `mandatory` will merge down into options in any meaningful way so this may need a little juggling.

http://ruby-doc.org/stdlib-1.9.3/libdoc/fileutils/rdoc/FileUtils.html#method-c-cp_r

mckern:
I am tentatively :+1: on this; I want to double check the list of methods deprecated against methods defined and validate they add up (minus `check_var`, which was partially migrated last year). But I also want to go home. Will follow up later tonight.

Great work. Appreciated and important.

lib/packaging/util/net.rb
underscorgan: weeee should probably move all those to tmpdir. someday.

mckern: s4e15_magic_man_disguised_as_jake

spec/lib/packaging/util/git_spec.rb
```diff
@@ -60,7 +60,7 @@
     it "should create a git bundle with random appendix and random output directory" do
       Pkg::Util::Version.should_receive(:is_git_repo?).and_return(true)
       Pkg::Util::File.should_receive(:mktemp).and_return(temp)
-      Pkg::Util::Git.should_receive(:rand_string).and_return(string)
+      Pkg::Util.should_receive(:rand_string).and_return(string)
```
mckern: I think this is broken?

underscorgan: Ok, so, I think what's happening here is rand_string was available in the Pkg::Util::Git namespace when it was imported as a rake task, but now that it's moved to Pkg::Util you need to stub it there so that when `Pkg::Util::Git.git_bundle(treeish)` is called later in the test it returns 'bar' for the random string.  It looks like namespacing was SUPER WEIRD when everything was a rake task?

mckern: That is probably true; Rake tasks only really have one namespace (Rake) and everything is mashed together inside it.

mckern: OK, so this does make sense. Got it. Sorry for the false alarm.

mckern:
:+1:

spec/lib/packaging/util/git_spec.rb
underscorgan: No worries! Helped me understand it better, too.

PR #503, opened 2016-02-10
(RE-6498) Fix permissions when shipping

underscorgan:
We want to set group and ownership correctly on staging, then preserve these permissions when shipping.

This also addresses a few failures we were seeing.


underscorgan:
This was tested on the dev shipping infrastructure.

McdonaldSeanp:
:+1:

PR #504, opened 2016-02-17
(RE-5817) Add foss-only mode

underscorgan:
This updates the retrieve jenkins task to only fetch specified platforms
if FOSS_ONLY is set to true. It will default to using foss_platforms if
specified, but that can be overwritten with the FOSS_PLATFORMS
environment variable which should be a space-separated list of
platform-tags.


underscorgan:
Still to-do: 

- [ ] Add some deduplication (will currently fetch all packages for the versions once per architecture, we could be smarter about this)
- [ ] Add more robustness / print warnings when something goes wrong and we want to fail back
- [ ] Add some spec testing maybe????

lib/packaging/config/params.rb
```diff
@@ -172,6 +173,8 @@ module Pkg::Params
               { :var => :default_cow,             :envvar => :COW },
               { :var => :default_mock,            :envvar => :MOCK },
               { :var => :final_mocks,             :envvar => :MOCK },
+              { :var => :foss_only,               :envvar => :FOSS_ONLY },
+              { :var => :foss_platforms,          :envvar => :FOSS_PLATFORMS,  :type => :array },
```
mckern: Yaaaassssssssss `:type => :array`! I knew that was going to be handy.

spec/lib/packaging/util/platform_spec.rb
```diff
@@ -32,6 +32,19 @@
     end
   end
 
+  describe '#artifacts_path' do
+    it 'should be correct' do
+      expect(Pkg::Util::Platform.artifacts_path('el-7-x86_64')).to eq('artifacts/el/7')
```
mckern: Would we benefit from testing other platforms?

underscorgan: Probably. I was slightly lazy / copypastad the tests, but I can udpate it to hit all the cases in the case statement.

underscorgan: More tests added!

tasks/retrieve.rake
```diff
@@ -17,17 +17,45 @@ namespace :pl do
...
-        # For the next person who needs to look these flags up:
-        # -r = recursive
-        # -l 0 = infinitely recurse, no limit
-        # --cut-dirs 3 = will cut off #{Pkg::Config.project}, #{Pkg::Config.ref}, and the first directory in #{remote_target} from the url when saving to disk
-        # -np = Only descend when recursing, never ascend
-        # -nH = Discard http://#{Pkg::Config.builds_server} when saving to disk
-        # --reject = Reject all hits that match the supplied regex
-        # -P = where to save to disk (defaults to ./)
-        sh "#{wget} -r -np -nH -l 0 --cut-dirs 3 -P #{local_target} --reject 'index*' #{package_url}/"
+        if Pkg::Config.foss_only and not Pkg::Config.foss_platforms
```
mckern: `&& !` is [probably better than](http://phrogz.net/ProgrammingRuby/language.html#table_18.4) `and not`. Wheeeeee!

lib/packaging/util/platform.rb
```diff
@@ -41,6 +41,28 @@ def get_attribute(platform_tag, attribute_name)
       end
     end
 
+    def artifacts_path(platform_tag)
```
melissa: This makes me sad... I wish we didn't have to define this everywhere

lib/packaging/util/platform.rb
```diff
@@ -41,6 +41,28 @@ def get_attribute(platform_tag, attribute_name)
...
+        File.join('artifacts', 'deb', get_attribute(platform_tag, :codename))
+      when 'svr4', 'ips'
+        # solaris/10/PC1 for example
+        File.join('artifacts', 'solaris', version)
+      when 'dmg'
+        File.join('artifacts', 'apple', version)
+      when 'msi'
+        File.join('artifacts', 'windows')
+      else
+        fail "Not sure what to do with a package format of '#{package_format}'"
```
melissa: Maybe this should be `"Not sure where to find packages that belong to the format type '#{package_format}'"` to be more explicit with what's failing?

lib/packaging/util/platform.rb
```diff
@@ -41,6 +41,28 @@ def get_attribute(platform_tag, attribute_name)
...
+      when 'rpm', 'swix'
+        # el/7/PC1/x86_64 for example
+        File.join('artifacts', platform, version)
+      when 'deb'
+        File.join('artifacts', 'deb', get_attribute(platform_tag, :codename))
+      when 'svr4', 'ips'
+        # solaris/10/PC1 for example
+        File.join('artifacts', 'solaris', version)
+      when 'dmg'
+        File.join('artifacts', 'apple', version)
```
melissa: soooo, this is a vanagon thing. It looks like dmgs that are built with the packaging repo don't have the version in the path (`artifacts/apple/puppet-3.8.6.dmg` )

lib/packaging/util/platform.rb
underscorgan: I know :\

It's slightly different than the repo_path though so this seemed cleanest?

lib/packaging/util/platform.rb
melissa: Luckily we don't have the same issue with windows

lib/packaging/util/platform.rb
melissa: It probably is :disappointed:

tasks/retrieve.rake
```diff
@@ -17,17 +17,45 @@ namespace :pl do
...
-        # -P = where to save to disk (defaults to ./)
-        sh "#{wget} -r -np -nH -l 0 --cut-dirs 3 -P #{local_target} --reject 'index*' #{package_url}/"
+        if Pkg::Config.foss_only && !Pkg::Config.foss_platforms
+          warn "FOSS_ONLY specified, but I don't know anything about FOSS_PLATFORMS. Fetch everything?"
+          unless ask_yes_or_no
+            warn "Retrieve cancelled"
+            exit
+          end
+        elsif Pkg::Config.foss_only && remote_target != 'artifacts'
+          warn "I only know how to fetch from remote_target 'artifacts' with FOSS_ONLY. Fetch everything?"
```
melissa: Let's expand the `Fetch everything?` to be `Fetch everything from artifacts?` to clarify that they're not going to get anything from whatever the remote target is

melissa: wait, nvm, I totally misinterpreted what's happening with this. ignore me

underscorgan: If they answer yes it'll fall back to default behavior so it won't fetch from artifacts, but from whatever target they specified.

melissa:
Other than the issue with the dmgs and this needing a rebase, I'm :+1:

underscorgan:
I've got a dmg yak to shave here. Will look again in the morning.

underscorgan:
This should be good to go now!

melissa:
@underscorgan does it blend?

underscorgan:
@melissa YEP. I've tested it locally (yay for easy to test things!), and I can fetch everything, fetch overridden platforms, fetch for puppet and puppet-agent (to test the osx oddities).

lib/packaging/util/net.rb
```diff
@@ -124,6 +124,17 @@ def curl_form_data(uri, form_data = [], options = {})
...
 
+    def uri_status_code(uri)
+      data = [
+        '--request GET',
+        '--silent',
+        '--location',
+        '--write-out "%{http_code}"',
+        '--output /dev/null'
+      ]
+      Pkg::Util::Net.curl_form_data(uri, data)
```
mckern: kind of love the abuse of `#curl_form_data`.

underscorgan: it was already there, so may as well reuse it :)

melissa:
:100:

PR #505, opened 2016-02-17
(RE-5372) Fix signing bug

McdonaldSeanp:
Inside the pe:sign rake task, there is
a branching statement checking for OS
family and version. the check against
version was a check of >= against a string
which likely caused problems. this commit
makes OS integer in to an integer instead.

See https://github.com/puppetlabs/packaging/blob/master/tasks/pe_sign.rake#L20-L24
for more details

### DIFF
--- a/tasks/pe_sign.rake
+++ b/tasks/pe_sign.rake
@@ -10,14 +10,22 @@ if Pkg::Config.build_pe
       # Find x86_64 noarch rpms that have been created as hard links and remove them
       rm_r Dir["pkg/pe/rpm/*-*-x86_64/*.noarch.rpm"]
       # We'll sign the remaining noarch
-      sign_dists = ['el4', 'el5', 'el6', 'el7', 'sles10', 'sles11', 'sles12']
+      sign_dists = [
+      { :OS => 'el', :version => 4 },
+      { :OS => 'el', :version => 5 },
+      { :OS => 'el', :version => 6 },
+      { :OS => 'el', :version => 7 },
+      { :OS => 'sles', :version => 10 },
+      { :OS => 'sles', :version => 11 },
+      { :OS => 'sles', :version => 12 }
+      ]
       ['i386', 'x86_64'].each do |arch|
         sign_dists.each do |dist|
-          family = dist[/[a-z]+/]
-          version = dist[/[0-9]+/]
+          family = dist[:OS]
+          version = dist[:version]
           rpm_stagedir        = "pkg/pe/rpm/#{family}-#{version}-#{arch}/*.rpm"
           srpm_stagedir       = "pkg/pe/rpm/#{family}-#{version}-srpms/*.rpm"
-          if (family == 'el' and version >= '6') || (family == 'sles' and version >= '12')
+          if (family == 'el' and version >= 6) || (family == 'sles' and version >= 12)
             modern_rpms += FileList[rpm_stagedir] + FileList[srpm_stagedir]
           else
             old_rpms += FileList[rpm_stagedir] + FileList[srpm_stagedir]


McdonaldSeanp:
This was tested for the pe-enterprise-release package and remedies the issue

tasks/pe_sign.rake
```diff
@@ -10,14 +10,18 @@ if Pkg::Config.build_pe
       # Find x86_64 noarch rpms that have been created as hard links and remove them
       rm_r Dir["pkg/pe/rpm/*-*-x86_64/*.noarch.rpm"]
       # We'll sign the remaining noarch
-      sign_dists = ['el4', 'el5', 'el6', 'el7', 'sles10', 'sles11', 'sles12']
+      sign_dists = [ \
```
shrug: You don't actually need the line continuations

McdonaldSeanp: Oh yeah, heh forgot about that

tasks/pe_sign.rake
```diff
@@ -10,14 +10,19 @@ if Pkg::Config.build_pe
...
       rm_r Dir["pkg/pe/rpm/*-*-x86_64/*.noarch.rpm"]
       # We'll sign the remaining noarch
-      sign_dists = ['el4', 'el5', 'el6', 'el7', 'sles10', 'sles11', 'sles12']
+      sign_dists = [
+      { :OS => 'el', :version => 4 }, { :OS => 'el', :version => 5 },
+      { :OS => 'el', :version => 6 }, { :OS => 'el', :version => 7 },
+      { :OS => 'sles', :version => 10 }, { :OS => 'sles', :version => 11 },
+      { :OS => 'sles', :version => 12 }
+      ]
       ['i386', 'x86_64'].each do |arch|
```
shrug: Completely out of scope for this, but I hate these assumptions

shrug:
:+1:

tasks/pe_sign.rake
McdonaldSeanp: Hopefully when we start to rip out the pe tasks from packaging (which I think we talked about doing? idk maybe?) we will take those assumptions out

tasks/pe_sign.rake
```diff
@@ -10,14 +10,19 @@ if Pkg::Config.build_pe
       # Find x86_64 noarch rpms that have been created as hard links and remove them
       rm_r Dir["pkg/pe/rpm/*-*-x86_64/*.noarch.rpm"]
       # We'll sign the remaining noarch
-      sign_dists = ['el4', 'el5', 'el6', 'el7', 'sles10', 'sles11', 'sles12']
+      sign_dists = [
+      { :OS => 'el', :version => 4 }, { :OS => 'el', :version => 5 },
+      { :OS => 'el', :version => 6 }, { :OS => 'el', :version => 7 },
+      { :OS => 'sles', :version => 10 }, { :OS => 'sles', :version => 11 },
+      { :OS => 'sles', :version => 12 }
+      ]
```
mckern: An Array of Hashes! Delicious!

McdonaldSeanp: It's my favorite thing to do in ruby

McdonaldSeanp: It makes my C brain all warm and fuzzy

melissa: Is it bad that I want you to put each on its own line? I started reading that, and I was like, "el5 is missing! Oh no! wait, no, it's there"

mckern: Agreed. Vertical space is free.

McdonaldSeanp: Coo'

melissa:
:+1: pending tests

PR #507, opened 2016-02-23
DO NOT MERGE (RE-6263) Update huaweios platform hash

ScottGarman:
The current packages use ppc as the arch; this will be changed back
to powerpc when the remaining code has been thoroughly tested.


mckern:
@ScottGarman on a scale of 1-10, 10 being "houses," and 1 being "horses," what is this as safe as?

ScottGarman:
@mckern how long is a piece of string?

PR #508, opened 2016-02-23
(MAINT) Wouldn't it be rad to view Pkg::Config in a readable format?

mckern:
I thought it would be, so in service of some of my other work I bashed two rocks together to make this. It'll add two rake tasks to make validating/viewing the current state of a project's config easier.

giphy 2

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -9,6 +9,13 @@ module Pkg
     require 'yaml'
 
     class << self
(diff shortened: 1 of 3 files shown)


underscorgan:
:100:

mckern:
Rebased against `HEAD`

mckern:
Force pushed to restart the Travis cycle.

PR #509, opened 2016-02-23
(MAINT) Fix broken calls to ::File class methods

mckern:
Namespacing is hard, and we use `File#directory?` and `File#exist?` all over Packaging. When I tried fixing them, I gave up around the 170th instance of `File#somethingorother` and instead made some passthrough methods in the existing `Pkg::Util::File` class.

This is a dirty hack, and it shouldn't be how we do business. But it'll unblock things while we rename the `Pkg::Util` classes that conflict with core or stdlib classes.

### DIFF
--- a/lib/packaging/util/file.rb
+++ b/lib/packaging/util/file.rb
@@ -4,6 +4,15 @@ require 'fileutils'
 module Pkg::Util::File
 
   class << self
+    def exist?(file)
+      ::File.exist?(file)
+    end
+    alias_method :exists?, :exist?
+
+    def directory?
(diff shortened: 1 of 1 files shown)


ScottGarman:
I'm :+1: on this as a temporary fix to get shipping to work again.

mckern:
@ScottGarman yeah, this is not how I want to solve problems at all but it'll take more than 5 minutes to rename these classes and make sure nothing breaks so I'm willing to compromise.

shermdog:
Now I feel dirty.

PR #510, opened 2016-02-23
(RE-6596) Fix missing argument to File.directory?

mckern:
This was left off of the last hotfix, and it causes Packaging
to explode in new and interesting ways.


ScottGarman:
I don't have any arguments with this. :+1:

mckern:
mea-culpa1

PR #511, opened 2016-02-24
(RE-6071) Ship 'Em All

mckern:
shipemall


lib/packaging/config/params.rb
```diff
@@ -75,9 +77,10 @@ module Pkg::Params
                   :ips_repo,
                   :ips_root_cert,
                   :ips_signing_cert,
-                  :ips_signing_server,
                   :ips_signing_key,
+                  :ips_signing_server,
```
stahnma: Do we have an IPS staging server? This has to run Solaris, IIRC.

McdonaldSeanp:
I like all the spec tests, that makes me happy

mckern:
@McdonaldSeanp it could use more spec tests. Will worry about that when the `ship_` and `deploy_` tasks are done.

lib/packaging/config/params.rb
mckern: This may not matter since it looks like the `:update_ips_repo` task ALSO ships p5p packages to `ips_host`. I don't know what the meaningful difference between the `p5p` and IPS stuff in Packaging is; I thought IPS *was* `p5p`.

mckern:
There are now `ship_` and `deploy_` tasks for every package type except gems, nuget, and IPS. Gems ship straight to rubygems.org, nuget ships to a Nexus server (I think?), and right IPS is a black box (to me). These tasks need to be added to `uber_ship`, but they're probably ready for review.

@stahnma Yes, I know commit messages are sparse. I will fill them in a little bit more before this is actually merged.

lib/packaging/util/net.rb
```diff
@@ -57,22 +57,78 @@ def remote_ssh_cmd(target, command, capture_output = false)
...
+              target_path: dest,
+              extra_flags: options[:extra_flags],
+              dryrun: options[:dryrun],
+              bin: options[:bin],
+            )
+
+      Pkg::Util::Execution.ex(cmd, true)
+    end
+
+    def rsync_from(source, target_host, dest, opts = {})
```
shermdog: Is there a way we can dedup these two functions?

mckern: probably but thats less of a priority in this PR than getting non-deb and non-rpm artifacts shipping to a staging server.

mckern: Maybe.

mckern:
Oh god, this merge did not go as expect.

mckern:
## What's left to do?

- [ ] Standardize the IPS & p5p tasks
- [ ] Improve the commit messages
- [ ] Add the new `ship_` and `deploy_` tasks to the ubership task
- [ ] Review, review, view

lib/packaging/util/net.rb
```diff
@@ -57,22 +57,101 @@ def remote_ssh_cmd(target, command, capture_output = false)
...
+        target_path: dest,
+        extra_flags: opts[:extra_flags],
+        dryrun: opts[:dryrun],
+        bin: opts[:bin],
+      )
+    end
+
+    # A wrapper method to maintain the existing interface for executing
+    # incoming rsync commands with minimal changes to existing code.
+    def rsync_from(source, target_host, dest, opts = {})
```
melissa: my brain wants `target_host` to be `origin_host` here

mckern: Good catch, updated

lib/packaging/config/params.rb
mckern: The IPS stuff has been pulled out of this, so this is probably moot now.

lib/packaging/config/params.rb
```diff
@@ -245,13 +263,28 @@ module Pkg::Params
...
-                   { :oldvar => :yum_staging_server,     :newvar => :yum_host },
-                   { :oldvar => :gem_devel_dependencies, :newvar => :gem_development_dependencies },
-                   { :oldvar => :pe_name,                :newvar => :project },
-                   { :oldvar => :project,                :newvar => :gem_name },
-                   { :oldvar => :gpg_name,               :newvar => :gpg_key }]
+  REASSIGNMENTS = [
+                    # These are fall-through values for shipping endpoints
+                    { :oldvar => :staging_server,         :newvar => :apt_staging_server },
+                    { :oldvar => :staging_server,         :newvar => :dmg_staging_server },
+
```
melissa: extra space?

mckern: Side effect of a rebase. I can clean it.

stahnma:
:+1: but maybe we should have some testing of this prior to merge?

mckern:
There's some commits going into this to fix a couple of typos, but there might be a problem with copying `.deb` packages to the test production server.

```
Executing 'rsync --hard-links --copy-links --omit-dir-times --progress --archive --update --verbose --perms --chmod='Dugo-s,Dug=rwx,Do=rx,Fug=rw,Fo=r' --exclude='dists/*-*' --exclude='pool/*-*' /opt/repository/apt downloadserver-aio1-test.delivery.puppetlabs.net:/opt/tools/freight' on weth-dev.delivery.puppetlabs.net
```

That ain't right.

ScottGarman:
I hesitatingly give this a :+1:  but I think we should be strategic in when me merge this - as in not just before we do a release in case something was missed.

lib/packaging/config/params.rb
```diff
@@ -166,56 +175,65 @@ module Pkg::Params
...
               { :var => :default_cow,             :envvar => :COW },
               { :var => :default_mock,            :envvar => :MOCK },
+              { :var => :dmg_host,                :envvar => :DMG_HOST },
+              { :var => :dmg_staging_server,      :envvar => :DMG_STAGING_SERVER },
               { :var => :final_mocks,             :envvar => :MOCK },
               { :var => :foss_only,               :envvar => :FOSS_ONLY },
               { :var => :foss_platforms,          :envvar => :FOSS_PLATFORMS,  :type => :array },
-              { :var => :gpg_name,                :envvar => :GPG_NAME },
+              { :var => :gem_host,                :envvar => :GEM_HOST },
+              { :var => :gpg_key,                 :envvar => :GPG_KEY },
```
underscorgan: gpg_key/GPG_KEY is defined twice

underscorgan:
How did this go testing with the razor ship? Other than my one nitpick above I'm +1 if shipping went well \o/

melissa:
@underscorgan there are a few things that need to get cleaned up still

mckern:
We shook out a single (important) issue while shipping Razor with this patchset today: `#rsync_from` included the flag `--ignore-existing`, which led to signed RPMs not being copied back to the shipping server... and unsigned RPMs being deployed. They were only live for a few minutes while we worked out what happened, and they were replaced shortly after. `.deb`'s and tarballs were unaffected because of how they're signed (detached signatures and signed metadata are both additive; signing them doesn't require replacing existing files). This behavior has been corrected, and a new spec test has been added to hopefully help prevent regression.

Of note: this does not address any of the long-standing permission problems that the use of the staging server has introduced, which we did encounter today. Fixing that and implementing Aptly are next on the big bucket list of making shipping better.

lib/packaging/config/params.rb
mckern: Fixed. Good catch. Thanks.

mckern:
(Updated to reflect the duplicate config key @underscorgan flagged)

stahnma:
If I see @underscorgan give this a :+1: , I'll merge.

lib/packaging/util/net.rb
```diff
@@ -57,22 +57,101 @@ def remote_ssh_cmd(target, command, capture_output = false)
...
+        target_path: dest,
+        extra_flags: opts[:extra_flags],
+        dryrun: opts[:dryrun],
+        bin: opts[:bin],
+      )
+    end
+
+    # A wrapper method to maintain the existing interface for executing
+    # incoming rsync commands with minimal changes to existing code.
+    def rsync_from(source, origin_host, dest, opts = {})
```
melissa: Didn't you say you were going to make a change so that `rsync_from` would default to no extra flags? It looks like this is still getting `--ignore-existing` from line 87.

lib/packaging/util/net.rb
```diff
@@ -57,22 +57,101 @@ def remote_ssh_cmd(target, command, capture_output = false)
...
+      rsync_exec(
+        source,
+        origin_host: origin_host,
+        target_path: dest,
+        extra_flags: opts[:extra_flags],
```
mckern: @melissa This value is `nil` by default, so when it hits line 87 that carries over (because it's a `merge` without `delete_if`). The spec test validates this:

```
Pkg::Util::Net
  #rsync_from
    should not include the flags '--ignore-existing' by default
    should fail if rsync is not present
    should rsync 'thing' from 'foo@bar' to '/home/foo' with flags '--recursive --hard-links --links --verbose --omit-dir-times --no-perms --no-owner --no-group'
```

underscorgan:
I'm concerned about some of the permissions issues, I had fixed almost all of those last month, so I'd like to make sure there isn't a regression here.

mckern:
@underscorgan there was a bunch of stuff that belonged to `morgan:allstaff` on weth and `root:allstaff` on Burji before we started the ship.

lib/packaging/util/net.rb
melissa: Should we remove line 87 in that case? When will the default on line 87 be used?

mckern: Done!

underscorgan:
I'll give this a :+1: , but I'm going to open a ticket to add some of the things from #503 to the new code paths so that we're managing permissions on staging and production for everything that's having two hops, that seems to have helped with deb/rpm.

melissa:
I'm also :+1:, but I'm nervous merging this before PA 1.3.6 and 1.4.0 are out...

mckern:
There is always another release right around the corner.

melissa:
I'm not saying don't merge this, I'm just saying I'm nervous

stahnma:
@melissa @mckern @underscorgan I am going to merge this. If things break during the next shipment set, revert. I know this was tested.

PR #513, opened 2016-03-01
(maint) Add note about master->stable targets in ci-job-configs

stahnma:


stahnma:
Merging as this is just @branan's PR but updated.

PR #515, opened 2016-03-16
Deferred maintenance for health and safety

mckern:
I've got a laundry list of nitpicks that are making shipping painful. Let's do some laundry.

### Preamble
What I'm aiming for is giving us the ability to selectively skip jobs when running `pl:jenkins:uber_ship` and breaking apart large tasks until they're functionally irreducible (while preserving existing interfaces as much as possible). My hope is that once we've reduced the quantity of work to irreducible tasks that we can then start recomposing the rollups (tasks like `pl:jenkins:uber_ship` or `pl:uber_ship`) into tools more resilient tasks with less churn around the process of getting software someplace our customers can see it.

### Obligatory dumb image for a new PR

laundry-list-pre-treat today-sort tomorrow-wash later-dry soon-fold maybe-iron get-real-vinyl-wall-art-lettering-49f4b322-656d-411e-b871-3defe1f0fb28_600

### DIFF
--- a/lib/packaging/gem.rb
+++ b/lib/packaging/gem.rb
@@ -1,21 +1,39 @@
 module Pkg::Gem
   class << self
+    # This is preserved because I don't want to update the deprecated code path
+    # yet; I'm not entirely sure I've fixed everything that might attempt
+    # to call this method so this is now a wrapper for a wrapper.
     def ship(file)
+      ship_stickler(file)
+      rsync_to_downloads(file)
+      ship_to_rubygems(file)
+    end
+
+    # Ship a Ruby gem file to a Stickler server, because
+    # you've lost the ability to feel joy anymore.
+    def ship_to_stickler(file)
+      Pkg::Util::Tool.check_tool("stickler")
+      Pkg::Util::Execution.ex("stickler push #{file} --server=#{Pkg::Config.internal_gem_host} 2>/dev/null")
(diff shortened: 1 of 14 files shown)


lib/packaging/gem.rb
```diff
@@ -1,21 +1,44 @@
...
+      puts "###########################################"
+      puts "#  Stickler failed, ensure it's installed"
+      puts "#  and you have access to #{Pkg::Config.internal_gem_host}"
+      puts "###########################################"
+    end
+
+    # Use rsync to deploy a file and any associated detached signatures,
+    # checksums, or other glob-able artifacts to an external download server.
+    def rsync_to_downloads(file)
+      Pkg::Util::Net.rsync_to("#{file}*", Pkg::Config.gem_host, Pkg::Config.gem_path, dryrun: ENV['DRYRUN'])
```
melissa: Is this something we're going to change at a later date to deploy to the staging server and sync to the download server?

mckern: I don't know. The whole gem shipping process is kind of up in the air right now. It will probably change when the whole of how we ship changes.

melissa:
:+1: we need to get `*_staging_server` set correctly, but that can be in another PR

melissa:
:+1:  still

stahnma:
Should we remove the don't merge label?

PR #519, opened 2016-03-19
Use definitions from file_mapping.yaml where possible

ccaviness:
@apple_bindir and @apple_libdir are manually entered, but can be generated using data from file_mapping.yaml.

### DIFF
--- a/tasks/puppet-agent-release-tickets.rake
+++ b/tasks/puppet-agent-release-tickets.rake
@@ -112,6 +112,15 @@ Lists JIRA tickets have a git commit, but the ticket is not resolved. Usually th
   * Make note in a comment of any tickets in git not found in JIRA (or visa versa) with an explanation of why the ticket is in that state.
 DOC
 
+  description[:create_release_branch] = <<-DOC
+Create the release branch for the release and switch PE promotions on the PE release being targeted to the release branch in ci-job-configs.
+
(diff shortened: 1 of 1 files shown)


tasks/apple.rake
```diff
@@ -30,9 +30,9 @@ task :setup do
                            Pkg::Config.version.split('.')[2].split('-')[0].split('rc')[0]
   @pm_restart            = 'None'
   @build_date            = Time.new.strftime("%Y-%m-%dT%H:%M:%SZ")
-  @apple_bindir          = '/usr/bin'
+  @apple_bindir          = @source_files['directories']['bin']
   @apple_sbindir         = '/usr/sbin'
```
ccaviness: Except for cleaning up _really_ old installs (in the preflight scripts), I don't think @apple_sbindir is necessary any longer.

tasks/apple.rake
```diff
@@ -30,9 +30,9 @@ task :setup do
...
-  @apple_libdir          = '/usr/lib/ruby/site_ruby'
+  @apple_libdir          = @source_files['directories']['lib']
   @apple_old_libdir      = '/usr/lib/ruby/site_ruby/1.8'
   @apple_docdir          = '/usr/share/doc'
```
ccaviness: If the preflights are changed, this could possibly be sourced from `@source_files['files']['[A-Z]*']`?

PR #521, opened 2016-03-22
(RE-4367) Add msi signing

melissa:
This commit adds in the variables and automation needed to sign msi
packages on the windows signer.

All signing needs to happen as Administrator. Individual users must have
their public key in the authorized_users file.


lib/packaging/msi.rb
```diff
@@ -0,0 +1,38 @@
...
+      Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, %Q(for msi in #{msis.map { |d| File.basename(d) }.join(" ")}; do
+        "/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" sign \
+          -n "Puppet" -i "http://www.puppetlabs.com" \
+          -h sha1 \
+          -pkcs12 "#{Pkg::Config.msi_signing_cert}" \
+          -pass "#{Pkg::Config.msi_signing_cert_pw}" \
+          -t "http://timestamp.verisign.com/scripts/timstamp.dll" \
+          -in "C:/#{work_dir}/$msi" \
+          -out "C:/#{work_dir}/signed-$msi"
+        "/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" sign \
```
shrug: Can you add a comment about why we're signing twice?

melissa: Good point! Done

McdonaldSeanp:
:+1:

PR #523, opened 2016-03-28
(MAINT) Remove short-sighted DRY-RUN hack

mckern:
This hack totally breaks git tag parsing if you try to use DRYRUN.
It's being pulled out until Pkg::Util::Execution#ex can be refactored.


McdonaldSeanp:
:+1:

PR #524, opened 2016-03-28
(maint) Add useful output to fetch task

shrug:
Currently, when retrieving the build_data fails for
some reason, the error message is incomplete and
doesn't provide the necessary information to troubleshoot.

This adds some (hopefully) useful output.


underscorgan:
:+1:

PR #528, opened 2016-04-01
(maint) add additional architectures available for deb repo generation

ScottGarman:
This enables the creation of powerpc deb repos, which we need now
that we're including huaweios builds for that platform.


stahnma:
Does that fix a problem? If so, how? Why?

mckern:
This looks like it'll solve the immediate problem. :+1:

PR #530, opened 2016-04-04
(maint) Improvements to IPS workflow

underscorgan:
We were doubly rsyncing p5p packages to rama. We should only be sending
them out once. Also, trying to create two repos in the same path in
immediate succession may be causing weird caching errors in the repo
management commands needed for signing. Since we're cleaning up the
workspace anyways after each run let's set it up in a new directory too.


shrug:
:+1: 
mei-totoro

mckern:
ants-in-my-eyes

PR #533, opened 2016-04-11
(RE-7003) Ship nightlies with symlink to latest repos

shermdog:
When shipping nightlies that use a whitelist (such as puppet-agent) the shipped artifacts were unique directories.  This was different than how non-whitelisted projects are shipped (such as puppetdb).  This update ensures that the shipped -latest has a symlink to the appropriate latest set.

```
[root@ilhoe2uxhhro2t1 puppet-agent-latest]# ls -la
total 0
drwxr-xr-x 3 root root  39 Apr 11 20:14 .
drwxr-xr-x 6 root root 108 Apr 11 21:04 ..
drwxr-xr-x 4 root root  28 Apr 11 18:55 repo_configs
lrwxrwxrwx 1 root root  62 Apr 11 20:14 repos -> ../puppet-agent/740f20af50043ff31bde943f0d273f4cd40e4acb/repos
```

Tested puppet-agent only ships the whitelisted FOSS platforms and that puppetdb ships all.
```
21:50 $ diff -rq pkg/puppet-agent-latest/repos tmp/pkg/puppet-agent-latest/repos
Only in pkg/puppet-agent-latest/repos: 740f20af50043ff31bde943f0d273f4cd40e4acb.yaml
Only in pkg/puppet-agent-latest/repos: aix
Only in pkg/puppet-agent-latest/repos/apt: lucid
Only in pkg/puppet-agent-latest/repos/apt: squeeze
Only in pkg/puppet-agent-latest/repos: cisco-wrlinux
Only in pkg/puppet-agent-latest/repos: deb
Only in pkg/puppet-agent-latest/repos/el: 4
Only in pkg/puppet-agent-latest/repos/fedora: f23
Only in pkg/puppet-agent-latest/repos: puppet-agent-1.4.1-79-signing_bundle.tar.gz
Only in pkg/puppet-agent-latest/repos: sles
Only in pkg/puppet-agent-latest/repos: solaris
```

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -160,15 +160,26 @@ namespace :pl do
         include_paths = ["./"]
       end
 
-      # Ship it to the target for consumption
-      # First we ship the latest and clean up any repo-configs that are no longer valid with --delete-after
-      Pkg::Util::Execution.ex(%(find #{include_paths.map { |path| "pkg/#{Pkg::Config.project}-latest/#{path}" }.join(' ') } | sort > include_file_latest))
-      Pkg::Util::Execution.ex(%(mkdir -p tmp_latest && tar -T include_file_latest -cf - | (cd ./tmp_latest && tar -xf -)))
-      Pkg::Util::Net.rsync_to("tmp_latest/pkg/#{Pkg::Config.project}-latest", target_host, target_basedir, extra_flags: ["--delete-after"])
-      # Then we ship the sha version with default rsync flags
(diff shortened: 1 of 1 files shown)


stahnma:
@haus is this going to break something unintentionally? e.g. was there a reason these weren't already symlinks?

haus:
@stahnma iirc, it was just the easier path at the time. less complexity. I don't think it will break anything.

melissa:
👍

PR #535, opened 2016-04-20
(maint) Use puppet.com for MSI signing

Iristyle:


Iristyle:
ping @melissa

melissa:
@Iristyle have you tested this at all?

melissa:
screen shot 2016-04-20 at 10 10 45 am
screen shot 2016-04-20 at 10 11 05 am
screen shot 2016-04-20 at 10 11 39 am

melissa:
I'm not actually seeing any changes. The email address isn't even showing up anywhere from what I can tell

melissa:
Maybe it's the cert itself that needs to be updated?

stahnma:
@Iristyle ping for answers...

stahnma:
@Iristyle ping agan.... (weekly review of open PRs)

stahnma:
@Iristyle ping agan.... (weekly review of open PRs)

Iristyle:
Sorry I missed your repeated pings ;0

You can't alter the certs used during the signing process - once they're issued, they're set in stone.  To change that we would indeed require a new cert, but given we recently renewed (and it costs 💰), I suspect we'll wait until it expires to make a change there.

The `-i` is strictly an informational switch, and AFAIK it gets written into the MSI comments - see the screenshot below under the details view in Windows Explorer:

screen shot 2016-05-10 at 10 55 21 pm

There are probably other spots that it pops up as well - I don't see any reason to not merge this change.

stahnma:
@McdonaldSeanp have a look here since @melissa is OOO

McdonaldSeanp:
I'm good merging this change 👍

geoffnichols:
I found a [Microsoft doc](Windows Authenticode Portable Executable Signature Format - Microsoft) that appears to have some information about the `1.3.6.1.4.1.311.2.1.12` attribute. 

```
### SpcSpOpusInfo

**SpcSpOpusInfo** is identified by SPC_SP_OPUS_INFO_OBJID (1.3.6.1.4.1.311.2.1.12) and is defined as follows:

    SpcSpOpusInfo ::= SEQUENCE {

programName [0] EXPLICIT SpcString OPTIONAL,

moreInfo [1] EXPLICIT SpcLink OPTIONAL,

    } --#public--

**SpcSpOpusInfo** has two fields:

programName

This field contains the program description:

 If publisher chooses not to specify a description, the **SpcString** structure contains a zero-length program name.

If the publisher chooses to specify a description, the **SpcString** structure contains a Unicode string.

moreInfo

This field is set to an **SPCLink** structure that contains a URL for a Web site with more information about the signer. The URL is an ASCII string.
```

PR #536, opened 2016-04-20
(maint) Create tickets in "Ready for Engineering"

haus:
Previously tickets were created in the default open status, which means
that a human needs to update them to be ready for engineering so that
they appear on agile boards. This commit updates the tickets to be
created in the "Ready for Engineering" state to avoid this manual step.

### DIFF
--- a/lib/packaging/util/jira.rb
+++ b/lib/packaging/util/jira.rb
@@ -156,7 +156,7 @@ module Pkg::Util
 
       if options_hash[:status]
         transition = issue.transitions.build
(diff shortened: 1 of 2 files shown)


underscorgan:
💯

haus:
You can see an example here: https://jira1-test.ops.puppetlabs.net/browse/SERVER-1292

haus:
The first travis with an x looks to be from when i accidentally pushed this branch to puppetlabs instead of my fork.

PR #538, opened 2016-04-25
(maint) Add subtasks for pe_repo during platform removal

jpartlow:
Adds two more subtasks to track deprecating pe_repo platform classes,
and to file a future ticket to remove them completely in the next PE
major.  This came up as a result of our forgetting to deal with pe_repo
during the removal of Debian6, Ubuntu 15.04 and 15.10, and the dance
around upgrades and installations which may have these platform classes
already declared.

### DIFF
--- a/tasks/platform-removal.rake
+++ b/tasks/platform-removal.rake
@@ -28,6 +28,14 @@ DOC
 Platform needs to be removed from PE pipelines.
 DOC
 
+  description[:pe_repo_deprecation] = <<-DOC
+The pe_repo::platforms::<platform> class needs to be deprecated in the puppetlabs-pe_repo module.  It should raise a warning, letting users know that the platform is no longer available and the class should be removed.  It cannot be deleted yet as this would cause compilation failures on upgraded installations which had made use of the removed platform.
+DOC
+
+  description[:pe_repo_removal] = <<-DOC
+Ticket needs to be filed for next major PE release to remove the deprecated platfrom from pe_repo.
+DOC
+
   description[:pa_pipeline] = <<-DOC
 Platform needs to be removed from puppet-agent pipelines.
 DOC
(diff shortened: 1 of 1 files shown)


tasks/platform-removal.rake
```diff
@@ -122,6 +130,22 @@ DOC
...
+    {
+      :short_name   => 'pe_repo_deprecation',
+      :project      => 'PE',
+      :summary      => "Deprecate #{vars[:platform_tag]} from puppetlabs-pe_repo",
+      :description  => description[:pe_repo_deprecation],
+      :story_points => '2',
+      :blocked_by   => ['initial_email'],
+    },
+    {
+      :short_name   => 'pe_repo_removal',
```
melissa: Should we make creating this ticket part of the depreciation work?

Also, we'll need to add descriptions for these tickets. Right now `description[:pe_repo_removal]` and `description[:pe_repo_deprecation]` are going to return empty strings

tasks/platform-removal.rake
```diff
@@ -122,6 +130,22 @@ DOC
       :summary      => "Remove #{vars[:platform_tag]} from PE integration pipelines",
       :description  => description[:pe_pipeline],
       :story_points => '3',
+      :blocked_by   => ['initial_email','pe_repo_deprecation'],
```
melissa: Rubocop wants a space after the comma

jpartlow:
Oops, I have offensively omitted a space after a comma, one sec...

tasks/platform-removal.rake
jpartlow: Do you mean add it as a note in the pe_repo_deprecation ticket description that you should create the removal ticket in addition to performing the deprecation work?

That could work; it's the sort of thing we tend to forget about though.  I think it's better to have them as separate steps so it doesn't get lost.

melissa: Yeah, that's what I was suggesting. But, if you believe it's better to have more tickets just to be sure we don't forget that step, than I'm totally on board.

melissa:
👍 looks good to me!

tasks/platform-removal.rake
jpartlow: Based on our (integrationt team's) track record with deprecations, I think so... :)

PR #539, opened 2016-04-26
(maint) Add the ability to push gems to both destinations

melissa:
We need to be able to ship gems to both nexus and stickler. This commit
splits out the internal_gem_host to allow us to either default to the
original value or to assign each separately.


melissa:
Required https://github.com/puppetlabs/build-data/pull/58 to work successfully

melissa:
cc/ @underscorgan @mckern @shrug

underscorgan:
👍

PR #540, opened 2016-04-26
(RE-7052) Copy build params into the specification

joshcooper:
Commit 72751690 added the ability to set gem_required_ruby_version (and
rubygems), but I forgot to git add the change to the gem task to copy
the build params to the generated gem specification.


underscorgan:
👍

mckern:
So many `.nil?`s.

PR #541, opened 2016-04-26
(maint) Missing trailing slash in rsync command

underscorgan:
Windows was shipping to /opt/downloads/windows/windows instead of
/opt/downloads windows due to a missing trailing slash in the source
path.


melissa:
👍

PR #542, opened 2016-04-27
(maint) RPM rsyncs failed while shipping today

underscorgan:
We don't want all the options --archive brings in, we actually only want
--recursive and --links. --archive was causing issues with timestamps on
symlinks. Manually syncing with --recursive and --links instead of
--archive allowed the rsync to run successfully.

### DIFF
--- a/lib/packaging/rpm/repo.rb
+++ b/lib/packaging/rpm/repo.rb
@@ -52,7 +52,8 @@ module Pkg::Rpm::Repo
       # fail due to permission errors.
       options = %w(
         rsync
-        --archive
+        --recursive
+        --links
         --hard-links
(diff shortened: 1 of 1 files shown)


lib/packaging/rpm/repo.rb
```diff
@@ -65,7 +66,6 @@ def repo_deployment_command(origin_path, destination_path, destination, dryrun =
         --no-perms
         --no-owner
         --no-group
-        --delay-updates
```
underscorgan: This isn't actually removing --delay-updates, it's specified twice in the option list so this just cleans that up a bit.

mckern:
👍

PR #543, opened 2016-04-28
(CPR-342) fix pl:agent_tickets and pl:platform_removal tasks

shermdog:
The copy pasta in the pl:platform_removal task had conflicting anonymous methods.  There was also a race condition introducted in an prior commit.


ScottGarman:
:+1:  thanks for digging into this.

PR #544, opened 2016-05-03
(maint) Eradicate bin/uber_ship

underscorgan:
The goal of this PR is to make it possible to ship ezbake packages with the same automation as vanagon projects.


stahnma:
Squash, rebase and cleanup commits?

McdonaldSeanp:
👍

PR #546, opened 2016-05-09
(RE-6989) Add component tag check

McdonaldSeanp:
This adds a component tagging check for packaging. This (for now) is only a
human check and not a barrier for any CI jobs or builds using
pl:jenkins:uber_build. This commit adds the vanagon:check_tags and
vanagon:build_all rake tasks. The latter of which should replace the human
invocation of pl:jenkins:uber_build when we ship vanagon projects. The new
vanagon:build_all task is really just a wrapper for uber_build that adds
check_tags and a prompt.

Basically this is a human checklist, meant to notify the user if something
"looks tagged". It will not prevent shipment or builds, it will simply output
what "looks tagged" and then ask if this looks correct.

### DIFF
--- a/lib/packaging/util/file.rb
+++ b/lib/packaging/util/file.rb
@@ -32,6 +32,12 @@ module Pkg::Util::File
       end
     end
 
+    # Returns an array of all files with #{ext} inside #{dir}
+    def files_with_ext(dir, ext)
+      Dir.glob("#{dir}/**/*#{ext}")
+    end
+
+
     def file_exists?(file, args = { :required => false })
       exists = File.exist? file
       if !exists and args[:required]
--- a/lib/packaging/util/git_tags.rb
+++ b/lib/packaging/util/git_tags.rb
@@ -1,0 +1,72 @@
+module Pkg::Util
+  class Git_tag
+    attr_reader :address, :ref, :ref_name, :ref_type, :branch_name
+
+    GIT = Pkg::Util::Tool::GIT
+    DEVNULL = Pkg::Util::OS::DEVNULL
+
+    # A SHA1 sum is 20 characters long, but Git will match on
+    # the first ~8 or so. And 8 is long enough for fun test sums
+    # like 'cafebeef' or 'deadfeed`.
+    SHA1 = /[0-9A-F]{8,20}/i
+
+    def initialize(address, reference)
+      @address = address
+      @ref = reference
+      parse_ref!
+    end
+
+    # Parse ref in one of three ways: if the ref is already in a good format just grab
+    # the ref type from the string. if it's not, check if it's a sha, if that is true
+    # then list it as a sha. finally if it's neither of those fetch the full ref and
+    # parse that.
+    def parse_ref!
+      if ref?
+        split_ref(ref)
+      elsif sha?
+        @ref_type = "sha"
+      else
+        split_ref(fetch_full_ref)
+      end
+    end
+
+    # Split the ref based on slashes, set ref_name and ref_type based on the last two
+    # items from the split. i.e. refs/tags/1.1.1 would return:
+    #       @ref_name => 1.1.1       @ref_type => tags
+    def split_ref(ref)
+      ref_parts = ref.split('/', 3)
+      @ref_name = ref_parts.pop
+      @ref_type = ref_parts.pop
+      [@ref_type, @ref_name]
+    end
+
+    # Fetch the full ref using ls-remote, this should raise an error if it returns non-zero
+    # because that means this ref doesn't exist in the repo
+    def fetch_full_ref
+      Pkg::Util::Execution.ex("#{GIT} ls-remote --tags --heads --exit-code #{address} #{ref}").split.last
+    rescue RuntimeError
+      raise "ERROR : Not a ref or sha!"
+    end
+
+    def branch_name
+      branch? ? ref_name : nil
+    end
+
+    def ref?
+      `#{GIT} check-ref-format #{ref} >#{DEVNULL} 2>&1`
+      $?.success?
+    end
+
+    def branch?
+      ref_type.downcase == "heads"
+    end
+
+    def tag?
+      ref_type.downcase == "tags"
+    end
+
+    def sha?
+      !!(ref =~ SHA1)
+    end
+  end
+end
--- a/lib/packaging/util/version.rb
+++ b/lib/packaging/util/version.rb
@@ -1,4 +1,5 @@
 # Utility methods used for versioning projects for various kinds of packaging
+require 'json'
 
 module Pkg::Util::Version
   class << self
@@ -311,5 +312,33 @@ module Pkg::Util::Version
       # ...and write it back on out.
       File.open(version_file, 'w') { |f| f.write contents }
     end
+
+    # Human readable output for json tags reporting. This will load the
+    # input json file and output if it "looks tagged" or not
+    #
+    # @param json_data [hash] json data hash containing the ref to check
+    def report_json_tags(json_data)
+      puts "component: " + File.basename(json_data["url"])
+      puts "ref: " + json_data["ref"].to_s
+      if tagged?(json_data["url"], json_data["ref"].to_s)
+        tagged = "Tagged? [ Yes ]"
+      else
+        tagged = "Tagged? [ No  ]"
+      end
+      col_len = (ENV["COLUMNS"] || 70).to_i
+      puts format("\n%#{col_len}s\n\n", tagged)
+      puts ("*" * col_len)
+    end
+
+    # Reports if a ref and it's corresponding git repo points to
+    # a git tag.
+    #
+    # @param url [string] url of repo grabbed from json file
+    # @param ref [string] ref grabbed from json file
+    def tagged?(url, ref)
+      reference = Pkg::Util::Git_tag.new(url, ref)
+      reference.tag?
+    end
+
   end
 end
--- a/tasks/vanagon.rake
+++ b/tasks/vanagon.rake
@@ -1,0 +1,35 @@
+namespace :vanagon do
+  # vanagon:check_tags is used to report the tagged status of vanagon components
+  # with "ref" fields indicated in their json files. This function does not create
+  # any barriers or stop execution, it's only a report
+  desc "Check for tagged components in vanagon projects"
+  task :check_tags do
+    # only available to vanagon projects
+    unless Pkg::Config.vanagon_project
+      puts "This is not a vanagon project"
+      return
+    end
+    puts "*" * (ENV["COLUMNS"] || 70).to_i
+    Pkg::Util.in_project_root do
+      Pkg::Util::File.files_with_ext("configs/components", ".json").each do |json_file|
+        json_data = Pkg::Util::Misc.load_from_json(json_file)
+        # Don't report on anything without a ref
+        if json_data["ref"].is_a?(String)
+          Pkg::Util::Version.report_json_tags(json_data)
+        end
+      end
+    end
+  end
+
+  # This is just a wrapper for uber_build that runs the check_tags report and
+  # pauses for input, mainly so we can manually check for tagged components
+  # before spending time on a build
+  desc "Run component tag reports, if the report is okay, run an uber_build"
+  task :build_all, [:poll_interval] do |t, args|
+    args.with_defaults(:poll_interval => 0)
+    Rake::Task["vanagon:check_tags"].invoke
+    puts "Does this look correct?"
+    return unless Pkg::Util.ask_yes_or_no
+    Rake::Task["pl:jenkins:uber_build"].invoke(args.poll_interval)
+  end
+end
--- a/lib/packaging/util.rb
+++ b/lib/packaging/util.rb
@@ -20,6 +20,7 @@ module Pkg::Util
   require 'packaging/util/tool'
   require 'packaging/util/rake_utils'
   require 'packaging/util/version'
+  require 'packaging/util/git_tags'
 
   def self.boolean_value(var)
     return TRUE if var == TRUE || ( var.is_a?(String) && ( var.downcase == 'true' || var.downcase =~ /^y$|^yes$/))
--- a/lib/packaging/util/misc.rb
+++ b/lib/packaging/util/misc.rb
@@ -21,5 +21,21 @@ module Pkg::Util::Misc
 
       search_string
     end
+
+    # Loads and parses json from a file. Will treat the keys in the
+    # json as methods to invoke on the component in question
+    #
+    # @param file [String] Path to the json file
+    # @raise [RuntimeError] exceptions are raised if there is no file, if it refers to methods that don't exist, or if it does not contain a Hash
+    def load_from_json(file)
+      data = JSON.parse(File.read(file))
+      unless data.is_a?(Hash)
+        raise "Hash required. Got '#{data.class}' when parsing '#{file}'"
+      end
+      # We explicity return data here b/c the unless clause above will cause the
+      # Function to return nil.
+      #               -Sean P. M. 05/11/2016
+      data
+    end
   end
 end
--- a/lib/packaging/util/rake_utils.rb
+++ b/lib/packaging/util/rake_utils.rb
@@ -102,6 +102,7 @@ module Pkg::Util::RakeUtils
         'version.rake',
         'z_data_dump.rake',
         'config.rake',
+        'vanagon.rake',
       ]
 
       tasks.each do |task|
--- a/spec/fixtures/configs/components/test_file.json
+++ b/spec/fixtures/configs/components/test_file.json
@@ -1,0 +1,1 @@
+{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "refs/tags/0.6.2"}
--- a/spec/fixtures/configs/components/test_file_not_tagged.json
+++ b/spec/fixtures/configs/components/test_file_not_tagged.json
@@ -1,0 +1,1 @@
+{"url": "git://github.com/puppetlabs/leatherman.git", "ref": "4eef05389ebf418b62af17406c7f9f13fa51f975"}
--- a/spec/lib/packaging/util/file_spec.rb
+++ b/spec/lib/packaging/util/file_spec.rb
@@ -51,6 +51,17 @@ describe "Pkg::Util::File" do
     end
   end
 
+  describe "#files_with_ext" do
+    it "returns nothing if there are no files with that extension" do
+      Pkg::Util::File.files_with_ext("./spec/fixtures/configs/components", ".fake").should be_empty
+    end
+
+    it "returns only the files with that extension" do
+      expect(Pkg::Util::File.files_with_ext("./spec/fixtures/configs/components", ".json")).to include("./spec/fixtures/configs/components/test_file.json")
+      expect(Pkg::Util::File.files_with_ext("./spec/fixtures/configs/components", ".json")).to include("./spec/fixtures/configs/components/test_file_2.json")
+    end
+  end
+
   describe "#install_files_into_dir" do
     it "selects the correct files to install" do
       Pkg::Config.load_defaults
--- a/spec/lib/packaging/util/git_tag_spec.rb
+++ b/spec/lib/packaging/util/git_tag_spec.rb
@@ -1,0 +1,36 @@
+# -*- ruby -*-
+require 'spec_helper'
+
+describe "Pkg::Util::Git_tag" do
+  context "parse_ref!" do
+    it "fails for a ref that doesn't exist'" do
+      expect { Pkg::Util::Git_tag.new("git://github.com/puppetlabs/leatherman.git", "garbagegarbage") }.to raise_error(RuntimeError, "ERROR : Not a ref or sha!")
+    end
+  end
+
+  context "sha?" do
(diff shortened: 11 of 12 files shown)


lib/packaging/util/file.rb
```diff
@@ -32,6 +32,16 @@ def directories(dir)
       end
     end
 
+    # Returns an array of all files with #{ext} inside #{dir}
+    def files_with_ext(dir, ext)
+      if File.directory?(dir)
+        Dir.chdir(dir) do
+          Dir.glob("*").select { |entry| File.extname(entry) == ext }
```
mckern: Let's talk about double splats.

```ruby
Dir.glob["#{dir}/**/*.#{ext}"]
```

mckern: (You can probably also lose the `chdir` unless you're trying to construct relative pathnames?)

mckern:
```ruby
>> dir = "tmp"
=> "tmp"
>> ext = ".string"
=> ".string"
=> nil
>> puts %(Dir.glob["#{dir}/**/*#{ext}"])
Dir.glob["tmp/**/*.string"]
>> Dir.glob("#{dir}/**/*#{ext}")
=> []
>>
```

McdonaldSeanp:
The glob returns an empty array if I use the double splat instead of select. So I took out the chdir and directory? but left that in

lib/packaging/util/file.rb
```diff
@@ -32,6 +32,12 @@ def directories(dir)
       end
     end
 
+    # Returns an array of all files with #{ext} inside #{dir}
+    def files_with_ext(dir, ext)
+      Dir.glob("#{dir}/*").select { |entry| File.extname(entry) == ext }
```
mckern: Methods that return arrays should return empty arrays if there's nothing to populate the array with. This will also return an empty array if nothing is found. This implementation holds to that (as does the double-splat in the `Dir.glob` I suggested), but this one isn't recursive (The double-splat just recurses.). You'd get the same result with `Dir.glob("#{dir}/*#{ext}")` if you don't want to recurse.

mckern: Follow-up on what `Dir.glob().select {}` returns when there's nothing to select:

```ruby
>> dir = "tmp"
=> "tmp"
>> File.exists? dir
=> false
>> ext = ".string"
=> ".string"
>> Dir.glob("#{dir}/*").select { |entry| File.extname(entry) == ext }
=> []
>> Dir.glob("#{dir}/**/*.#{ext}")
=> []
>> Dir.glob("#{dir}/*.#{ext}")
=> []
```

lib/packaging/util/version.rb
```diff
@@ -2,6 +2,7 @@
 
 module Pkg::Util::Version
   class << self
+    require 'json'
```
mckern: I would prefer we not require in the Module body if possible. `json` is part of Stdlib, so it's safe to require practically anywhere. Outside of the Module body is a good start.

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,32 @@ def versionbump(workdir = nil)
       # ...and write it back on out.
       File.open(version_file, 'w') { |f| f.write contents }
     end
+
+    def report_json_tag(path_to_json)
+      json_data = load_from_json(path_to_json)
+      output_string = " tagged at: " + json_data["ref"].to_s
+      output_string = output_string.ljust(60)
```
mckern: Don't use hard justifications. Start by making this output into something simple and machine readable, so you can wrap a formatter around it later if it matters.

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,32 @@ def versionbump(workdir = nil)
...
+      if /refs\/tags\/[0-9]*\.[0-9]*\.[0-9]*/.match(json_data["ref"].to_s)
+        output_string += " Tagged? [ Yes ]"
+      else
+        output_string += " Tagged? [ No  ]"
+      end
+    end
```
mckern: This function is busy and it's doing too much(matching/processing data, formatting & output). Let's unpack it piece by piece:

- Don't use hard justifications for output formatting; calculate them whenever possible and then feed those values into pre-formatted strings and `sprintf`. If you **really** need to do a lot of it, look into a dedicated CLI library like Highline.
- I'd probably start by making this output into something simple and machine readable, and then wrapping a formatter around it later if it matters. Even if you keep the extra-hardcoded formatting right now, it should probably still cut out into a method that operates over a placeholder instance variable.
- If you're wedded to solving "check this ref" with a regex, I think sometimes they read better if you wrap 'em in `%r{}` instead of `/ /` (at least when where's slashes in them). You won't have to escape forward slashes if you use the percent-defined version.
- You've got your "thing to match" and "match criteria" backwards. The plain string should come first, then match criteria:

```ruby
>> "refs/tags/0.9.10".match(/refs\/tags\/[0-9]*\.[0-9]*\.[0-9]*/)
=> #<MatchData "refs/tags/0.9.10">`)
>>
```

- It looks like you're really string matching (not pulling pieces out of the ref) so `#match` is heavyweight. Consider `#include?`:

```ruby
# Actual Leatherman ref from the pe-client-tools' leatherman.json; note that it's not
# fully qualified like the refs we use in puppet-agent. Git will accept a branch name,
# a tag, a fully qualified ref, or a SHA and we should strive to be accommodating.
>> ref = "0.6.2"
=> "0.6.2"
>> # Actual Leatherman refs from `git ls-remote`
>> puts %x(git ls-remote --tags https://github.com/puppetlabs/leatherman.git)
d6ebb9c49122fa6bd9d1b243ebe46dc031ab080e  refs/tags/0.6.2
565f1e811c53fc6917441fc4dcf8ea3cbb4816aa  refs/tags/0.6.2^{}
929098f3ebde88f8eec8e820b995e80f93620df2  refs/tags/0.6.3
4017302d316c7ce67a3f85c8b0b527c668ee4052  refs/tags/0.6.3^{}
>> >> "refs/tags/0.6.3".include? ref
=> false
>> "refs/tags/0.6.2".include? ref
=> true
>>
```
- Since the JSON data seems to be read-only, I'd consider wrapping the loading of JSON data into a function tied to an instance variable. Then you can memoize the instance variable and treat the function the same way you'd treat a local variable:

```ruby
def json_data
  # Try to load from an Env. variable file location first,
  # then from a default location
  @json_data ||= load_json_from_file(ENV["SOMEFILE"] || "somefile.json")
end

def report_json_tag
  output_string = " tagged at: #{json_data["ref"]}"
  ### snip
```

### But now that you hate me with all of your might!

Consider actually parsing the ref. I just wrote some code for Vanagon to do that (for the Git repos used as component sources), and it'll probably be a little more reusable than this. It's probably not drop-in perfect for your needs, but here's an example I hacked out of the pending work:

```ruby
class Git
  attr_reader :address, :ref, :ref_name, :ref_type, :branch_name

  # A SHA1 sum is 20 characters long, but Git will match on
  # the first ~8 or so. And 8 is long enough for fun test sums
  # like 'cafebeef' or 'deadfeed`.
  SHA1 = /[0-9A-F]{8,20}/i

  def initialize(address, reference)
    @address = address
    @ref = reference
    parse_ref!
  end

  def parse_ref!
    _, @ref_type, @ref_name = ref.split('/', 3)
    ref_type
  end

  def ref_type
    @ref_type ||= mangle_ref
  end

  def mangle_ref
    sha? ? "sha" : "tags"
  end

  def branch_name
    branch? ? ref_name : nil
  end

  def ref?
    return false if sha?
    !!(ref_type && ref_name)
  end

  def branch?
    ref_type.downcase == "heads"
  end

  def tag?
    ref_type.downcase == "tags"
  end

  def sha?
    !!(ref =~ SHA1)
  end
end
```

Now, brace yourself for the next novella I'm writing for the `load_from_json` function.

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,32 @@ def versionbump(workdir = nil)
...
+
+    # Loads and parses json from a file. Will treat the keys in the
+    # json as methods to invoke on the component in question
```
mckern: Looks like Vanagon copypasta

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,32 @@ def versionbump(workdir = nil)
...
+    #
+    # @param file [String] Path to the json file
+    # @raise [RuntimeError] exceptions are raised if there is no file, if it refers to methods that don't exist, or if it does not contain a Hash
+    def load_from_json(file)
+      if File.exists?(file)
```
mckern: Skip this. Just try to parse the data. It'll raise `Errno::ENOENT: No such file or directory`, which is the right error for this situation.

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,32 @@ def versionbump(workdir = nil)
...
+        data = JSON.parse(File.read(file))
+        raise "Hash required. Got '#{data.class}' when parsing '#{file}'" unless data.is_a?(Hash)
```
mckern: This should probably be `raise RuntimeError, blahblahblah`. It's a little more specific than plain old `StandardError`.

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,32 @@ def versionbump(workdir = nil)
...
+      else
+        fail "Cannot load component data from '#{file}'. It does not exist."
```
mckern: If you lose the redundant sanity check at the top of this method, you can ditch this too.

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,32 @@ def versionbump(workdir = nil)
...
+      end
+      data
```
mckern: As goes the redundant sanity check, so goes the explicit return value.

tasks/jenkins_dynamic.rake
```diff
@@ -15,9 +15,25 @@ namespace :pl do
   namespace :jenkins do
     desc "Dynamic Jenkins UBER build: Build all the things with ONE job"
     task :uber_build, [:poll_interval] => "pl:fetch" do |t, args|
-      args.with_defaults(:poll_interval => 0)
+      args.with_defaults(:poll_interval => 0, :check_component_tags => false)
       poll_interval = args.poll_interval.to_i
 
+      # This is a check for vanagon component tags, and only a report.
+      # The report itself does not stop the build process.
+      if ENV['CHECK_VANAGON_COMPONENT_TAGS'] && Pkg::Config.vanagon_project
```
mckern: Should this be moved into its own Rake task? And should that task move a lot of this formatting into dedicated output functions?

lib/packaging/util/version.rb
McdonaldSeanp: Not sure why I did that.

lib/packaging/util/version.rb
McdonaldSeanp: I like these ideas, PR update to come

lib/packaging/util/version.rb
McdonaldSeanp: yup

lib/packaging/util/version.rb
McdonaldSeanp: 👍

lib/packaging/util/version.rb
McdonaldSeanp: 👍

lib/packaging/util/version.rb
McdonaldSeanp: 👍

lib/packaging/util/version.rb
McdonaldSeanp: 👍

tasks/jenkins_dynamic.rake
McdonaldSeanp: So that was my original thought, but wanted to see what others thought before adding yet another rake task to packaging. I'll add one to this PR and we can discuss

lib/packaging/util/version.rb
McdonaldSeanp: Rubocop does not like adding the RuntimeError explicitly, it says it's redundant

mckern: Oh, you're right. `raise` throws `RuntimeError` by default, not `StandardError`. OK.

McdonaldSeanp:
For some reason if I don't explicitly state the return in load_json it returns nil, so I left that in

mckern:
@McdonaldSeanp `#load_from_json` returns `nil` without the explicit return because of the `unless`. Since the conditional in `unless` never returns anything, that's the last value to return.

It's gross, but we can probably leave it as is; building a whole new "read this junk from JSON otherwise do a thing" set of methods for Packaging isn't how you or I want to spend the afternoon. Can you add a comment explaining why you made this choice, why it works that way, and date it? I want us to know what we were thinking when we need to update this in 6-12 months.

McdonaldSeanp:
Yeah, I'll leave that comment

tasks/vanagon.rake
```diff
@@ -0,0 +1,24 @@
...
+
+  desc "Run component tag reports, if the report is okay, run an uber_build"
+  task :build_all, [:poll_interval] do |t, args|
+    args.with_defaults(:poll_interval => 0)
+    Rake::Task["vanagon:check_tags"].invoke
+    puts "Does this look correct?"
+    return unless Pkg::Util.ask_yes_or_no
+    Rake::Task["pl:jenkins:uber_build"].invoke(args.poll_interval)
+  end
+end
```
underscorgan: Missing newline

underscorgan:
It looks like the commit message for this is out of date?

Also, do you have an example of what the output looks like running against puppet-agent? (just curious)

McdonaldSeanp:
snip20160511_9

McdonaldSeanp:
Also updated commit message

tasks/vanagon.rake
McdonaldSeanp: fixed

lib/packaging/util/version.rb
```diff
@@ -311,5 +312,35 @@ def versionbump(workdir = nil)
...
+    def report_json_tags(path_to_json)
+      json_data = Pkg::Util::Misc.load_from_json(path_to_json)
+      puts "component: " + File.basename(json_data["url"])
+      puts "tag: " + json_data["ref"].to_s
+      if tagged?(json_data["ref"].to_s)
+        tagged = "Tagged? [ Yes ]"
+      else
+        tagged = "Tagged? [ No  ]"
+      end
+      puts format("%70s", tagged)
```
mckern: You know that `COLUMNS` is (almost always) an environment variable in your terminal, right? You can use them to dynamically calculate terminal width and figure out how many spaces **AND** asterisks to print.

```ruby
>> spaces = "\s" * (ENV["COLUMNS"].to_i || 70)
=> "                                                                                                                                        "
>> puts "#{spaces}magic!"
                                                                                                                                        magic!
=> nil
```

McdonaldSeanp: I did not

McdonaldSeanp:
I cleaned up the output a little and made it not report on json files without a ref: 
snip20160512_10

underscorgan:
This still has a false negative but not sure how easy that would be to correct for without getting false positives? (cfpropertylist)

McdonaldSeanp:
Yeah, what would be ideal is if we could enforce a standard, like always using refs/tags/<tag>

mckern:
@McdonaldSeanp Aaaaaand that's why I wrote you a Git class that parsed refs.

underscorgan:
Or maybe parse for a SHA-1 ref?

mckern:
@underscorgan that's in the class!

```ruby
class Git
  attr_reader :address, :ref, :ref_name, :ref_type, :branch_name

  # A SHA1 sum is 20 characters long, but Git will match on
  # the first ~8 or so. And 8 is long enough for fun test sums
  # like 'cafebeef' or 'deadfeed`.
  SHA1 = /[0-9A-F]{8,20}/i

  def initialize(address, reference)
    @address = address
    @ref = reference
    parse_ref!
  end

  def parse_ref!
    _, @ref_type, @ref_name = ref.split('/', 3)
    ref_type
  end

  def ref_type
    @ref_type ||= mangle_ref
  end

  def mangle_ref
    sha? ? "sha" : "tags"
  end

  def branch_name
    branch? ? ref_name : nil
  end

  def ref?
    return false if sha?
    !!(ref_type && ref_name)
  end

  def branch?
    ref_type.downcase == "heads"
  end

  def tag?
    ref_type.downcase == "tags"
  end

  def sha?
    !!(ref =~ SHA1)
  end
end
```

McdonaldSeanp:
Yeah, ryan had already done that, the reason I decided to leave in the regex in lieu of using the class was from a conversation I had with rob about enforcing the refs/tags/#{tag} bit. Whatever is fine with me but I liked the idea of parsing for less things rather than looking for more things.

McdonaldSeanp:
I'll redo this with the class, gimme a minute

mckern:
I think enforcing a ref standard is going to blow up in our faces. It's orthogonal to Git's workflow (poke around in the Git docs for the term "commitish"), and it might limit what we need to look for but it's also going to require writing tooling to enforce it. I'd rather just write tooling to parse the valid options, make reasonable assumptions about what *kind* of commitish a ref is, and then let Git do the heavy lifting.

McdonaldSeanp:
Again, totally fine with me. It was all just based on a conversation.

McdonaldSeanp:
New set of changes, new output: 
snip20160512_11

McdonaldSeanp:
I added some spec tests to the git_tag class and made everything a little shinier should be pretty close at this point

lib/packaging/util/git_tags.rb
```diff
@@ -0,0 +1,71 @@
...
+      ref_parts = ref.split('/', 3)
+      @ref_name = ref_parts.pop
+      @ref_type = ref_parts.pop
+    end
+
+    # Fetch the full ref using ls-remote, this should raise an error if it returns non-zero
+    # because that means this ref doesn't exist in the repo
+    def fetch_full_ref
+      Pkg::Util::Execution.ex("#{GIT} ls-remote --tags --heads --exit-code #{address} #{ref}").split.last
+    rescue
```
mckern: I'd probably only rescue `RuntimeError` here; I know it's the general error class but `rescue

lib/packaging/util/git_tags.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    # Split the ref based on slashes, set ref_name and ref_type based on the last two
+    # items from the split. i.e. refs/tags/1.1.1 would return:
+    #       @ref_name => 1.1.1       @ref_type => tags
+    def split_ref(ref)
+      ref_parts = ref.split('/', 3)
+      @ref_name = ref_parts.pop
+      @ref_type = ref_parts.pop
+    end
```
mckern: As originally written this used parallel assignment and returned an array. As written now it uses serial assignment and returns a string.

```ruby
>> def split_ref_pop(ref)
|      ref_parts = ref.split('/', 3)
|      @ref_name = ref_parts.pop
|      @ref_type = ref_parts.pop
|  end
=> :split_ref_pop
>>
>> def split_ref_multiple(ref)
|      _, @ref_type, @ref_name = ref.split('/', 3)
|  end
=> :split_ref_multiple
>> ref = "refs/tags/1.4.5"
=> "refs/tags/1.4.5"
>> split_ref_multiple(ref)
=> [
    [0] "refs",
    [1] "tags",
    [2] "1.4.5"
]
>> @ref
@ref_name  @ref_type
>> @ref_type
=> "tags"
>> @ref_name
=> "1.4.5"
>> @ref_type = nil
=> nil
>> @ref_name = nil
=> nil
>> split_ref_pop(ref)
=> "tags"
>> ref = "refs/tags"
=> "refs/tags"
>> split_ref_pop(ref)
=> "refs"
>> split_ref_multiple(ref)
=> [
    [0] "refs",
    [1] "tags"
]
>>
```

Is this desirable? Should we enforce any kind of constraint with `git check-ref-format` before we try to split it?

```bash
mckern@flexo Desktop $ git check-ref-format --normalize "ref/heads/master"
ref/heads/master
mckern@flexo Desktop $ echo $?
0
mckern@flexo Desktop $ git check-ref-format --normalize "master"
mckern@flexo Desktop $ echo $?
1
mckern@flexo Desktop $ 
```

mckern:
I could nitpick this some more, but I'm generally 👍 for the sake of getting it in and getting us using it. I've got some questions up, but consider them non-blocking.

lib/packaging/util/git_tags.rb
McdonaldSeanp: Yeah I think it's desirable. The problem with the parallel assignment is both "heads/master" and "refs/heads/master" are acceptable ref formats. So with the first option, the parallel assignment will incorrectly return ref_name as nil and ref_type as "master". what we could do if we didn't just make the function private (which I was thinking of doing) Is we could just force the return value as: [ref_name, ref_type] or something like that.

lib/packaging/util/git_tags.rb
McdonaldSeanp: SGTM

lib/packaging/util/git_tags.rb
McdonaldSeanp: So I did the above (added the explicit return of [ref_type, ref_name] ) so it would have a nice return value, I think that solves this problem

PR #548, opened 2016-05-16
(maint) update P-A ticket automation to mention vanagon:build_all

McdonaldSeanp:
We now have a vanagon:build_all rake task that humans should use when shipping
puppet-agent. the task is a wrapper around pl:jenkins:uber_build that adds a
tag check beforehand. update the ticket automation to reference this task.


underscorgan:
👍

PR #552, opened 2016-05-20
(QENG-3891) Added a polling message to print out at a specific frequency

ferglor:


pinkypie:
👍

waynr:
:+1: but we will probably need review/feedback from folks in RE. @ferglor, you should be able to get their attention if you pop into the "release-new" hipchat channel and request :eyes:

PR #553, opened 2016-05-24
(QE-3891) PuppetDB packaging job is timing out

ferglor:
Now syncing stdout so we get polling messages as and when they happen


pinkypie:
👍

jcoconnor:
I think you might need to say a little more than "hey happen..." in you commit message @ferglor

shrug:
Yup, 👍  pending some cleanup of the commit message, thanks! this is useful

PR #554, opened 2016-05-31
(maint) Add ticket to compile facter with java

melissa:
I keep forgetting to ensure facter is compiled for new platforms with
java. This isn't an issue until we try to add a server package for the
new platform, and everything fails because the server package can't
utilize the facter available via puppet-agent. Rather than try to
remember this, let's just be explicit and get facter.jar for all
platforms.

### DIFF
--- a/tasks/agent-tickets.rake
+++ b/tasks/agent-tickets.rake
@@ -67,6 +67,14 @@ This means you'll have to verify the build to resolve this ticket. If vanagon do
 you need to create a new ticket, block this ticket against that one, and add in vanagon support!
 DOC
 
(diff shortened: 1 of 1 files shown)


mckern:
Should the note specify that this is only really needed for Master platforms and that more exotic/esoteric/oddball platforms may not need it?

melissa:
I've included that information in the description of the ticket

mckern:
Derp. I read it, and it didn't register for me at all. That's on me, not on you.

McdonaldSeanp:
👍

PR #557, opened 2016-06-07
(CPR-368) Only run uber_ship tasks when packages exist

sarameisburger:
Currently in uber_ship we run tasks to update and deploy repos every time.
We only want to run these if the packages actually exist.


melissa:
rubocop failure: trailing whitespace

sarameisburger:
@melissa fixed!

sarameisburger:
Tested by individually deleting all the directories from puppet-agent to see if it would skip the tasks.

melissa:
👍

PR #560, opened 2016-06-09
Remove all the labs, also questions

sbeaulie:


sbeaulie:
Does it also support setting `MOCK` as an environment variable instead of passing it in the rake command line?
See https://github.com/sbeaulie/packaging/blame/master/README.md#L89

sbeaulie:
trivial change

stahnma:
So, I kind of think we should s/labs/inc since otherwise it's really difficult to figure out when we're talking about puppet, the software vs puppet the company.

shrug:
I think I agree with @stahnma, +1 to 'inc'

sbeaulie:
@stahnma and @shrug this PR was more about asking my question above than removing the labs. My understanding is that Github has this limitation where you cannot link a line number for a `.md` file, so I created a PR I could blame-link above. I read the docs, still unsure how I can use that MOCK variable, see my first comment above please ^^

stahnma:
Should we rebase this and update it or just close it out?

PR #561, opened 2016-06-09
(RE-7428) Bug fix of check to see if packages exist before deploying

sarameisburger:
Previous version did not recursively check all subdirectories for packages. Fixed this issue so multiple levels of subdirectories can be searched.


sarameisburger:
@ScottGarman I think I might have fixed the issue from: https://tickets.puppetlabs.com/browse/RE-7428
We will have to run some tests on it to find out!

underscorgan:
👍

PR #562, opened 2016-06-10
(RE-7411) Sign MSIs prior to s3/agent ship

melissa:
Prior to this commit, the s3/internal puppet-agent ship called this rake
task to generate signed repos to ship out. This task called automation
to sign both packages and repos prior to bundling them up and shipping
them out. We thought for some reason that the sign_all task was used to
sign the packages. This was an incorrect assumption. In order to sign
the MSI packages, we need to explicitly invoke the msi sign method on
the packages. This commit adds that logic to ensure we are actually
signing our packages.

Hopefully I'm not missing anything this time.


shrug:
👍

PR #564, opened 2016-06-14
(RE-6147) Internal Puppet Agent Shipping Creates Recursive Symlink

sarameisburger:
This commit adds a 'verbose" to one of the lines in order to debug the issue with the recursive symlink


mckern:
Trivial merge invoked

PR #566, opened 2016-06-21
(maint) update rake tasks to stop loading ticket tasks

McdonaldSeanp:
The ticket automation no longer lives under packaging, so stop loading those tasks


mckern:
👍 pending Jenkins

PR #567, opened 2016-06-23
(RE-6147) Fix recursive links in internal puppet-agent ship

sarameisburger:
When running the nightly_repos.rake file there would be a circular symlink and directories would be recursively made indefinitely. This fix takes out one of the calls to create a symlink (because it was called twice and should not have been) which fixes the issue. The symlink is set properly and the correct directories are made in the correct locations. 

(Absolutely do not merge this yet; it's being "tuned")


shrug:
Closing, replaced by https://github.com/puppetlabs/packaging/pull/568

PR #568, opened 2016-06-23
(RE-6147) Fix recursive links in internal puppet-agent ship

sarameisburger:
There was an issue when running the nightly_repos.rake file. A circular symlink was created and directories would be recursively made indefinitely. This fix takes out one of the calls to create a symlink (because it was called twice and should not have been) which fixes the issue. The symlink is set properly and the correct directories are made in the correct locations.

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -141,7 +141,7 @@ namespace :pl do
         end
 
         # Make a latest symlink for the project
-        FileUtils.ln_s(File.join("..", local_target, "repos"), File.join(Pkg::Config.project + "-latest", "repos"), :verbose => true)
+        FileUtils.ln_sf(File.join("..", local_target, "repos"), File.join(Pkg::Config.project + "-latest"), :verbose => true)
       end
     end
 
@@ -168,13 +168,12 @@ namespace :pl do
 
       # Then we find grab the appropriate meta-data only
       Pkg::Util::Execution.ex(%(find #{include_paths.map { |path| "pkg/#{Pkg::Config.project}-latest/#{path}" unless path.include? "repos" }.join(' ') } | sort > include_file_latest))
+
+      #include /repos in the include_file_latest so we correctly include the symlink in the final file list to ship
+      Pkg::Util::Execution.ex(%(echo "pkg/#{Pkg::Config.project}-latest/repos" >> include_file_latest))
       Pkg::Util::Execution.ex(%(tar -T include_file_latest -cf - | (cd ./tmp && tar -xf -)))
 
       Dir.chdir("tmp/pkg") do
-        # Link the latest repo that was trimmed down
-        local_target = Dir.glob(File.join(Pkg::Config.project, "/*/repos"))[0].split("/")[-2]
-        FileUtils.ln_s(File.join("..", Pkg::Config.project, local_target, "repos"), File.join(Pkg::Config.project + "-latest", "repos"))
-
         # Ship it to the target for consumption
         # First we ship the latest and clean up any repo-configs that are no longer valid with --delete-after
         Pkg::Util::Net.rsync_to("#{Pkg::Config.project}-latest", target_host, target_basedir, extra_flags: ["--delete-after", "--keep-dirlinks"])
@@ -244,7 +243,7 @@ namespace :pl do
       pa_source = File.join(remote_dir, Pkg::Config.project)
       pe_target = File.join(remote_dir, pe_version, Pkg::Config.project)
       local_pa = File.join(pa_source, version_string)
-      local_pe = File.join(pe_target, version_string)
+      local_pe = pe_target
       local_pa_latest = "#{pa_source}-latest"
       local_pe_latest = "#{pe_target}-latest"
 


tasks/nightly_repos.rake
```diff
@@ -140,12 +140,12 @@ namespace :pl do
           end
         end
 
-        # Make a latest symlink for the project
```
melissa: We should keep this comment

tasks/nightly_repos.rake
```diff
@@ -164,20 +164,20 @@ namespace :pl do
...
 
       Dir.chdir("tmp/pkg") do
-        # Link the latest repo that was trimmed down
-        local_target = Dir.glob(File.join(Pkg::Config.project, "/*/repos"))[0].split("/")[-2]
-        FileUtils.ln_s(File.join("..", Pkg::Config.project, local_target, "repos"), File.join(Pkg::Config.project + "-latest", "repos"))
 
         # Ship it to the target for consumption
         # First we ship the latest and clean up any repo-configs that are no longer valid with --delete-after
         Pkg::Util::Net.rsync_to("#{Pkg::Config.project}-latest", target_host, target_basedir, extra_flags: ["--delete-after", "--keep-dirlinks"])
+
```
melissa: We probably don't want these additional new lines

shermdog:
https://github.com/puppetlabs/packaging/commit/068c07af653ba090ff90df3b262c67fbaa427867  This was a thing.

melissa:
Hmm, I had forgotten that PDB had that issue. I'm wondering if it was the nightly ship job that was incorrectly configured?

stahnma:
@sarameisburger could we get more about why on that second commit message? I think it seems ok, but more context would help when we look back.

tasks/nightly_repos.rake
```diff
@@ -168,12 +168,19 @@ namespace :pl do
 
       # Then we find grab the appropriate meta-data only
       Pkg::Util::Execution.ex(%(find #{include_paths.map { |path| "pkg/#{Pkg::Config.project}-latest/#{path}" unless path.include? "repos" }.join(' ') } | sort > include_file_latest))
+
+      Pkg::Util::Execution.ex(%(echo "\npkg/#{Pkg::Config.project}-latest/repos" >> include_file_latest))
+
+      puts include_files_latest
```
melissa: This should be `puts File.read("include_file_latest")`. As is, it'll try to interpolate `include_files_latest` as a variable and it'll get confused since it's not a variable

tasks/nightly_repos.rake
```diff
@@ -168,12 +168,19 @@ namespace :pl do
...
       # Then we find grab the appropriate meta-data only
       Pkg::Util::Execution.ex(%(find #{include_paths.map { |path| "pkg/#{Pkg::Config.project}-latest/#{path}" unless path.include? "repos" }.join(' ') } | sort > include_file_latest))
+
+      Pkg::Util::Execution.ex(%(echo "\npkg/#{Pkg::Config.project}-latest/repos" >> include_file_latest))
+
+      puts File.read("include_file_latest")
+
       Pkg::Util::Execution.ex(%(tar -T include_file_latest -cf - | (cd ./tmp && tar -xf -)))
 
+      Pkg::Util::Execution.ex(%(echo "\npkg/tmp/#{Pkg::Config.project}-latest/repos" >> include_file_latest))
```
melissa: This is a duplicate of line 172

tasks/nightly_repos.rake
```diff
@@ -168,12 +168,19 @@ namespace :pl do
...
+
       Dir.chdir("tmp/pkg") do
-        # Link the latest repo that was trimmed down
-        local_target = Dir.glob(File.join(Pkg::Config.project, "/*/repos"))[0].split("/")[-2]
-        FileUtils.ln_s(File.join("..", Pkg::Config.project, local_target, "repos"), File.join(Pkg::Config.project + "-latest", "repos"))
+
+        #check if symlink is in directory
+        Pkg::Util::Execution.ex(%(ls -l #{Pkg::Config.project}/repos), true)
```
melissa: Oops, this should be `#{Pkg::Config.project}-latest/repos`

melissa: that might explain why we weren't seeing any output

tasks/nightly_repos.rake
sarameisburger: Yeah I thought maybe we could try manually adding repos into the tmp directory but I just realized the logic is wrong anyways, I'm gonna delete it

tasks/nightly_repos.rake
```diff
@@ -168,12 +168,18 @@ namespace :pl do
...
+
+      puts "PRINTING INCLUDE_FILE_LATEST"
+      puts File.read("include_file_latest")
+
       Pkg::Util::Execution.ex(%(tar -T include_file_latest -cf - | (cd ./tmp && tar -xf -)))
 
       Dir.chdir("tmp/pkg") do
-        # Link the latest repo that was trimmed down
-        local_target = Dir.glob(File.join(Pkg::Config.project, "/*/repos"))[0].split("/")[-2]
-        FileUtils.ln_s(File.join("..", Pkg::Config.project, local_target, "repos"), File.join(Pkg::Config.project + "-latest", "repos"))
```
melissa: @McdonaldSeanp figured this out. In the other link commands in this file, remove the target directory. So, in this case, remove `repos` after `Pkg::Config.project + "-latest"` and then change `FileUtils.ln_s` to `FileUtils.ln_sf` which will overwrite any existing links!

tasks/nightly_repos.rake
```diff
@@ -5,6 +5,8 @@ namespace :pl do
   # This is an INTERNAL rake task and should not be considered part of the packaging API.
   # Please do not depend on it.
   #
+  # NOTE: if trying to invoke the same link command multiple times (i.e., "ln -s ../puppet-agent/1.5.1/repos puppet-agent-latest/repos") you will get nested directories. To avoid the issue, if you need to invoke the nightly_repos.rake task again for some reason, remove the previous version directory before running the rake task again.
```
melissa: We don't need this comment anymore with the changes we just made!!

tasks/nightly_repos.rake
```diff
@@ -168,12 +170,15 @@ namespace :pl do
 
       # Then we find grab the appropriate meta-data only
       Pkg::Util::Execution.ex(%(find #{include_paths.map { |path| "pkg/#{Pkg::Config.project}-latest/#{path}" unless path.include? "repos" }.join(' ') } | sort > include_file_latest))
+
+      #include /repos in the include_file_latest so that we can set the symlink properly
```
melissa: Hmm, I'd word this more along the lines of "so we correctly include the symlink in the final file list to ship" or something

tasks/nightly_repos.rake
```diff
@@ -168,12 +170,15 @@ namespace :pl do
...
+      Pkg::Util::Execution.ex(%(echo "pkg/#{Pkg::Config.project}-latest/repos" >> include_file_latest))
       Pkg::Util::Execution.ex(%(tar -T include_file_latest -cf - | (cd ./tmp && tar -xf -)))
 
       Dir.chdir("tmp/pkg") do
-        # Link the latest repo that was trimmed down
-        local_target = Dir.glob(File.join(Pkg::Config.project, "/*/repos"))[0].split("/")[-2]
-        FileUtils.ln_s(File.join("..", Pkg::Config.project, local_target, "repos"), File.join(Pkg::Config.project + "-latest", "repos"))
+
+        #check if symlink is in directory
+        Pkg::Util::Execution.ex(%(ls -l #{Pkg::Config.project}-latest/repos), true)
```
melissa: We don't need this anymore! Yay!

melissa:
👍  it totally works!

tasks/nightly_repos.rake
```diff
@@ -244,7 +243,7 @@ namespace :pl do
       pa_source = File.join(remote_dir, Pkg::Config.project)
       pe_target = File.join(remote_dir, pe_version, Pkg::Config.project)
       local_pa = File.join(pa_source, version_string)
-      local_pe = File.join(pe_target, version_string)
+      local_pe = pe_target
```
underscorgan: I understand the other changes but this one confuses me.

melissa: This is the culprit behind the `1.5.1/1.5.1/1.5.1` directories. From what @sarameisburger and I found, local_pe is only used when creating the symlink. If we invoke `ln -sf puppet-agent/1.5.1 2012.2/puppet-agent/1.5.1` multiple times, then `ln` treats the final `1.5.1` element as a directory (since it already exists) and created a new symlink to 1.5.1 in that directory

sarameisburger: When we are setting the symlink we used to be giving it the target and the version, but we changed that because that's what was giving us the recursive version/version/version. Basically because if it includes the version in both parameters (local_pa and local_pe) when setting the symlink we will create extra directories that aren't needed. Similar to the change we made on line 144

melissa: JINX

underscorgan: OHH Cool!

PR #569, opened 2016-06-23
(maint) update solaris signing with hardcoded values

McdonaldSeanp:
I've had enough of failing ENV vars. This is going to hard code the locations
of the solaris signing certs as an interim fix. We will need to entirely
refactor where we set these environment vars anyway, so for now at least things
won't fail anymore.


shrug:
👍

PR #571, opened 2016-06-27
(RE-7431) Add overrides for build_data project branch

shrug:
Currently, data is loaded from the build-data repo for the project
and the content of the team variable. This works for the most part but
is a little inflexible.

This adds an environment variable ($DEV_BUILD) that will allow the user
to override the default project branch from #{project-name} to #{project-name}-dev.


underscorgan:
👍

PR #572, opened 2016-06-29
(RE-7406) Timestamps erroneously updated for tar, msi, swix, and dmg

underscorgan:
When we deploy tar, msi, swix, and dmg files from staging to production,
they end up with a newer timestamp on production than they do on staging
since we don't preserve timestamps (and preserving timestamps causes
permissions headaches so we shouldn't do that). For the other deploy
tasks we add the --update rsync flag so that we ignore files that are
newer in production than in staging. This commit adds that flag to the
tasks that ship tar, msi, swix, and dmg files.


shrug:
👍

PR #574, opened 2016-06-29
(RE-7584) Create complete set of student training VM links

sarameisburger:
This commit updates the symlinking so that the vbox and vmware files are linked to the student ova when the training VM's are shipped.


melissa:
cc/ @underscorgan is this correct? From the previous links that were there, this looks like what's supposed to happen

underscorgan:
👍 that looks right.

melissa:
👍  awesome!

PR #579, opened 2016-07-27
(RE-7445) Migrate pe repos to use aptly

underscorgan:
This replaces our use of reprepro and repsimple in the shipping of pe
repos with aptly.

### DIFF
--- a/tasks/pe_ship.rake
+++ b/tasks/pe_ship.rake
@@ -38,11 +38,11 @@ if Pkg::Config.build_pe
           target_path = %x(ssh -t #{Pkg::Config.apt_host} 'mktemp -d -t incoming-XXXXXX').chomp
         end
 
-        #   For reprepro, we ship just the debs into an incoming dir. On the remote end,
-        #   reprepro will pull these debs in and add them to the repositories based on the
+        #   For aptly, we ship just the debs into an incoming dir. On the remote end,
+        #   aptly will pull these debs in and add them to the repositories based on the
         #   dist, e.g. lucid, architecture notwithstanding.
         #
-        #   The layout that the reprepro library will expect is:
+        #   The layout that the aptly library will expect is:
         #
         #     incoming_dir/{$dists}/*.deb
         #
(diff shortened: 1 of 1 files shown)


underscorgan:
This needs to wait for a maintenance window on jenkins-compose so we can switch over to aptly repos!

stahnma:
This is happening in 15 minutes 🕐 ⌛

stahnma:
👍

tasks/pe_ship.rake
```diff
@@ -152,57 +152,30 @@ if Pkg::Config.build_pe
...
-            ", true)
-        else
-          stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "/usr/bin/repsimple add_all \
-              --confdir #{reprepro_confdir} \
-              --basedir #{reprepro_basedir} \
-              --databasedir #{reprepro_dbdir} \
-              --incomingdir #{incoming_dir} \
-              --verbose", true)
-        end
+        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "aptly repo add -remove-files #{Pkg::Config::pe_version}-#{dist} #{incoming_dir}/#{dist}
```
mckern: I'd consider using a HEREDOC instead of double quotes for this string, since it's multiline.

tasks/pe_ship.rake
```diff
@@ -152,57 +152,30 @@ if Pkg::Config.build_pe
...
+          if [ -d /opt/tools/aptly/public/#{Pkg::Config::pe_version}/dists/#{dist} ]; then
+            aptly publish update -gpg-key=\"8BBEB79B\" #{dist} #{Pkg::Config::pe_version}
```
mckern: Should the key ID be pulled out into a variable so it's easier to swap out?

tasks/pe_ship.rake
```diff
@@ -152,57 +152,30 @@ if Pkg::Config.build_pe
...
+          else
+            aptly publish repo -gpg-key=\"8BBEB79B\" #{Pkg::Config::pe_version}-#{dist} #{Pkg::Config::pe_version}
```
mckern: If we build this string in a HEREDOC instead, we can lose all of the escapes around double quotes inside this string.

PR #581, opened 2016-08-03
Revert "Revert "(RE-7445) Migrate pe repos to use aptly""

underscorgan:
Reverts puppetlabs/packaging#580


underscorgan:
We're going to try maintenance again sometime next week after all the shipping is done. We've figured out what went wrong, should go more smoothly next time.

PR #582, opened 2016-08-05
(maint) Move reprepro repo creation to aptly

melissa:
This commit changes how we create one-off repos for development use.
Prior to this, we had been using reprepro to create and then later to
sign repos. These repos contain development builds on
builds.delivery.puppetlabs.net, nightlies.puppetlabs.com,
agent-downloads.delivery.puppetlabs.net, and s3.

Some of the repos, such as those on s3, do contain final builds made
available to our enterprise customers.

This commit switches out reprepro for aptly. Aptly handles repos
slightly differently, which adds a little complexity to this work. We
have to explicitly create the repo, add packages to it, and publish the
repo. In order to maintain the same directory structure that was
generated with reprepro, we must then link the dists and pool
directories into our working directory. This is done with a relative
link to enable these repos to be tar'd up and delivered to different
locations (i.e., s3). I do hope the repos are handled as a unit rather
than individually (i.e., the entire apt directory is moved rather than
the subdirectories contained within the apt directory).

These created repos are initially not signed. Signing happens later in
the process. This is done by updating the published repo with the needed
gpg key. This commit adds in a provision that checks to make sure the
repo we are signing is an aptly repo. If it isn't, it will default back
to the reprepro way of signing a repo. This should not be needed for
long, but it a holdover in case we do want to ship repos that had been
created prior to this change going in.

### DIFF
--- a/lib/packaging/deb/repo.rb
+++ b/lib/packaging/deb/repo.rb
@@ -89,21 +89,49 @@ module Pkg::Deb::Repo
 
       # Make the conf directory and write out our configuration file
       cmd << "rm -rf apt && mkdir -p apt ; pushd apt ; "
-      cmd << %Q(for dist in $dists ; do mkdir -p $dist/conf ; pushd $dist ;
-      echo "
-Origin: Puppet Labs
-Label: Puppet Labs
-Codename: $dist
-Architectures: i386 amd64 arm64 armel armhf powerpc sparc mips mipsel
-Components: #{subrepo}
-Description: Apt repository for acceptance testing" >> conf/distributions ; )
 
-      # Create the repositories using reprepro. Since these are for acceptance
+      # Write out aptly configuration file
+      # We need a unique config for each project sha/ref we ship so that we can
+      # deliver these directories bundled up and the relative links will still work
+      architectures = %w(i386 amd64 arm64 armel armhf powerpc sparc mips mipsel)
+      description = "Apt repository for acceptance testing"
+      aptly_config_path = "#{artifact_directory}/repos/apt"
+      aptly_config_file = "#{aptly_config_path}/.aptly.conf"
+      aptly_flags = "-config='#{aptly_config_file}' -component='#{subrepo}' -distribution=$dist"
+      aptly_config_contents = {
+        :rootDir => File.join(aptly_config_path, '.aptly'),
+        :architectures => architectures
+      }
+      cmd << %Q(echo '#{aptly_config_contents.to_json}' > #{aptly_config_file} ; )
+
+      cmd << %Q(for dist in $dists ; do mkdir -p $dist ; pushd $dist ; )
+
+      # Create the repositories using aptly. Since these are for acceptance
       # testing only, we'll just add the debs and ignore source files for now.
-      #
-      cmd << "reprepro=$(which reprepro) ; "
-      cmd << %Q($reprepro includedeb $dist ../../#{prefix}deb/$dist#{Pkg::Config.apt_repo_name ? "/#{subrepo}" : ""}/*.deb ; popd ; done ; )
-      cmd << "popd ; popd "
+      cmd << "aptly=$(which aptly) ; "
+      cmd << %Q([[ -z "$aptly" ]] && echo "Unable to find the aptly command. Unable to create a repo." && exit 1 ; )
+
+      # First we need to create the aptly repo to ship to
+      cmd << %Q($aptly repo create #{aptly_flags} -comment='#{description}' #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist ; )
+
+      # Next we add the package to the newly created repo
+      # the packages we want to add may or may not live in a subdirectory, hence
+      # the extra complexity here
+      file_glob = "*.deb"
+      file_glob = File.join(subrepo, file_glob) if Pkg::Config.apt_repo_name
+      deb_packages = File.join('..', '..', prefix, 'deb', '$dist', file_glob)
+      cmd << %Q($aptly repo add -config="#{aptly_config_file}" #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist #{deb_packages} ; )
+
+      # Now we have to publish the repo to make it available
+      cmd << %Q($aptly publish repo #{aptly_flags} --skip-signing #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist; )
+
+      # Aptly publishes repos under a public directory in the .aptly dir. We
+      # need to add symlinks from this directory in order to maintain the
+      # currently expected structure. Otherwise, this would break a lot of
+      # code currently in use.
+      cmd << %Q(ln -s ../.aptly/public/#{Pkg::Config.project}-#{Pkg::Config.ref}-$dist/dists ; )
+      cmd << %Q(ln -s ../.aptly/public/#{Pkg::Config.project}-#{Pkg::Config.ref}-$dist/pool ; )
+      cmd << "popd ; done ; popd ; popd "
 
       return cmd
     end
@@ -146,25 +174,37 @@ Description: Apt repository for acceptance testing" >> conf/distributions ; )
 
     def sign_repos(target = "repos", message = "Signed apt repository")
       subrepo = Pkg::Config.apt_repo_name || 'main'
-      reprepro = Pkg::Util::Tool.check_tool('reprepro')
       Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
 
       dists = Pkg::Util::File.directories("#{target}/apt")
 
+      if File.exists?("../.aptly.conf")
+        aptly = Pkg::Util::Tool.check_tool('aptly')
+      else
+        reprepro = Pkg::Util::Tool.check_tool('reprepro')
+      end
+
       if dists
         dists.each do |dist|
           Dir.chdir("#{target}/apt/#{dist}") do
-            File.open("conf/distributions", "w") do |f|
-              f.puts "Origin: Puppet Labs
+            if aptly
+              Pkg::Util::Execution.ex(%Q(#{aptly} -config='../.aptly.conf' publish update -gpg-key="#{Pkg::Config.gpg_key}" #{dist} "#{Pkg::Config.project}-#{Pkg::Config.ref}-#{dist}"))
+            elsif reprepro
+              # This block can be removed once we are sure there are no more
+              # reprepro based repos that need to be signed.
+              File.open("conf/distributions", "w") do |f|
+                f.puts "Origin: Puppet Labs
 Label: Puppet Labs
 Codename: #{dist}
 Architectures: i386 amd64 arm64 armel armhf powerpc sparc mips mipsel
 Components: #{subrepo}
 Description: #{message} for #{dist}
 SignWith: #{Pkg::Config.gpg_key}"
+              end
+              Pkg::Util::Execution.ex("#{reprepro} -vvv --confdir ./conf --dbdir ./db --basedir ./ export")
+            else
+              fail "Neither aptly nor reprepro found. Cannot sign repos"
             end
-
-            Pkg::Util::Execution.ex("#{reprepro} -vvv --confdir ./conf --dbdir ./db --basedir ./ export")
           end
         end
       else
--- a/spec/lib/packaging/deb/repo_spec.rb
+++ b/spec/lib/packaging/deb/repo_spec.rb
@@ -76,7 +76,7 @@ describe "Pkg::Deb::Repo" do
 
     it "returns a command to make repos" do
       command = Pkg::Deb::Repo.repo_creation_command(prefix, artifact_directory)
-      command.should match(/reprepro/)
+      command.should match(/aptly/)
       command.should match(/#{prefix}/)
       command.should match(/#{artifact_directory}/)
     end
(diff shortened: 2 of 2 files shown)


lib/packaging/deb/repo.rb
```diff
@@ -98,12 +98,16 @@ def repo_creation_command(prefix, artifact_directory)
...
 Description: Apt repository for acceptance testing" >> conf/distributions ; )
 
-      # Create the repositories using reprepro. Since these are for acceptance
+      # Create the repositories using aptly. Since these are for acceptance
       # testing only, we'll just add the debs and ignore source files for now.
       #
-      cmd << "reprepro=$(which reprepro) ; "
-      cmd << %Q($reprepro includedeb $dist ../../#{prefix}deb/$dist#{Pkg::Config.apt_repo_name ? "/#{subrepo}" : ""}/*.deb ; popd ; done ; )
-      cmd << "popd ; popd "
+      cmd << "aptly=$(which aptly) ; "
```
underscorgan: should we add something in here so we don't end up failing silently if aptly isn't found? Or maybe something more substantial around this block as a whole, looks like a lot of opportunity for silent failures.

shrug: Maybe https://github.com/puppetlabs/packaging/blob/master/lib/packaging/util/tool.rb#L12-L30?

melissa: Hmm, the issue is that this command is run over ssh on a different host... so, find_tool isn't going to work for us. I could add a check in to ensure that variable is populated though

melissa: Alright, what do you think of line 105?

shrug: oh, yeah, I'm dumb.

I think that works.

lib/packaging/deb/repo.rb
```diff
@@ -98,12 +98,17 @@ def repo_creation_command(prefix, artifact_directory)
...
+      cmd << %Q([[ -z "$aptly" ]] && echo "Unable to find the aptly command. Unable to create a repo." && exit 1 ; )
+      cmd << %Q($aptly repo create -component=#{subrepo} -distribution=$dist -comment='Apt repository for acceptance testing' #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist ; )
+      cmd << %Q($aptly repo add #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist ../../#{prefix}deb/$dist#{Pkg::Config.apt_repo_name ? "/#{subrepo}" : ""}/*.deb ; )
+      cmd << %Q($aptly publish repo -component=#{subrepo} -distribution=$dist -comment='Apt repository for acceptance testing' -gpg-key=#{Pkg::Config.gpg_key} #{Pkg::Config.project}-#{Pkg::Config.ref} ; )
+      cmd << %Q(ln -s /opt/tools/aptly/public/#{Pkg::Config.project}-#{Pkg::Config.ref}/dists ; )
```
underscorgan: you're missing targets for the link

lib/packaging/deb/repo.rb
```diff
@@ -98,12 +98,17 @@ def repo_creation_command(prefix, artifact_directory)
...
+      cmd << %Q(ln -s /opt/tools/aptly/public/#{Pkg::Config.project}-#{Pkg::Config.ref}/pool ; )
```
underscorgan: this one too

lib/packaging/deb/repo.rb
underscorgan: that looks good

lib/packaging/deb/repo.rb
melissa: That's on purpose. The link creation is run in the destination directory, and it just creates a link with the same name as the source

lib/packaging/deb/repo.rb
melissa: That's on purpose. The link creation is run in the destination directory, and it just creates a link with the same name as the source

underscorgan: woops, yup, i always forget about that.

lib/packaging/deb/repo.rb
```diff
@@ -89,6 +89,9 @@ def repo_creation_command(prefix, artifact_directory)
 
       # Make the conf directory and write out our configuration file
       cmd << "rm -rf apt && mkdir -p apt ; pushd apt ; "
+
+      # Write out aptly configuration file
+      cmd << %Q(echo '{"rootDor": "#{artifact_directory}/repos/apt/.aptly"}' > .aptly.conf ; )
```
underscorgan: should be `rootDir`

underscorgan:
This looks good! have you tested yet?

lib/packaging/deb/repo.rb
melissa: I was wondering what I was doing wrong!

melissa:
I'm in the middle of testing. Issues with signing the published repos :( I'm considering not signing...

underscorgan:
how did the signing used to work? do you need to use a different key?

melissa:
It looks like we didn't sign at repo creation time originally. Instead we signed at a later date. I assume we were using the gpg agent. I'm not super sure though

underscorgan:
@melissa from looking at some job output it looks like maybe only nightlies get signed now, which probably runs on mozart. It looks like the normal repo jobs are not signed, so \o/

I'm only seeing https://github.com/puppetlabs/packaging/blob/master/lib/packaging/deb/repo.rb#L94-L99 in the console output from repo jobs.

spec/lib/packaging/deb/repo_spec.rb
```diff
@@ -119,24 +119,30 @@
...
-      Pkg::Util::Tool.should_receive(:find_tool).with('reprepro', {:required => true}).and_raise(RuntimeError)
+    it "fails without aptly" do
+      Pkg::Util::Tool.should_receive(:find_tool).with('aptly', {:required => true}).and_raise(RuntimeError)
       expect { Pkg::Deb::Repo.sign_repos }.to raise_error(RuntimeError)
     end
 
+    it "fails without aptly conf file" do
+      File.should_receive(:exists?).with("../.aptly.conf").and_return(FALSE)
+      expect { Pkg::Deb::Repo.sign_repos }.should_receive(:fail).with(".aptly.conf not found in apt directory. Failing.")
+    end
```
melissa: I can't get this test to pass. What am I doing wrong? (p.s., I'm terrible at rspec)

cc/ @underscorgan @mckern

lib/packaging/deb/repo.rb
```diff
@@ -89,6 +89,14 @@ def repo_creation_command(prefix, artifact_directory)
 
       # Make the conf directory and write out our configuration file
       cmd << "rm -rf apt && mkdir -p apt ; pushd apt ; "
+
+      # Write out aptly configuration file
+      # We need a unique config for each project sha/ref we ship so that we can
+      # deliver these directories bundled up and the relative links will still work
+      aptly_config_path = "#{artifact_directory}/repos/apt"
+      aptly_config = "#{aptly_config_path}/.aptly.conf"
+      cmd << %Q(echo '{"rootDir": "#{aptly_config_path}/.aptly"}' > #{aptly_config} ; )
```
mckern: Nitpick: we can safely omit the `Q` part of `%Q()` so it becomes `%()`. Rubocop prefers the latter form (I know that it was `%Q()` before this PR; this is a stray observation, not a blocker).

mckern: BUT instead of constructing bespoke JSON by hand like this, can we generate it?

```ruby
aptly_dir = { :rootDir => File.join(aptly_config_path, '.aptly') }
cmd << %(#{aptly_dir.to_json}  > #{aptly_config} ; )
```

Probably not a blocker; I'm thinking outloud. But I'm always weary/suspicious of raw, hand-rolled, bespoke JSON.

lib/packaging/deb/repo.rb
```diff
@@ -98,12 +106,30 @@ def repo_creation_command(prefix, artifact_directory)
...
 Description: Apt repository for acceptance testing" >> conf/distributions ; )
 
-      # Create the repositories using reprepro. Since these are for acceptance
+      # Create the repositories using aptly. Since these are for acceptance
       # testing only, we'll just add the debs and ignore source files for now.
-      #
-      cmd << "reprepro=$(which reprepro) ; "
-      cmd << %Q($reprepro includedeb $dist ../../#{prefix}deb/$dist#{Pkg::Config.apt_repo_name ? "/#{subrepo}" : ""}/*.deb ; popd ; done ; )
-      cmd << "popd ; popd "
+      aptly_flags = "-config='#{aptly_config}' -component='#{subrepo}' -distribution=$dist -architectures='amd64,arm64,armel,armhf,i386,mips,mipsel,powerpc,sparc'"
```
mckern: Can the values of `-architectures` be stored outside of the configuration flag definition here? 


```ruby
architectures = %w(amd64 arm64 armel armhf i386 mips mipsel powerpc sparc)
aptly_flags = %(-config='#{aptly_config}' -component='#{subrepo}' -distribution=$dist -architectures='#architectures.join(',')')
```

Alternatively, is there anywhere we can derive those values from already?

lib/packaging/deb/repo.rb
```diff
@@ -89,6 +89,15 @@ def repo_creation_command(prefix, artifact_directory)
...
       # Make the conf directory and write out our configuration file
       cmd << "rm -rf apt && mkdir -p apt ; pushd apt ; "
+
+      # Write out aptly configuration file
+      # We need a unique config for each project sha/ref we ship so that we can
+      # deliver these directories bundled up and the relative links will still work
+      aptly_config_path = "#{artifact_directory}/repos/apt"
+      aptly_config = "#{aptly_config_path}/.aptly.conf"
+      aptly_config_contents = { :rootDir => File.join(aptly_config_path, '.aptly') }
+      cmd << %Q(echo '#{aptly_json.to_json}' > #{aptly_config} ; )
```
mckern: Is it `aptly_config_contents` or `aptly_json`?

lib/packaging/deb/repo.rb
melissa: We can't derive them, as far as I can tell :(

lib/packaging/deb/repo.rb
```diff
@@ -89,21 +89,51 @@ def repo_creation_command(prefix, artifact_directory)
...
+
+      architectures = %w(i386 amd64 arm64 armel armhf powerpc sparc mips mipsel)
+      description = "Apt repository for acceptance testing"
+
       cmd << %Q(for dist in $dists ; do mkdir -p $dist/conf ; pushd $dist ;
       echo "
-Origin: Puppet Labs
-Label: Puppet Labs
+Origin: Puppet, Inc.
+Label: Puppet, Inc.
```
melissa: I probably shouldn't have made this change here, but I did. Does this file need to be consistent with the signing key? (as far as I can tell, it's just used for freight, not for aptly)

mckern: Given how these tools work (spoiler: poorly), I'd guess this is innocuous and safe.

lib/packaging/deb/repo.rb
```diff
@@ -89,21 +89,51 @@ def repo_creation_command(prefix, artifact_directory)
...
       # testing only, we'll just add the debs and ignore source files for now.
-      #
-      cmd << "reprepro=$(which reprepro) ; "
-      cmd << %Q($reprepro includedeb $dist ../../#{prefix}deb/$dist#{Pkg::Config.apt_repo_name ? "/#{subrepo}" : ""}/*.deb ; popd ; done ; )
-      cmd << "popd ; popd "
+      aptly_flags = "-config='#{aptly_config}' -component='#{subrepo}' -distribution=$dist -architectures='#{architectures.join(',')}'"
+
+      # Let's make sure aptly actually exists on this machine
+      cmd << "aptly=$(which aptly) ; "
+      cmd << %Q([[ -z "$aptly" ]] && echo "Unable to find the aptly command. Unable to create a repo." && exit 1 ; )
```
mckern: You could actually test for aptly directly here:

```bash
$ if which aptly 2>1 > /dev/null; then
> echo "Aptly found"
> else
> "You have no Aptly"
> fi
Aptly found
```

So this would look like:

```ruby
cmd << %(which aptly 2>1 > /dev/null && echo "Unable to find the aptly command. Unable to create a repo." && exit 1 ; )
```

lib/packaging/deb/repo.rb
melissa: Good point. Fixed!

lib/packaging/deb/repo.rb
```diff
@@ -89,21 +89,51 @@ def repo_creation_command(prefix, artifact_directory)
...
-Origin: Puppet Labs
-Label: Puppet Labs
+Origin: Puppet, Inc.
+Label: Puppet, Inc.
 Codename: $dist
-Architectures: i386 amd64 arm64 armel armhf powerpc sparc mips mipsel
+Architectures: #{architectures.join(' ')}
 Components: #{subrepo}
-Description: Apt repository for acceptance testing" >> conf/distributions ; )
+Description: #{description}" >> conf/distributions ; )
```
underscorgan: I believe this is a reprepro specific file, can probably get rid of this

lib/packaging/deb/repo.rb
```diff
@@ -89,21 +89,51 @@ def repo_creation_command(prefix, artifact_directory)
 
       # Make the conf directory and write out our configuration file
       cmd << "rm -rf apt && mkdir -p apt ; pushd apt ; "
+
+      # Write out aptly configuration file
+      # We need a unique config for each project sha/ref we ship so that we can
+      # deliver these directories bundled up and the relative links will still work
+      aptly_config_path = "#{artifact_directory}/repos/apt"
+      aptly_config = "#{aptly_config_path}/.aptly.conf"
+      aptly_config_contents = { :rootDir => File.join(aptly_config_path, '.aptly') }
```
underscorgan: should the arch stuff get added to the aptly config?

underscorgan:
commit message still references removing the signing task, but that got left in.

lib/packaging/deb/repo.rb
melissa: Ooh, that's a good thought!

lib/packaging/deb/repo.rb
```diff
@@ -89,21 +89,51 @@ def repo_creation_command(prefix, artifact_directory)
...
+
+      # Let's make sure aptly actually exists on this machine
+      cmd << "aptly=$(which aptly) ; "
+      cmd << %Q([[ -z "$aptly" ]] && echo "Unable to find the aptly command. Unable to create a repo." && exit 1 ; )
+
+      # First we need to create the aptly repo to ship to
+      cmd << %Q($aptly repo create #{aptly_flags} -comment='#{description}' #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist ; )
+
+      # Next we add the package to the newly created repo
+      cmd << %Q($aptly repo add -config="#{aptly_config}" #{Pkg::Config.project}-#{Pkg::Config.ref}-$dist ../../#{prefix}deb/$dist#{Pkg::Config.apt_repo_name ? "/#{subrepo}" : ""}/*.deb ; )
```
mckern: Can the ternary operator and interpolation (... and path construction) here be moved out of the string?

```ruby
file_glob = "*.deb"
file_glob = File.join(subrepo, file_glob) if Pkg::Config.apt_repo_name
useful_complexity = File.join('..', '..', prefix, 'deb', '$dist', file_glob)
cmd << %Q($aptly repo add -config="#{aptly_config}" #{useful_complexity} ; )
```

(This is kind of an example, not necessarily useful as written; turns out this is a surprisingly complex path!)

melissa:
@underscorgan @mckern updated!

melissa:
@underscorgan do you have any thoughts on testing nightlies/agent-downloads/s3 with this code? I'm concerned what's going to happen if we tar up and ship the aptly repos to a system without aptly on it

melissa:
https://github.com/puppetlabs/puppetdb/pull/2040 should be merged prior to this

melissa:
http://jenkins-compose.delivery.puppetlabs.net/job/internal_puppet-agent_ship/479/console was the test I wanted to do. Unfortunately, it failed because it was trying to sign using reprepro, and failed because it was expecting reprepro specific information that is no longer present with aptly. I'm not sure a good way to test this outside of merging this PR :( PLEASE NOTE this should not be merged until the above PDB pr is merged, otherwise their s3 repos will be incomplete.

melissa:
Blocked on above PDB PR. Once that's merged, we just have to pick a good time and get this in.

melissa:
Clear to merge. Let's do it in the morning, early in the week.

shrug:
maybe after the next agent release?

shrug:
@melissa merge this tomorrow (Wednesday) morning?

underscorgan:
👍

PR #583, opened 2016-08-12
(PA-467) Add windows versionless symlinks to ship jobs

McdonaldSeanp:
This is functionality to add the project-arch.msi symlinks that have no
version. The code itself looks for the symlink (if it's there already) and if
the source package exists before symlinking.

This should not break any legacy builds of puppet-agent as the lookup for the
existing file will prevent overwriting anything

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -449,6 +449,21 @@ namespace :pl do
         mv(packaging_bundle, local_dir)
       end
 
+      # This is functionality to add the project-arch.msi symlinks that have no
+      # version. The code itself looks for the symlink (if it's there already)
+      # and if the source package exists before symlinking. Searching for the
+      # packages has been restricted specifically to just the pkg/windows dir
+      # on purpose, as this is where we currently have all windows packages
+      # building to. Once we move the Metadata about the output location in
+      # to one source of truth we can refactor this to use that to search
+      #                                           -Sean P. M. 08/12/16
+      packages = Dir["#{local_dir}/windows/*"]
+      ["x86", "x64"].each do |arch|
+        if !packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi")) && packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"))
+          File.symlink(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"), File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi"))
+        end
+      end
+
       Pkg::Util::Execution.retry_on_fail(:times => 3) do
         Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.distribution_server, "mkdir --mode=775 -p #{project_basedir}")
         Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.distribution_server, "mkdir -p #{artifact_dir}")


tasks/ship.rake
```diff
@@ -449,6 +449,21 @@ namespace :pl do
...
+      # version. The code itself looks for the symlink (if it's there already)
+      # and if the source package exists before symlinking. Searching for the
+      # packages has been restricted specifically to just the pkg/windows dir
+      # on purpose, as this is where we currently have all windows packages
+      # building to. Once we move the Metadata about the output location in
+      # to one source of truth we can refactor this to use that to search
+      #                                           -Sean P. M. 08/12/16
+      packages = Dir["#{artifact_dir}/windows/*"]
+      ["x86", "x64"].each do |arch|
+        unless packages.include?("#{Pkg::Config.project}-#{arch}.msi") || !package.include?("#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi")
```
mckern: Can this be inverted? The `unless` makes it seem like a logical exception whereas an `if` reads more like the desired state here. Also, pairing an `||` with `unless` sometimes makes people (not me, but people who are not me) cranky.

```ruby
["x86", "x64"].each do |arch|
  if package.include?("#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi") || !packages.include?("#{Pkg::Config.project}-#{arch}.msi")
    File.symlink("#{artifact_dir}/#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi", "#{artifact_dir}/#{Pkg::Config.project}-#{arch}.msi")
  end
end
```

McdonaldSeanp: Sure.

tasks/ship.rake
```diff
@@ -449,6 +449,21 @@ namespace :pl do
...
+      # version. The code itself looks for the symlink (if it's there already)
+      # and if the source package exists before symlinking. Searching for the
+      # packages has been restricted specifically to just the pkg/windows dir
+      # on purpose, as this is where we currently have all windows packages
+      # building to. Once we move the Metadata about the output location in
+      # to one source of truth we can refactor this to use that to search
+      #                                           -Sean P. M. 08/12/16
+      packages = Dir["#{artifact_dir}/windows/*"]
+      ["x86", "x64"].each do |arch|
+        if !packages.include?("#{Pkg::Config.project}-#{arch}.msi") && package.include?("#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi")
```
mckern: This got swapped to a `&&` when it changed from an `unless` to `if` -- is it safe to move the negated conditional to the back of the `if` or are we now relying on the short-circuit nature of `&&`?

tasks/ship.rake
```diff
@@ -449,6 +449,21 @@ namespace :pl do
...
+          File.symlink("#{artifact_dir}/#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi", "#{artifact_dir}/#{Pkg::Config.project}-#{arch}.msi")
+        end
+      end
```
mckern: I know we don't do it anywhere else in the Packaging repo, but if possible I'd like to see things like `"#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi}"` and `#{Pkg::Config.project}-#{arch}.msi` turned into intermediate variables before they're used. We construct a LOT of paths-as-strings in Packaging on the fly, and I think we tend to lose track of them over time so they end up subtly wrong any time we change things.
Probably not appropriate for this PR but I like to think out loud.

tasks/ship.rake
```diff
@@ -449,6 +449,21 @@ namespace :pl do
...
+      # and if the source package exists before symlinking. Searching for the
+      # packages has been restricted specifically to just the pkg/windows dir
+      # on purpose, as this is where we currently have all windows packages
+      # building to. Once we move the Metadata about the output location in
+      # to one source of truth we can refactor this to use that to search
+      #                                           -Sean P. M. 08/12/16
+      packages = Dir["#{artifact_dir}/windows/*"]
+      ["x86", "x64"].each do |arch|
+        if !packages.include?("#{Pkg::Config.project}-#{arch}.msi") && package.include?("#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi")
+          File.symlink("#{artifact_dir}/#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi", "#{artifact_dir}/#{Pkg::Config.project}-#{arch}.msi")
```
mckern: Nitpick: `File.join(artifact_dir, "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi")` instead of `"#{artifact_dir}/#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"` (also for `"#{artifact_dir}/#{Pkg::Config.project}-#{arch}.msi"`).

Non-blocking

mckern:
I'm nitpicking this to death, but none of them are really worth blocking. I'm 👍 on what looks like it'll do the Right Thing. @McdonaldSeanp have you tested this and do you have any test output from it?

tasks/ship.rake
McdonaldSeanp: unless (condition || condition) != if (condition || condition)

I changed it to && to retain the same behavior

McdonaldSeanp: The way this works should retain the behavior of the original unless. the || must be changed to && to retain the original behavior.

ScottGarman:
:+1:  if we get confirmation that some testing has been done on this.

McdonaldSeanp:
I've tested a version of this with the rest of this rake task commented out, and it creates the symlinks successfully.

tasks/ship.rake
```diff
@@ -449,6 +449,21 @@ namespace :pl do
...
+      # on purpose, as this is where we currently have all windows packages
+      # building to. Once we move the Metadata about the output location in
+      # to one source of truth we can refactor this to use that to search
+      #                                           -Sean P. M. 08/12/16
+      packages = Dir["#{local_dir}/windows/*"]
+      ["x86", "x64"].each do |arch|
+        if !packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi")) && packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"))
+          File.symlink(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"), File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi"))
+        end
+      end
```
mckern: It's a little less How-We've-Always-Done-It, but consider the readability of something like this vs. a lot of inline string construction:

```ruby
packages = Dir["#{local_dir}/windows/*"]
["x86", "x64"].each do |arch|
  this_file = File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi")
  that_file = File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi")

  if !packages.include?(this_file) && packages.include?(that_file)
    File.symlink(that_file, this_file)
  end
end
```

mckern: (I don't recommend `this_file` and `that_file` in seriousness; I just used them because they were fun to read outloud. I'd probably suggest `versioned_msi` and `unversioned_msi` for clarity of intent.

mckern: Also, 👍 on the comment explaining what this is for. That's my favorite part so far.

McdonaldSeanp: Yeah, that is much better, I'll open a maint to make this look better

PR #584, opened 2016-08-15
(maint) update windows symlink logic to use correct version

McdonaldSeanp:
The logic uses inline string construction alot, this PR updates all that string
construction to instead use variables.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -459,8 +459,11 @@ namespace :pl do
       #                                           -Sean P. M. 08/12/16
       packages = Dir["#{local_dir}/windows/*"]
       ["x86", "x64"].each do |arch|
-        if !packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi")) && packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"))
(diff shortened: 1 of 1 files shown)


tasks/ship.rake
```diff
@@ -459,8 +459,11 @@ namespace :pl do
       #                                           -Sean P. M. 08/12/16
       packages = Dir["#{local_dir}/windows/*"]
       ["x86", "x64"].each do |arch|
-        if !packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi")) && packages.include?(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"))
-          File.symlink(File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi"), File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi"))
+        package_filename = File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{Pkg::Config.version}-#{arch}.msi")
+        link_filename = File.join("#{local_dir}", "windows", "#{Pkg::Config.project}-#{arch}.msi")
```
mckern: Use `local_dir` instead of "#{local_dir}" in these `File.join`'s, or if you think we absolutely have to cast it to a String, use `local_dir.to_s`

mckern:
1 nitpick, 👍 otherwise

mckern:
@McdonaldSeanp commit typo:


> Vanagon outputs **pacakges** with versions stripped right from git describe, so we
need to do the same in packaging when looking for packages to match.

tvpartytonight:
👍

PR #585, opened 2016-08-18
(maint) Windows symlinks were not getting created correctly

underscorgan:
Since the source MSI lives in the same directory as the link target, the
source should just be the file name not a more qualified file path.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -460,11 +460,12 @@ namespace :pl do
       packages = Dir["#{local_dir}/windows/*"]
       ["x86", "x64"].each do |arch|
         package_version = Pkg::Util::Version.git_describe.tr('-', '.')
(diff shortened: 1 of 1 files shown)


ScottGarman:
:+1:

Iristyle:
Is this a follow up on [PA-467](https://tickets.puppetlabs.com/browse/PA-467)?

McdonaldSeanp:
This will not work, as the search that we do will have the more qualified path in it. Note that ://github.com/underscorgan/packaging/blob/4afc4c0be04ceb77b2cea524cace0850962528f4/tasks/ship.rake#L466 needs to have the more qualified path, while the symlink just needs the name

McdonaldSeanp:
My suggestion would be to create a third variable with just the package name (including version) in it and update https://github.com/underscorgan/packaging/blob/8f14ea75263ce4378fdca63c9ad2b2066ffc48f4/tasks/ship.rake#L467 to use the thrid variable

underscorgan:
Yup, about to push that up.

PR #586, opened 2016-08-18
(PA-467) More linking work

underscorgan:
So, there's a lot of history here. We set packages to be immutable as
part of the shipping so they don't get overwritten, which doesn't play
well with symlinks. So we need to use FileUtils.ln(). Also,
File.symlink() and FileUtils.ln() behave differently with respect to
whether or not they expect the path for the source to be relative to pwd
or relative to the link target. By using hard links we should be able to
set them to be immutable without breaking everything.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -449,9 +449,9 @@ namespace :pl do
         mv(packaging_bundle, local_dir)
       end
 
-      # This is functionality to add the project-arch.msi symlinks that have no
-      # version. The code itself looks for the symlink (if it's there already)
-      # and if the source package exists before symlinking. Searching for the
+      # This is functionality to add the project-arch.msi links that have no
+      # version. The code itself looks for the link (if it's there already)
+      # and if the source package exists before linking. Searching for the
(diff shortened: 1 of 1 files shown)


tasks/ship.rake
```diff
@@ -460,12 +460,17 @@ namespace :pl do
...
       ["x86", "x64"].each do |arch|
         package_version = Pkg::Util::Version.git_describe.tr('-', '.')
-        package_name = "#{Pkg::Config.project}-#{package_version}-#{arch}.msi"
-        package_filename = File.join(local_dir, "windows", "#{package_name}")
+        package_filename = File.join(local_dir, "windows", "#{Pkg::Config.project}-#{package_version}-#{arch}.msi")
         link_filename = File.join(local_dir, "windows", "#{Pkg::Config.project}-#{arch}.msi")
 
         if !packages.include?(link_filename) && packages.include?(package_filename)
-          File.symlink(package_name, link_filename)
+          # Dear future code spelunkers:
```
stahnma: 😆

stahnma:
👍

PR #587, opened 2016-08-19
(maint) Update to use a new key

underscorgan:
The exiting key is expiring soon, so update the readme to use the new
one.


McdonaldSeanp:
:+1:

PR #589, opened 2016-09-12
(maint) Update linking to be smarter

melissa:
This commit adds in logic to check to see if the symlink to be
replaced is broken. If it is, then it removes the symlink so it can
create a new one. This will help reduce manual intervention when the
symlinks get all strange, because computers.

### DIFF
--- a/tasks/education.rake
+++ b/tasks/education.rake
@@ -38,6 +38,10 @@ namespace :pl do
 
       # Update symlink to point to the VM we just shipped
       link_target.each do |link|
+        link_path = File.join(target_directory, link)
+        link_md5_path = "#{link_path}.md5"
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
         Pkg::Util::Net.remote_ssh_cmd(target_host, "cd #{target_directory} ; ln -sf #{File.basename(vm)} #{link}")
         Pkg::Util::Net.remote_ssh_cmd(target_host, "cd #{target_directory} ; ln -sf #{File.basename(md5)} #{link}.md5")
       end


tasks/education.rake
```diff
@@ -36,8 +36,14 @@ namespace :pl do
       Pkg::Util::Net.rsync_to(vm, target_host, target_directory)
       Pkg::Util::Net.rsync_to(md5, target_host, target_directory)
 
+      # Update sticky bit on newly shipped files
```
underscorgan: nitpick: this isn't the sticky bit, this is setting the file to be immutable

melissa: awww, for some reason I had it in my head those were interchangeable terms. Thanks for correcting me!

tasks/education.rake
```diff
@@ -38,6 +38,8 @@ namespace :pl do
 
       # Update symlink to point to the VM we just shipped
       link_target.each do |link|
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "[[ ! -e '#{File.join(target_directory, link)}' ]] && echo '#{File.join(target_directory, link)} is a broken link, deleting' && unlink '#{File.join(target_directory, link)}'")
```
mckern: There's more than one `&&` in here; an `if; then` would be preferable. That can still be constructed on one line, or defined in an intermediate variable and passed in.

melissa: Updated

tasks/education.rake
```diff
@@ -38,6 +38,10 @@ namespace :pl do
 
       # Update symlink to point to the VM we just shipped
       link_target.each do |link|
+        link_path = File.join(target_directory, link)
+        link_md5_path = "#{File.join(target_directory, link)}.md5"
```
mckern: This would be identical to `"#{link_path}.md5"`, right?

melissa: Updated

tasks/education.rake
```diff
@@ -38,6 +38,10 @@ namespace :pl do
 
       # Update symlink to point to the VM we just shipped
       link_target.each do |link|
+        link_path = File.join(target_directory, link)
+        link_md5_path = "#{link_path}.md5"
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
```
mckern: I am excited to extract this out into a method soonish.

mckern:
I am 👍 on this because I ❤️ sanity checks

tasks/education.rake
melissa: yeah, that sounds like a smarter thing to do...

mckern: We can get smart later if we need "working" now. Smart takes time.

PR #590, opened 2016-09-12
(maint) Unlink only if it's actually a broken link

melissa:
Previously, this could would attempt to unlink if the link didn't even
exist or if it was truly a broken symlink. This commit updates the check
to be sure there is actually a link that points to a non-existent file.
This means if we're creating a brand new set of links, then our
automation won't bomb out on us.

### DIFF
--- a/tasks/education.rake
+++ b/tasks/education.rake
@@ -40,8 +40,8 @@ namespace :pl do
       link_target.each do |link|
         link_path = File.join(target_directory, link)
         link_md5_path = "#{link_path}.md5"
-        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
-        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ -L '#{link_path}' ]] && [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ -L '#{link_md5_path}' ]] && [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
         Pkg::Util::Net.remote_ssh_cmd(target_host, "cd #{target_directory} ; ln -sf #{File.basename(vm)} #{link}")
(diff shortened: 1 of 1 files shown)


tasks/education.rake
```diff
@@ -40,8 +40,8 @@ namespace :pl do
       link_target.each do |link|
         link_path = File.join(target_directory, link)
         link_md5_path = "#{link_path}.md5"
-        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
-        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ -h '#{link_path}' ]] && [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ -h '#{link_md5_path}' ]] && [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
```
mckern: From `man test`:

```
-h file       True if file exists and is a symbolic link.  This operator is retained for compatibility with previous
                 versions of this program.  Do not rely on its existence; use -L instead.
```

melissa: Updated

tasks/education.rake
```diff
@@ -40,8 +40,8 @@ namespace :pl do
       link_target.each do |link|
         link_path = File.join(target_directory, link)
         link_md5_path = "#{link_path}.md5"
-        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
-        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ -L '#{link_path}' ]] && [[ ! -e '#{link_path}' ]] ; then echo '#{link_path} is a broken link, deleting' ; unlink '#{link_path}' ; fi")
+        Pkg::Util::Net.remote_ssh_cmd(target_host, "if [[ -L '#{link_md5_path}' ]] && [[ ! -e '#{link_md5_path}' ]] ; then echo '#{link_md5_path} is a broken link, deleting' ; unlink '#{link_md5_path}' ; fi")
```
mckern: Offhand, I just realized that we're using `[[` everywhere, which means we're assuming that we're going to get a remote `bash` shell by default. This isn't the PR to fix that but we might want to start working out if that's a safe assumption or not.

mckern:
👍

PR #591, opened 2016-09-16
(maint) Update aptly check

melissa:
Previously the check to see if we were using reprepro or aptly was in
the wrong place. This was failing in such a way that all repos were
reprepro based. This commit fixes that mistake.


shrug:
@melissa Can you update the commit message with the ticket?

shrug:
👍

PR #592, opened 2016-09-19
(maint) Ensure aptly directories not hidden

melissa:
We need to access the aptly files via wget. However, wget doesn't
retrieve hidden files. This commit removes that quality of the aptly
config file and directory

### DIFF
--- a/lib/packaging/deb/repo.rb
+++ b/lib/packaging/deb/repo.rb
@@ -96,10 +96,10 @@ module Pkg::Deb::Repo
       architectures = %w(i386 amd64 arm64 armel armhf powerpc sparc mips mipsel)
       description = "Apt repository for acceptance testing"
       aptly_config_path = "#{artifact_directory}/repos/apt"
-      aptly_config_file = "#{aptly_config_path}/.aptly.conf"
+      aptly_config_file = "#{aptly_config_path}/aptly.conf"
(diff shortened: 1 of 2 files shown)


lib/packaging/deb/repo.rb
```diff
@@ -181,15 +181,17 @@ def sign_repos(target = "repos", message = "Signed apt repository")
...
-            else
+            if File.exists?("conf/distributions")
               reprepro = Pkg::Util::Tool.check_tool('reprepro')
+            else
+              aptly = Pkg::Util::Tool.check_tool('aptly')
             end
 
             if aptly
+              puts "Signing published aptly repo for #{dist}"
               Pkg::Util::Execution.ex(%Q(#{aptly} -config='../.aptly.conf' publish update -gpg-key="#{Pkg::Config.gpg_key}" #{dist} "#{Pkg::Config.project}-#{Pkg::Config.ref}-#{dist}"))
```
underscorgan: I think this will still explode if the .aptly.conf file isn't getting synced over

underscorgan:
switching the order made tests 💥

lib/packaging/deb/repo.rb
melissa: yeah, it's definitely still going to explode

PR #593, opened 2016-09-19
(maint) Change aptly rootdir to relative path

melissa:
We need this to work as an atomic unit. In order to do that, this path
needs to work on systems that aren't builds.delivery.puppetlabs.net. The
best way to do this is by referencing a relative path


melissa:
I'm open to changes for sure. Thoughts?

PR #596, opened 2016-10-18
(maint) Don't fail if nil

melissa:
We recently ran into an issue where one or the other of these values
return nil, which broke the automation. This should fix it.

### DIFF
--- a/tasks/pe_ship.rake
+++ b/tasks/pe_ship.rake
@@ -167,7 +167,7 @@ if Pkg::Config.build_pe
           fi
           ", true)
 
-        output = stdout + stderr
+        output = stdout.to_s + stderr.to_s
 
         if output.include?("ERROR:") || output.include?("There have been errors!")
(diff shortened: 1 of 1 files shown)


tasks/pe_ship.rake
```diff
@@ -167,7 +167,7 @@ if Pkg::Config.build_pe
           fi
           ", true)
 
-        output = stdout + stderr
+        output = (stdout.nil? ? '' : stdout) + (stderr.nil? ? '' : stderr)
```
mckern: I believe this is functionally identical:

```ruby
output = stdout.to_s + stderr.to_s
```

melissa: oh lovely!

underscorgan:
👍

jonathannewman:
Promotions are blocked by this.  
`irb(main):001:0> nil.to_s
=> ""
`

PR #595, opened 2016-09-22
(PA-552) OSX signing should not echo the command output

shrug:
This fixes a couple of issues with the remote_ssh_command method to prevent it from echoing output when we don't want it, and to actually return output when we want it to.

This also updates the OSX signing command to correctly exit on errors instead of proceeding and potentially silently failing while returning unsigned packages.

PR #597, opened 2016-10-19
Revert "(PA-552) OSX signing should not echo the command output"

underscorgan:
Reverts puppetlabs/packaging#595


underscorgan:
things are crashing and since we don't have output anymore it's really hard to debug.

mckern:
Will merge when Travis says OK

PR #599, opened 2016-11-12
(RE-8179) Add the ability to ubership quietly

underscorgan:
This PR includes a number of stability improvements and safety checks to make it more reliable to run pl:jenkins:uber_ship with FOSS_ONLY=true.

Still todo:

- [x]  Make ANSWER_OVERRIDE work more like the normal config parameters
- [x]  Clean up retrieve output so we can actually see the warnings
- [x]  PR to add foss_platforms to the release team build_data. -- https://github.com/puppetlabs/build-data/pull/77

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -3,7 +3,8 @@ require "packaging/platforms"
 # These are all of the parameters known to our packaging system.
 # They are ingested by the config class as class instance variables
 module Pkg::Params
-  BUILD_PARAMS = [:apt_host,
+  BUILD_PARAMS = [:answer_override,
+                  :apt_host,
                   :apt_releases,
                   :apt_repo_command,
                   :apt_repo_name,
@@ -174,6 +175,7 @@ module Pkg::Params
   #           Note: :type is assumed :string if not present
   #
   ENV_VARS = [
+              { :var => :answer_override,         :envvar => :ANSWER_OVERRIDE },
               { :var => :apt_host,                :envvar => :APT_HOST },
               { :var => :apt_releases,            :envvar => :APT_RELEASES,    :type => :array },
               { :var => :apt_repo_path,           :envvar => :APT_REPO },
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -196,7 +196,7 @@ module Pkg::Params
               { :var => :dmg_host,                :envvar => :DMG_HOST },
               { :var => :dmg_staging_server,      :envvar => :DMG_STAGING_SERVER },
               { :var => :final_mocks,             :envvar => :MOCK },
-              { :var => :foss_only,               :envvar => :FOSS_ONLY },
+              { :var => :foss_only,               :envvar => :FOSS_ONLY,       :type => :bool },
               { :var => :foss_platforms,          :envvar => :FOSS_PLATFORMS,  :type => :array },
               { :var => :gem_host,                :envvar => :GEM_HOST },
               { :var => :gpg_key,                 :envvar => :GPG_KEY },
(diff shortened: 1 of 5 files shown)


shrug:
👍

tasks/jenkins.rake
```diff
@@ -254,6 +254,7 @@ namespace :pl do
         jenkins:retrieve
         jenkins:sign_all
         uber_ship
+        ship_gem
```
melissa: isn't gem_ship included in the uber_ship task?

melissa: Wait, yeah, I don't really follow how this change adds consistency?

tasks/retrieve.rake
```diff
@@ -21,30 +21,33 @@ namespace :pl do
...
-              puts "Skipping fetch for #{platform}, #{url} has already been fetched"
-            else
-              urls[url] = true
+            begin
+              platform_path = Pkg::Util::Platform.artifacts_path(platform, package_url)
+              _, _, arch = Pkg::Util::Platform.parse_platform_tag(platform)
+              url = "#{package_url}/#{platform_path}"
               puts "Fetching: Platform = #{platform}, URL = #{url}"
-              sh "#{wget} -r -np -nH -l 0 --cut-dirs 3 -P #{local_target} --reject 'index*' #{url}/"
+              sh "#{wget} --quiet -r -np -nH -l 0 --cut-dirs 3 -P #{local_target} --reject 'index*' --accept '*#{arch}*' #{url}/"
```
melissa: should we be using wget or rsync here?

tasks/jenkins.rake
underscorgan: I removed it from the uber_ship task. It felt more consistent because if you're running uber_ship with the prompts that was happening during the uber_ship rather than after that with all the other update/deploy X tasks.

lib/packaging/config/params.rb
```diff
@@ -174,6 +175,7 @@ module Pkg::Params
   #           Note: :type is assumed :string if not present
   #
   ENV_VARS = [
+              { :var => :answer_override,         :envvar => :ANSWER_OVERRIDE },
```
underscorgan: I didn't add a type because this can be a bool or a string, and it's coerced into a bool when we actually use it, but want to be able to pass true, false, y, n, yes, no

tasks/retrieve.rake
underscorgan: I discussed this with @shrug yesterday, and while not using wget is probably better, that change is out of the scope of this PR

tasks/retrieve.rake
```diff
@@ -21,30 +21,33 @@ namespace :pl do
...
+            rescue => e
+              warn "Encountered error fetching #{platform}:"
+              warn e
```
underscorgan: Yes, as opposed to right now when it would fail.

lib/packaging/config/params.rb
melissa: Good to know

tasks/retrieve.rake
melissa: fair

tasks/retrieve.rake
melissa: hooray!

tasks/jenkins.rake
melissa: Hmm. This change makes sense from that perspective, but it still throws me off. @mckern you've spent a lot of time with this code. Do you have an opinion?

underscorgan: Yeah, this path always confused me. I felt like gems should've been their own thing but they were part of the uber_ship. I might be thinking about this wrong though.

stahnma:
👍  merge in the morning so we can watch the fallout.

PR #600, opened 2016-11-16
(maint) cleanup config printing

underscorgan:
This modifies config:print to call pl:fetch first so that you see the actual variables that will be loaded during the ship, and also adds a config:print_hosts to allow you to check where things will be shipping.

### DIFF
--- a/lib/packaging/util.rb
+++ b/lib/packaging/util.rb
@@ -104,6 +104,12 @@ module Pkg::Util
     Pkg::Util.ask_yes_or_no(true)
   end
 
+  def self.filter_configs(filter = nil)
+    return Pkg::Config.instance_values.select { |key, _| key.match(/#{filter}/) } if filter
+    Pkg::Config.instance_values
+  end
+
+
   # Construct a probably-correct (or correct-enough) URI for
   # tools like ssh or rsync. Currently lacking support for intuitive
   # joins, ports, protocols, fragments, or 75% of what Addressable::URI
--- a/lib/packaging/util.rb
+++ b/lib/packaging/util.rb
@@ -98,6 +98,10 @@ module Pkg::Util
   end
 
   def self.confirm_ship(files)
+    $stdout.puts "Artifacts will be shipped to the following hosts:"
+    Pkg::Util.filter_configs('host').each { |key, value| puts "#{key}: #{value}" }
+    $stdout.puts "Does this look right?? [y,n]"
+    Pkg::Util.ask_yes_or_no(true)
     $stdout.puts "The following files have been built and are ready to ship:"
     files.each { |file| puts "\t#{file}\n" unless File.directory?(file) }
     $stdout.puts "Ship these files?? [y,n]"
--- a/tasks/config.rake
+++ b/tasks/config.rake
@@ -1,8 +1,14 @@
 namespace :config do
   desc "print Pkg::Config values for this repo"
   task :print => 'pl:fetch' do
-    Pkg::Config.instance_values.each do |key, value|
+    Pkg::Util.filter_configs.each do |key, value|
       puts "#{key}: #{value}"
+    end
+  end
+
+  task :print_hosts => 'pl:fetch' do
+    Pkg::Util.filter_configs('host').each do |key, value|
+        puts "#{key}: #{value}"
     end
   end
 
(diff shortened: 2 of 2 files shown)


mckern:
Why not `config:print:hosts`?

trollface-png

underscorgan:
@mckern if you really want that I can...

mckern:
... actually, it does make me wonder if we should think about how the config hash is structured, down the road. Not now.

shrug:
👍

lib/packaging/util.rb
```diff
@@ -98,12 +98,25 @@ def self.ask_yes_or_no(force = false)
   end
 
   def self.confirm_ship(files)
+    $stdout.puts "This ship is going to ship to the following hosts:"
```
mckern: Wordsmithing: `"Artifacts will be shipped to the following hosts:"` ?

lib/packaging/util.rb
```diff
@@ -98,12 +98,25 @@ def self.ask_yes_or_no(force = false)
...
+    Pkg::Util.filter_configs('host').each { |key, value| puts "#{key}: #{value}" }
+    $stdout.puts "Does this look right?? [y,n]"
```
mckern: We're calling `$stdout.puts` a lot directly here. By default, `puts` will write to `$stdout` anyway; do we want to clean that up? Doesn't have to be in this PR but it's probably worth thinking about.

lib/packaging/util.rb
```diff
@@ -98,12 +98,25 @@ def self.ask_yes_or_no(force = false)
...
   end
 
+  def self.filter_configs(filter = nil)
+    if filter
+      configs = Pkg::Config.instance_values.select { |key, value| key.match(/#{filter}/) }
+    else
+      configs = Pkg::Config.instance_values
+    end
+    configs
+  end
```
mckern: Structural nitpick but we could cut down a little bit on the intermediate initialization if we jostled a few lines around:

```ruby
def self.filter_configs(filter = nil)
  return Pkg::Config.instance_values.select { |key, _| key.match(/#{filter}/) } if filter
  Pkg::Config.instance_values
end
```

This also makes our intent to discard the values in the `value` iterator, since nothing is being actively done with them.

tasks/config.rake
```diff
@@ -1,7 +1,13 @@
 namespace :config do
   desc "print Pkg::Config values for this repo"
-  task :print do
-    Pkg::Config.instance_values.each do |key, value|
+  task :print => 'pl:fetch' do
+    Pkg::Util.filter_configs.each do |key, value|
+      puts "#{key}: #{value}"
+    end
+  end
```
mckern: Neither here nor there but I find myself wondering if we should consider outputting in a machine readable serialized format like JSON or if the basic `key: value` format really is all we're going to need? Food for thought, not a blocking point.

tasks/config.rake
```diff
@@ -1,7 +1,13 @@
...
+
+  task :print_hosts => 'pl:fetch' do
+    Pkg::Util.filter_configs('host').each do |key, value|
       puts "#{key}: #{value}"
     end
```
mckern: I really like this because it's simple but it's a clever, well structured extension of prior art.

lib/packaging/util.rb
underscorgan: I may be misremembering, but IIRC there are some weird things going on between various levels of rake/ruby extraction which made this output disappear if $stdout wasn't explicit? 😢

mckern: That's entirely possible. We've got a pretty gross stack here.

lib/packaging/util.rb
underscorgan: 👍 good call. updated

lib/packaging/util.rb
underscorgan: updated

lib/packaging/util.rb
underscorgan: I don't think I want to touch this yak here.

tasks/config.rake
mckern: I'd probably make it another task entirely, and we'd just serialize it for output. Something to consider in the future, but not needed for this.

PR #603, opened 2016-12-01
(PA-552) Do not echo osx signing passphrase

underscorgan:
Using the -v flag for `security unlock-keychain` echoes the command that
will be run, which interpolates the environment variable and prints the
passphrase passed on the command line. By switching that to -q (quiet)
the output is no longer printed.


underscorgan:
Output from a test I ran is available in a comment on the [internal ticket](https://tickets.puppetlabs.com/browse/PA-552)

underscorgan:
This should be a lower-impact change than #595, so hopefully it fixes the issue without breaking everything else.

PR #604, opened 2016-12-01
(RE-8186) improved error handling

underscorgan:


lib/packaging/util/execution.rb
```diff
@@ -23,9 +23,9 @@ def success?(statusobject = $?)
     # while also raising an exception if a command does not succeed (ala `sh "cmd"`).
     def ex(command, debug = false)
       puts "Executing '#{command}'..." if debug
-      ret = `#{command}`
+      ret = `#{command} 2>&1`
```
underscorgan: @puppetlabs/release-engineering this is still definitely a wip, but opinions on whether we should be piping stderr like this, doing similar with `Open3.capture2e()`, or breaking apart stdout/stderr with `Open3.capture3()`. 

I feel like capturing both stdout and stderr together will give us the best idea for what went wrong since programs aren't always consistent about where messages go. I don't know a lot about the differences between backticks and Open3 though.

melissa: I think this is consistent with what we're currently doing in a lot of our other automation. I agree with you that capturing both sdtout and stderr together is the way to go.

As to the differences between backticks and Open3, that sounds like a line of investigation we should have. We have. We've always had issues with returning output of commands, and we've yet to actually standardize on one way of doing that. I think backticks are what we've generally been defaulting to, but beyond that, I'm not 100% sure...

melissa:
Did you grep for all the `rescue` blocks in this repo to make sure they're all being useful like those changes in this PR?

lib/packaging/util/execution.rb
underscorgan: We seem to have a mix of backticks and Open3 :\

underscorgan:
@melissa yup, I grepped for all the `rescue` blocks. I didn't print the error messages in every case (but I did test the failures to see if there were errors to print). Some cases the commands being called just didn't have any extra info.

lib/packaging/util/execution.rb
melissa: 🍤

PR #605, opened 2016-12-05
(RE-8405) Run retrieve before deleting tasks

underscorgan:


melissa:
@underscorgan could you add some description about why you're making this change?

underscorgan:
Updated!

PR #607, opened 2016-12-07
(maint) cisco was never added as a platform

underscorgan:


underscorgan:
Tested via `FOSS_ONLY=true pl:fetch pl:jenkins:retrieve` in puppet-agent

PR #609, opened 2016-12-09
(RE-8412) add canary tests

kevpl:


acceptance/tests/canary/yum.rb
```diff
@@ -0,0 +1,62 @@
...
+    # When talking about this originally w/Ryan McKern & Morgan Rhodes, we talked
+    # about potentially getting the code that originally puts an artifact at this
+    # location to provide the URLs rather than having to duplicate that effort
+    # here or using the duplication that we've built into beaker.
+    #
+    # Until we do this work, there's a question of where the tests should be getting
+    # the artifacts they should test from. I'm using a placeholder to create the
+    # tests, but perhaps to make this easier on us we should use a job parameter
+    # and have the testing know how to get all artifacts from a folder URL? -- ki 12/7/16
+    artifact_url = 'http://yum.puppetlabs.com/el/6/PC1/x86_64/puppet-agent-1.8.2-1.el6.x86_64.rpm'
```
kevpl: if you checkout the comment above, you'll see that this is placeholder really, for better work to figure out what the URLs should be set to here for testing purposes. There are a number of options that we could pursue at this point regarding this issue:

- this testing could wait on what we discussed in that meeting, which is to pull the logic that determines these URLs into a separate reference that any number of tools we have at puppet could use, including other tools like beaker. The issue with this is that it makes these canary tests wait on other work yet to be done/scheduled.
- the testing code could call the code that actually builds these URLs, if they're reachable from this point. This is where my knowledge runs out. If this is something we'd like to pursue, please point out to me where that code is, or hopefully we can setup a pairing session to pass that and other info to me about this.
- we could fake up the URL structure with a helper in testing that's not very complicated, but understands just the range of every value needed in the URL so that we could cover the spectrum of available build artifacts. The problem with this is that it's tribal knowledge that needs to be maintained in parallel, and we'll have to know all of the platforms differences as well.
- we could use beaker's logic that builds URLs to install packages from these locations. This option is another very unappealing one to me, as that logic a) will need refactoring to get at alone, and b) is reverse-engineered to begin with, so probably isn't correct, or at the very least has some errors from misunderstanding if not worse.

@underscorgan what do you think of these options? Are there other ones that I'm not thinking of that we should consider?

acceptance/Rakefile
```diff
@@ -29,6 +29,17 @@ namespace :test do
     ]
     sh('beaker', *args)
   end
+
+  desc 'Tests basic artifact information w/packaging tools'
+  task :canary do
```
kevpl: since this is supposed to be the more efficient test, I'm assuming you'd like this task to become the default one, rather than the full install, yes?

acceptance/tests/canary/yum.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    assert(!artifact_path_on_sut.nil?, 'Could not get artifact to test')
+  end
+
+  step 'Test : Artifact exists on SUT' do
+    assert(el_host.file_exist?(artifact_path_on_sut), 'Artifact exists on host')
+  end
+
+  step 'Test : Artifact has a non-zero size' do
+    ls_result = on(el_host, "ls -al #{artifact_path_on_sut}")
```
kevpl: @underscorgan this check is pretty trivial really, but I just wanted to give an example of what many of these checks / asserts look like in beaker tests. If you let me know some simple checks that you would do on an artifact once you have it, I'll do the translation of those to beaker tests so you have a few more examples to go from as we transfer over to more complex tests that will require the RE-specific knowledge that only you guys have.

acceptance/tests/canary/yum.rb
```diff
@@ -0,0 +1,55 @@
...
+    # and have the testing know how to get all artifacts from a folder URL? -- ki 12/7/16
+    artifact_url = 'http://yum.puppetlabs.com/el/6/PC1/x86_64/puppet-agent-1.8.2-1.el6.x86_64.rpm'
+
+    artifact_folder = create_tmpdir_on(el_host)
+    artifact_path_on_sut = "#{artifact_folder}/puppet-agent-1.8.2-1.el6.x86_64.rpm"
+    
+    retry_on(el_host, "curl -o #{artifact_path_on_sut} #{artifact_url}", {
+      :max_retries      => 30,
+      :retry_intervals  => 2
+    })
```
kevpl: I tend to think that a lot of the transient errors that we face are because we don't use beaker's built-in repetition facilities to get over temporary network blips and whatnots that can frequently come up. There's no question here, I just want to make sure to point this out so that as we build out testing for RE we're using the correct patterns that can really make this testing resistant to hiccups that come up all the time in life.

acceptance/Rakefile
underscorgan: 👍

acceptance/tests/canary/yum.rb
underscorgan: I think in the long term I want the first, however, I don't like the thought of blocking on it since as you said, it's unplanned/unscoped. I like the idea of using the code that makes the url but I'll have to find where that is, I will circle back around to this next week. If that ends up being impractical/impossible (which wouldn't be super surprising) it seems like the third would be a decent stop-gap.

acceptance/tests/canary/yum.rb
underscorgan: Cool! Yeah, I'll get you some of these checks early next week!

underscorgan: I'm going to need to think a little bit more about this, but a few generic things that might be helpful:
* size of the downloaded file? Probably want to check against a (configurable in the job params) min and max size, min to make sure we didn't corrupt anything, max is important for agent since we run on some network/embedded systems with size constraints
* Are all the files there? Can we find / fetch files for all the foss supported rpm/deb platforms? Can we find dmgs on downloads.puppetlabs.com? Can we find facter/hiera/mco/puppet tarballs? (those should be parametrizable, won't always have these with agent releases)

kevpl: @underscorgan it sounds like the size ranges will be too platform-specific to be something we'd want to put in job params. It sounds like something that would be specific to the platform-specific tests themselves, and as such should live in the tests. 

Only the file that we've reached out for will be there the way things currently are. We could probably find and fetch files for all of that stuff, but it would be better to just grab what we need and fully specify that, so we're not grabbing things we don't need. If you want every location to be specifiable through job params, then I would suggest having tests that are specific for each thing that you want to test, and then have your beaker tests parameter be a job parameter. It's messy, but it seems to me to be the best way to turn these things off and on at will.

acceptance/tests/canary/yum.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    assert(!artifact_path_on_sut.nil?, 'Could not get artifact to test')
```
mckern: Previous question/statement about "Does this require inversion **and** `#.nil?`" applies here too.

kevpl:
@mckern we could do header calls instead, but I was under the impression we'd want to do checksums & other verifications against these artifacts that would require having them on the system.

Something that we could do that's more of a compromise would be to select certain artifacts at random to do a more in-depth verification, but do header checks on all of the rest.

Looking at other review comments now, thanks for taking a look at this!

acceptance/tests/canary/yum.rb
```diff
@@ -0,0 +1,55 @@
+test_name 'Yum File Checks' do
+  confine :to, :platform => /el/
```
kevpl: :rofl: I actually know nothing about these platforms and whether or not they belong to the same OS-family _enough_ that they can participate in the same tests.

I'm relying on RE's expertise to determine where the line is between the actual testing groups. What should the line be?

stahnma:
@mckern @kevpl what are next steps here?

kevpl:
@stahnma @mckern either before we merge this or after (I like before, so that we can make any changes needed here before it goes in), we create a Jenkins job to run these tests on whatever cadence you'd like. Once that's there, you can use the Jenkins reporting to gain confidence in future changes that you make.

The tests will need to be flushed out before they can really give anyone confidence though. This can happen independently of the job being created. If you guys provide me with a few checks that you'd like to see happen to begin these tests, I can make them into beaker tests for examples.

acceptance/tests/canary/yum.rb
melissa: If these are specifically yum tests, it should be isolated to el varieties (centos, redhat, scientific linux, etc). Sles uses zypper for repo management. Fedora also uses yum, but more recent versions have switched over to dnf, which is "the next generation version of yum". If you call yum on fedora, it'll automagically reroute to using dnf.

acceptance/tests/canary/yum.rb
melissa: Does the work Molly is doing with https://github.com/puppetlabs/packaging/pull/638 impact this at all?

shrug:
@melissa we should pick this up as part of the packaging 1.0 work

melissa:
yeah, sure, that sounds like a logical thing to do

shrug:
@melissa can you pick this up?

mwaggett:
no one is working on this and many things have changed since this work was initially done; closing.

PR #610, opened 2016-12-13
(RE-8186) Improved error handling in packaging

underscorgan:
This is a reimplementation from #604. Rather than modifying the existing `ex()` method, I'm adding a new capture3 method so as not to interfere and change behavior with code using this method in enterprise-dist. I also converted all instances of `ex()` in the packaging repo to use `capture3`.

### DIFF
--- a/lib/packaging/deb/repo.rb
+++ b/lib/packaging/deb/repo.rb
@@ -29,14 +29,14 @@ module Pkg::Deb::Repo
       # First test if the directory even exists
       #
       begin
-        wget_results = Pkg::Util::Execution.ex("#{wget} --spider -r -l 1 --no-parent #{repo_base} 2>&1")
+        stdout, _, _ = Pkg::Util::Execution.capture3("#{wget} --spider -r -l 1 --no-parent #{repo_base} 2>&1")
       rescue RuntimeError
         warn "No debian repos available for #{Pkg::Config.project} at #{Pkg::Config.ref}."
         return
       end
 
       # We want to exclude index and robots files and only include the http: prefixed elements
-      repo_urls = wget_results.split.uniq.reject { |x| x =~ /\?|index|robots/ }.select { |x| x =~ /http:/ }.map { |x| x.chomp('/') }
+      repo_urls = stdout.split.uniq.reject { |x| x =~ /\?|index|robots/ }.select { |x| x =~ /http:/ }.map { |x| x.chomp('/') }
 
 
       # Create apt sources.list files that can be added to hosts for installing
@@ -61,9 +61,10 @@ module Pkg::Deb::Repo
       FileUtils.mkdir_p("pkg/#{target}")
       config_url = "#{base_url}/#{target}/deb/"
       begin
-        Pkg::Util::Execution.ex("#{wget} -r -np -nH --cut-dirs 3 -P pkg/#{target} --reject 'index*' #{config_url}")
-      rescue
-        fail "Couldn't retrieve deb apt repo configs. See preceding http response for more info."
+        stdout, _, _ = Pkg::Util::Execution.capture3("#{wget} -r -np -nH --cut-dirs 3 -P pkg/#{target} --reject 'index*' #{config_url}")
+        stdout
+      rescue => e
+        fail "Couldn't retrieve deb apt repo configs.\n#{e}"
       end
     end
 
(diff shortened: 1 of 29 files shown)


lib/packaging/deb/repo.rb
```diff
@@ -29,14 +29,14 @@ def generate_repo_configs(source = "repos", target = "repo_configs")
       # First test if the directory even exists
       #
       begin
-        wget_results = Pkg::Util::Execution.ex("#{wget} --spider -r -l 1 --no-parent #{repo_base} 2>&1")
+        stdout, _, _ = Pkg::Util::Execution.capture3("#{wget} --spider -r -l 1 --no-parent #{repo_base} 2>&1")
```
melissa: We don't care about stderr at all with these?

lib/packaging/gem.rb
```diff
@@ -91,7 +95,7 @@ def rsync_to_downloads(file)
     # any idea who you are.
     def ship_to_rubygems(file)
       Pkg::Util::File.file_exists?("#{ENV['HOME']}/.gem/credentials", :required => true)
-      Pkg::Util::Execution.ex("gem push #{file}")
+      Pkg::Util::Execution.capture3("gem push #{file}")
```
melissa: should we wrap this in a rescue also?

melissa:
+1, I'm definitely into this change. It seems like you've preserved a lot of the functionality that `ex` was previously doing, so I'd like to see this code in action!

lib/packaging/deb/repo.rb
underscorgan: I was mostly trying to preserve existing behavior, which was just using stdout. If the command fails `capture3` will raise an exception that includes both stdout and stderr, though I think this was one of the cases where wget printed 0 useful information in stderr?

lib/packaging/gem.rb
underscorgan: Yeah probably, I'll add that.

underscorgan: Added!

underscorgan:
@mwaggett yes, definitely! I can work through this with you next time we're both in the office! I feel like it will be easier in person :)

mwaggett:
@underscorgan Sounds good! Hopefully tomorrow. 😸

lib/packaging/gem.rb
melissa: I'm sure there are a bunch of other places that's a good idea, but it looks like your capture3 method also has some failure handling, so we should be okay

underscorgan:
URGH. I need to go through and make sure we aren't checking the status of `$?` anywhere I converted to use this, since apparently we do that. 😡

melissa:
What kind of testing have you done so far?

underscorgan:
@melissa I think I did some testing before break but I can't remember. Did a lot of work with unit testing, but I'll kick off a build now.

underscorgan:
welp. testing issues.

underscorgan:
I have successfully built puppet-agent and puppetserver using this updated code.

lib/packaging/util/jenkins.rb
```diff
@@ -23,7 +23,8 @@ def create_jenkins_job(name, xml_file)
     def jenkins_job_exists?(name)
       job_url = "http://#{Pkg::Config.jenkins_build_host}/job/#{name}/config.xml"
       form_args = ["--silent", "--fail"]
-      _, retval = Pkg::Util::Net.curl_form_data(job_url, form_args, :quiet => true)
+      output, retval = Pkg::Util::Net.curl_form_data(job_url, form_args, :quiet => true)
+      return output if retval.nil?
```
shrug: retval can be nil?

7vuyyav

lib/packaging/util/version.rb
```diff
@@ -67,7 +70,8 @@ def is_git_repo?
     # Return the basename of the project repo
     def git_project_name
       Pkg::Util.in_project_root do
-        Pkg::Util::Execution.ex("#{GIT} config --get remote.origin.url").split('/')[-1].chomp(".git").chomp
+        stdout, _, _ = Pkg::Util::Execution.capture3("#{GIT} config --get remote.origin.url")
+        stdout.split('/')[-1].chomp(".git").chomp
```
mckern: chain-chomp-cat-bed-catastrophicreations-fb

spec/lib/packaging/util/git_tag_spec.rb
```diff
@@ -4,7 +4,7 @@
 describe "Pkg::Util::Git_tag" do
   context "parse_ref!" do
     it "fails for a ref that doesn't exist'" do
-      expect { Pkg::Util::Git_tag.new("git://github.com/puppetlabs/leatherman.git", "garbagegarbage") }.to raise_error(RuntimeError, "ERROR : Not a ref or sha!")
+      expect { Pkg::Util::Git_tag.new("git://github.com/puppetlabs/leatherman.git", "garbagegarbage") }.to raise_error(RuntimeError, /ERROR : Not a ref or sha!/)
```
mckern: Can we get away with not checking the value of the raised error at all? 

```ruby
expect { Pkg::Util::Git_tag.new("git://github.com/puppetlabs/leatherman.git", "garbagegarbage") }.to raise_error(RuntimeError)
```

PR #611, opened 2016-12-14
(RE-8187) Default to failing if anything in the remote_ssh_cmd fails

underscorgan:
This adds `set -e` to the remote ssh command by default.

### DIFF
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -64,8 +64,12 @@ module Pkg::Util::Net
       return errs
     end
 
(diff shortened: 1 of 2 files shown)


mwaggett:
Seems like this makes sense.

underscorgan:
I've used this on a branch to build puppet-agent and got up through signing in an uber_ship, so I'm fairly confident this isn't going to make anything 💥. It's going to be hard to test this exhaustively, probably easiest to test by seeing if it makes pipelines explode?

PR #612, opened 2016-12-14
Fix Bundler's `fetch` not found error

smcclellan:
We were calling the internal `fetch` method inside Bundler::Fetcher. Upstream
has started calling this an internal method, and moved it around, so now we
should be using the public version of this functionality.


smcclellan:
Tracked by https://github.com/puppetlabs/packaging/pull/613 instead.

PR #613, opened 2016-12-14
(RE-8415) Fix Bundler's `fetch` not found error

smcclellan:
We were calling the internal `fetch` method inside Bundler::Fetcher. Upstream
has started calling this an internal method, and moved it around, so now we
should be using the public version of this functionality.

### DIFF
--- a/tasks/vendor_gems.rake
+++ b/tasks/vendor_gems.rake
@@ -105,6 +105,9 @@ if Pkg::Config.pre_tar_task
         elsif Bundler::Fetcher.respond_to?(:fetch, true)
           # Retain this call in case we are dealing with an older version of bundler.
           Bundler::Fetcher.fetch(spec) if spec.source.is_a?(Bundler::Source::Rubygems)
+        else
+          # shrug
+          fail "Cannot find a way to download the gem for caching! Did the `fetch_gem` method move?"
         end
         # Cache everything but bundler itself...
         spec.source.cache(spec) unless spec.name == "bundler"
--- a/tasks/vendor_gems.rake
+++ b/tasks/vendor_gems.rake
@@ -100,11 +100,15 @@ if Pkg::Config.pre_tar_task
       # Cache the gems
       definition.specs.each do |spec|
         # Fetch Rubygem specs
-        Bundler::Fetcher.fetch(spec) if spec.source.is_a?(Bundler::Source::Rubygems)
+        if spec.source.respond_to?(:fetch_gem, true)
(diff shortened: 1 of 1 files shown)


underscorgan:
lots of testing still going on here

tasks/vendor_gems.rake
```diff
@@ -100,11 +100,12 @@ if Pkg::Config.pre_tar_task
...
       definition.specs.each do |spec|
         # Fetch Rubygem specs
-        Bundler::Fetcher.fetch(spec) if spec.source.is_a?(Bundler::Source::Rubygems)
-        # Cache everything but bundler itself...
-        spec.source.cache(spec) unless spec.name == "bundler"
+        if spec.source.respond_to?(:fetch_gem, true)
+          spec.source.send(:fetch_gem, spec) if spec.source.is_a?(Bundler::Source::Rubygems)
+          # Cache everything but bundler itself...
+          spec.source.cache(spec) unless spec.name == "bundler"
+        end
```
underscorgan: My one question here is do you know how this works with older bundlers  / does this method exist? I don't know that this exists anywhere, but it might be worthwhile to do a begin/rescue where we try the new one and if that doesn't exist fail back to the old behavior?

smcclellan: It didn't exist in the old version, so I'll add that as a backup.

tasks/vendor_gems.rake
```diff
@@ -100,11 +100,15 @@ if Pkg::Config.pre_tar_task
       # Cache the gems
       definition.specs.each do |spec|
         # Fetch Rubygem specs
-        Bundler::Fetcher.fetch(spec) if spec.source.is_a?(Bundler::Source::Rubygems)
+        if spec.source.respond_to?(:fetch_gem, true)
+          spec.source.send(:fetch_gem, spec) if spec.source.is_a?(Bundler::Source::Rubygems)
+        elsif Bundler::Fetcher.respond_to?(:fetch, true)
+          # Retain this call in case we are dealing with an older version of bundler.
+          Bundler::Fetcher.fetch(spec) if spec.source.is_a?(Bundler::Source::Rubygems)
+        end
```
underscorgan: lolsob. Probably should be worded, "what the hell, bundler changed their API again" for consistency :P

smcclellan: It's not technically the API, since it's a protected method anyway... but I'll add an error to this effect!

haus:
conan-obrien-shock-then-approval

mwaggett:
I feel like those commits should be squashed..

PR #615, opened 2016-12-16
(RE-8187) fail more!

underscorgan:
This adds `set -e` to `remote_ssh_cmd` and also fixes the two cases I found where we were depending on a failure in a command passed to `remote_ssh_cmd`

### DIFF
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -64,8 +64,12 @@ module Pkg::Util::Net
       return errs
     end
 
-    def remote_ssh_cmd(target, command, capture_output = false, extra_options = '')
+    def remote_ssh_cmd(target, command, capture_output = false, extra_options = '', fail_fast = true)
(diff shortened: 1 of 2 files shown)


underscorgan:
testing with a puppetserver build right now

underscorgan:
http://jenkins-release.delivery.puppetlabs.net/job/puppetserver-packaging-2016-12-16-09-46-27-2.7.2.master.SNAPSHOT.2016.12.16T0946/ is the test url, now that I've worked through some oddities in my testing env.

underscorgan:
I believe these are the only places we were depending on failures in commands passed to `remote_ssh_cmd`. I checked this by looking for where we were using `$?`, and, in particular, where we were checking for `-eq 1` in packaging and enterprise-dist. There are a few other places where we are doing that, but not in strings that get passed to `remote_ssh_cmd` so it _should_ be fine.

Puppetserver build above worked and was definitely using the new code, so at least there's that!

PR #616, opened 2016-12-20
(PE-16465) Make a tarball of all signed repos

shrug:
This adds a set of tasks to make a tarball of *all* of the
prepared repos, in addition to the individual per-platform tarballs.

This is useful for cutomers who have a PE master with
no internet access. They can unpack the puppet-agent-all tarball into the location that
pe_repo expects and use simplified agent install without needing internet access, or having to
manually download each agent that they need to feed to pe_repo.

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -73,6 +73,60 @@ namespace :pl do
       end
     end
 
+    # This is pretty similar to the 'pack_signed_repo' task. The difference here is that instead
+    # of creating a tarball for each repo passed, it adds each repo to a single archive, creating
+    # one 'all' tarball with all of the repos. This is useful for cutomers who have a PE master with
+    # no internet access. They can unpack the puppet-agent-all tarball into the location that
+    # pe_repo expects and use simplified agent install without needing internet access, or having to
+    # manually download each agent that they need to feed to pe_repo.
+    # This task should be invoked after prepare_signed_repos, so that there are repos to pack up.
+    task :pack_all_signed_repos, [:path_to_repo, :name_of_archive, :versioning] => ["pl:fetch"] do |t, args|
+      # path_to_repo should be relative to ./pkg
+      name_of_archive = args.name_of_archive or fail ":name_of_archive is a required argument for #{t}"
+      versioning = args.versioning or fail ":versioning is a required argument for #{t}"
+      tar = Pkg::Util::Tool.check_tool('tar')
+
+      Dir.chdir("pkg") do
+        if versioning == 'ref'
+          local_target = File.join(Pkg::Config.project, Pkg::Config.ref, "repos")
+        elsif versioning == 'version'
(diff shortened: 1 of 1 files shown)


stahnma:
How much time does this add to the compose process, or does this happen elsewhere? If so, where?

shrug:
this would happen in the puppet-agent 'internal ship' and 's3 ship' jobs

shrug:
Oops, I misunderstood what pe_repo wanted to see, need to rework this a bit

shrug:
OK this seems to work. Unpack 'puppet-agent-all.tar.gz' into the pe_repo staging area (where it downloads the S3 tarballs):
```/opt/puppetlabs/server/data/staging/pe_repo-puppet-agent-1.8.2```

and it looks like you can apply new pe_repo classes on the master and it will use the files already in the staging directory instead of going to S3

File: tasks/nightly_repos.rake [hunk 1]
```diff
@@ -73,6 +73,61 @@ namespace :pl do
...
     end
 
+    # This is pretty similar to the 'pack_signed_repo' task. The difference here is that instead
+    # of creating a tarball for each repo passed, it adds each repo to a single archive, creating
+    # one 'all' tarball with all of the repos. This is useful for cutomers who have a PE master with
+    # no internet access. They can unpack the puppet-agent-all tarball into the location that
+    # pe_repo expects and use simplified agent install without needing internet access, or having to
+    # manually download each agent that they need to feed to pe_repo.
+    # This task should be invoked after prepare_signed_repos, so that there are repos to pack up.
+    # This task should be invoked after prepare_signed_repos, so that there are repos to pack up.
```
underscorgan: duplicate sentence in comment

tasks/nightly_repos.rake
```diff
@@ -73,6 +73,61 @@ namespace :pl do
...
+    task :pack_all_signed_repos, [:path_to_repo, :name_of_archive, :versioning] => ["pl:fetch"] do |t, args|
+      # path_to_repo should be relative to ./pkg
+      name_of_archive = args.name_of_archive or fail ":name_of_archive is a required argument for #{t}"
+      versioning = args.versioning or fail ":versioning is a required argument for #{t}"
+      tar = Pkg::Util::Tool.check_tool('tar')
+
+      Dir.chdir("pkg") do
+        if versioning == 'ref'
+          local_target = File.join(Pkg::Config.project, Pkg::Config.ref)
```
underscorgan: should this still have 'repos' like below?

tasks/nightly_repos.rake
shrug: derp.

tasks/nightly_repos.rake
shrug: Yeah, the actual use case for this hardcodes 'version' so this wasn't getting executed, but you're right. Fixed.

shrug:
using agent 1.8, the 'puppet-agent-all' tarball is 1.7GB. Not ideal but better than nothing for people without internet access on their masters

PR #620, opened 2017-01-09
(RE-8186) Using pipes in Open3.capture3 with JRuby is special

underscorgan:
After the initial work for this was merged in, failures became apparent
with razor packaging. Razor builds using JRuby 1.7.19. Whereas with ruby
2.1.8, we see
```
irb(main):004:0> Open3.capture3("echo 'hi' >/dev/null")
=> ["", "", #<Process::Status: pid 95032 exit 0>]
```
with JRuby 1.7.9 the same command results in
```
irb(main):005:0> Open3.capture3("echo 'hi' >/dev/null")
=> ["hi >/dev/null\n", "", #<Process::Status: pid=95178,exited(0)>]
```

This commit updates unnecessary piping in the code that checks
versioning and updates the method for curling to discard stdout when
curling quietly rather than dealing with pipes.

There are still pipes elsewhere in the code but this removes them from
enough places to allow razor to build successfully, and as that is the
only project that builds using jruby this should be sufficient.

### DIFF
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -233,11 +233,11 @@ module Pkg::Util::Net
       post_string << "#{uri}"
 
       # If this is quiet, we're going to silence all output
-      if options[:quiet]
-        post_string << " >#{Pkg::Util::OS::DEVNULL} 2>&1"
-      end
       begin
         stdout, _, retval = Pkg::Util::Execution.capture3("#{curl} #{post_string}")
+        if options[:quiet]
+          stdout = ''
+        end
         return stdout, retval
       rescue RuntimeError => e
         puts e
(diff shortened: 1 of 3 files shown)


underscorgan:
razor built successfully. p-a and puppetserver building right now.

shrug:
👍

mckern:
😭

lib/packaging/util/net.rb
```diff
@@ -233,11 +233,11 @@ def curl_form_data(uri, form_data = [], options = {})
...
 
       # If this is quiet, we're going to silence all output
-      if options[:quiet]
-        post_string << " >#{Pkg::Util::OS::DEVNULL} 2>&1"
-      end
       begin
         stdout, _, retval = Pkg::Util::Execution.capture3("#{curl} #{post_string}")
+        if options[:quiet]
+          stdout = ''
+        end
```
mckern: I cannot unsee this 😞 .
I hate our tools.

PR #622, opened 2017-01-19
(maint) Double the timeouts, double the fun

underscorgan:
We're having a bunch of packaging timeouts on the deb builders possibly
due to hardware oversubscription. There is more hardware planned for
these boxes, but in the interim let's increase the timeout so fewer
things fail.


steveax:
👍 yes please!

PR #623, opened 2017-01-23
(maint) Make sure aptly isn't running before we try to update

underscorgan:
This adds a loop to make sure aptly isn't running before we try to lock
the database.  LevelDB doesn't remove/re-add its lockfile each time it
takes/releases the lock, rather it seems to be doing some filesystem
locking.

### DIFF
--- a/tasks/pe_ship.rake
+++ b/tasks/pe_ship.rake
@@ -159,13 +159,19 @@ if Pkg::Config.build_pe
         incoming_dir or fail "Adding packages to apt repo requires an incoming directory"
         Pkg::Util::RakeUtils.invoke_task("pl:fetch")
 
-        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "aptly repo add -remove-files #{Pkg::Config::pe_version}-#{dist} #{incoming_dir}/#{dist}
+        cmd = <<-eos
+          if ! flock /opt/tools/aptly/db/LOCK --wait 1200 --command /bin/true; then
+            echo "Unable to acquire aptly lock, giving up" 1>&2
+            exit 1
+          fi
+          aptly repo add -remove-files #{Pkg::Config::pe_version}-#{dist} #{incoming_dir}/#{dist}
           if [ -d /opt/tools/aptly/public/#{Pkg::Config::pe_version}/dists/#{dist} ]; then
             aptly publish update -gpg-key=\"8BBEB79B\" #{dist} #{Pkg::Config::pe_version}
           else
             aptly publish repo -gpg-key=\"8BBEB79B\" #{Pkg::Config::pe_version}-#{dist} #{Pkg::Config::pe_version}
           fi
-          ", true)
+        eos
+        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, cmd, true)
 
         output = stdout.to_s + stderr.to_s
 


tasks/pe_ship.rake
```diff
@@ -159,13 +159,19 @@ if Pkg::Config.build_pe
         incoming_dir or fail "Adding packages to apt repo requires an incoming directory"
         Pkg::Util::RakeUtils.invoke_task("pl:fetch")
 
-        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "aptly repo add -remove-files #{Pkg::Config::pe_version}-#{dist} #{incoming_dir}/#{dist}
+        cmd = <<-eos
+          while pgrep aptly > /dev/null; do
```
shrug: We can use flock() semantics?

mckern: ^^ What he said

underscorgan: I'm going to dig more in the db implementation to try to figure out what they're actually calling to lock the files.

mckern: I was working backwards from this thread: https://groups.google.com/forum/#!topic/aptly-discuss/Gsemphn9CEo

underscorgan:
@shrug @mckern updated

mckern:
kum4bxe

Does it blend? It looks like it should blend.

underscorgan:
I'm not sure how to test this without merging it since it's in the promotion codepath.
`jenkins@saturn:~$ flock /opt/tools/aptly/db/LOCK -c /bin/true` executes successfully though.

mckern:
@underscorgan suggestion:

On any system with Aptly and a pile of packages on it, kick off an Aptly `repo -add` in the background/another tab or session. Run a second Aptly `repo -add` with the `flock` code in a separate session. See if it blends or if it explodes in a shower of sparks and matchbox car parts?

underscorgan:
@mckern good call. Spun up a vm, mirrored PC1 for wheezy, imported that into a local repo. That took a few minutes. In another window started `flock .aptly/db/LOCK -c "date"`, it returned after the publishing of the local repo finished.

tasks/pe_ship.rake
```diff
@@ -159,13 +159,16 @@ if Pkg::Config.build_pe
         incoming_dir or fail "Adding packages to apt repo requires an incoming directory"
         Pkg::Util::RakeUtils.invoke_task("pl:fetch")
 
-        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "aptly repo add -remove-files #{Pkg::Config::pe_version}-#{dist} #{incoming_dir}/#{dist}
+        cmd = <<-eos
+          flock /opt/tools/aptly/db/LOCK -c /bin/true
```
shrug: Do you think we should have a timeout here? I don't have an opinion on what might be sane... something long-ish?

underscorgan: 15 minutes? I feel like that should be long enough, and probably better to let the job explode in failing to update aptly than hang?

tasks/pe_ship.rake
```diff
@@ -159,13 +159,19 @@ if Pkg::Config.build_pe
         incoming_dir or fail "Adding packages to apt repo requires an incoming directory"
         Pkg::Util::RakeUtils.invoke_task("pl:fetch")
 
-        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "aptly repo add -remove-files #{Pkg::Config::pe_version}-#{dist} #{incoming_dir}/#{dist}
+        cmd = <<-eos
+          if ! flock /opt/tools/aptly/db/LOCK --wait 1200 --command /bin/true; then
+            2&> echo "Unable to acquire aptly lock, giving up"
+            exit 1
+          fi
```
underscorgan: @shrug this look reasonable?

tasks/pe_ship.rake
```diff
@@ -159,13 +159,19 @@ if Pkg::Config.build_pe
         incoming_dir or fail "Adding packages to apt repo requires an incoming directory"
         Pkg::Util::RakeUtils.invoke_task("pl:fetch")
 
-        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.apt_host, "aptly repo add -remove-files #{Pkg::Config::pe_version}-#{dist} #{incoming_dir}/#{dist}
+        cmd = <<-eos
+          if ! flock /opt/tools/aptly/db/LOCK --wait 1200 --command /bin/true; then
+            2&> echo "Unable to acquire aptly lock, giving up"
```
mckern: This is not an "output to STDERR" format that I've seen before. Look at that. But it doesn't work.

```bash
mckern@flexo errors $ ls -1
echo_errors.sh
mckern@flexo errors $ bash ./echo_errors.sh
mckern@flexo errors $ bash -x ./echo_errors.sh
+ 2 'Unable to acquire aptly lock, giving up'
mckern@flexo errors $ echo
mckern@flexo errors $ ls -1
echo
echo_errors.sh
mckern@flexo errors $ cat echo
echo            echo_errors.sh
mckern@flexo errors $ cat echo
./echo_errors.sh: line 3: 2: command not found
mckern@flexo errors $
```

underscorgan: I'm not real familiar with bash outputting to stderr, this is what I found on stackoverflow. Do you have one that works/you like?

mckern: Here's an output redirector cheatsheat I cribbed from somewhere else:

| Stream         | Description                                            |
|----------------|--------------------------------------------------------|
| `cmd > file` | Redirect the standard output (stdout) of cmd to a file.|
| `cmd 2>&1` | Redirect stderr to stdout. `2` is the file descriptor for stderr.|
| `cmd 1>&2` | Redirect stdout to stderr.|
| `cmd >&2` | Same as `cmd 1>&2`. `1` can often be omitted when redirecting stderr.|
| `cmd 1> file` | Same as `cmd > file`. `1` is the default file descriptor (fd) for stdout.|
| `cmd 2> file` | Redirect the standard error (stderr) of cmd to a file. 2 is the default fd for stderr.|
| `cmd >> file` | Append stdout of cmd to a file.|
| `cmd 2>> file` | Append stderr of cmd to a file.|
| `cmd > file 2>&1` | Redirect both stdout and stderr of cmd to a file. This is not the same as cmd 2>&1 > file. Redirection order matters!|

mckern: (FWIW, redirection at the front of the line will work if constructed like this:

```bash
mckern@flexo errors $ bash ./echo_errors.sh >/dev/null
Front of line redirection
mckern@flexo errors $ cat echo_errors.sh
#!/bin/bash

>&2 echo "Front of line redirection"

mckern@flexo errors $
```

underscorgan: ah, cool. thank you. I think I got to over thinking this one.

underscorgan:
@mckern updated!

PR #624, opened 2017-01-24
(RE-6628) Add new pack_signed_repo tasks

McdonaldSeanp:
Our automation is going to need to handle more that one set of platforms to
promote to PE. Instead of adding seperate jobs for seperate pipelines (i.e.
one for master, one for stable etc.) and continuing to carry around static
lists on top of information we already have, this commit adds the ability to
use the platform information in build_defaults.yaml to identify what platforms
promote.

This is done through a new pack_all_signed_repos_indivudually task, to replace
the pack_signed_repo task (that created the need for the list in the jenkins
job).

to faciliate this, the platform_repos data has been added to the available
data loaded in build_defaults.yaml, this data is in the following form:

platform_repos:
  - name: platform-version-arch
    repo_location: repos/platform/repo/location

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -125,6 +125,7 @@ module Pkg::Params
                   :pe_platforms,
                   :pe_version,
                   :pg_major_version,
+                  :platform_repos,
                   :pre_tar_task,
                   :pre_tasks,
                   :privatekey_pem,
@@ -324,5 +325,4 @@ module Pkg::Params
                   { :var => :gpg_name, :message => "
     DEPRECATED, 29-Jul-2014: 'gpg_name' has been replaced with 'gpg_key'.
                    Please update this field in your project_data.yaml" }]
-
 end
--- a/lib/packaging/repo.rb
+++ b/lib/packaging/repo.rb
@@ -1,0 +1,36 @@
+module Pkg::Repo
+
+  class << self
+    def create_signed_repo_archive(path_to_repo, name_of_archive, versioning)
+      tar = Pkg::Util::Tool.check_tool('tar')
+      Dir.chdir("pkg") do
+        if versioning == 'ref'
+          local_target = File.join(Pkg::Config.project, Pkg::Config.ref)
+        elsif versioning == 'version'
+          local_target = File.join(Pkg::Config.project, Pkg::Util::Version.get_dot_version)
+        end
+        Dir.chdir(local_target) do
+          if Pkg::Util::File.empty_dir?(path_to_repo)
+            if ENV['FAIL_ON_MISSING_TARGET'] == "true"
+              raise "ERROR: missing packages under #{path_to_repo}"
+            else
+              warn "Skipping #{name_of_archive} because #{path_to_repo} has no files"
+            end
+          else
+            puts "Archiving #{path_to_repo} as #{name_of_archive}"
+            stdout, _, _ = Pkg::Util::Execution.capture3("#{tar} --owner=0 --group=0 --create --gzip --file #{File.join('repos', "#{name_of_archive}.tar.gz")} #{path_to_repo}")
+            stdout
+          end
+        end
+      end
+    end
+
+    def create_all_repo_archives(project, versioning)
+      platforms = Pkg::Config.platform_repos
+      platforms.each do |platform|
+        archive_name = "#{project}-#{platform['name']}"
+        create_signed_repo_archive(platform['repo_location'], archive_name, versioning)
+      end
+    end
+  end
+end
--- a/spec/lib/packaging/repo_spec.rb
+++ b/spec/lib/packaging/repo_spec.rb
@@ -1,0 +1,97 @@
+# -*- ruby -*-
+require 'spec_helper'
+describe "#Pkg::Repo" do
+  let(:platform_repo_stub) do
+    [
+      {"name"=>"el-4-i386", "repo_location"=>"repos/el/4/**/i386"},
+      {"name"=>"el-5-i386", "repo_location"=>"repos/el/5/**/i386"},
+      {"name"=>"el-6-i386", "repo_location"=>"repos/el/6/**/i386"}
+    ]
+  end
+  describe "#create_signed_repo_archive" do
+    it "should change to the correct dir" do
+      allow(Pkg::Util::Tool).to receive(:check_tool).and_return("tarcommand")
+      allow(Pkg::Config).to receive(:project).and_return("project")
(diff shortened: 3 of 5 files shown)


McdonaldSeanp:
This PR should be purely additive for now, when we go to update the jenkins jobs that run the promotion steps, then we will need to add the new data to build_data.yaml. but for now, everything should work fine with the old jobs.

mckern:
I'm going to take another look at this in the morning after I sleep but at 1:00am, nothing jumped out at me as WTFery.

lib/packaging/repo.rb
```diff
@@ -0,0 +1,37 @@
+module Pkg::Repo
+
+  class << self
+    def create_signed_repo_archive(path_to_repo, name_of_archive, versioning)
+      tar = Pkg::Util::Tool.check_tool('tar')
+      puts "DEBUG: #{name_of_archive}"
+      puts "DEBUG: #{path_to_repo}"
```
underscorgan: These puts should probably be behind a verbose flag or removed?

lib/packaging/repo.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def create_all_repo_archives(project, versioning)
+      platforms = Pkg::Config.platform_repos
+      platforms.each do |platform|
+        archive_name = "#{project}-#{platform['name']}"
+        create_signed_repo_archive(platform['repo_location'], archive_name, versioning)
+      end
+    end
+  end
+end
```
underscorgan: [nitpick] no newline at end of file.

spec/lib/packaging/repo_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    it "should invoke create_signed_repo_archive correctly for multiple entries in platform_repos" do
+      allow(Pkg::Config).to receive(:platform_repos).and_return(platform_repo_stub)
+
+      expect(Pkg::Repo).to receive(:create_signed_repo_archive).with("repos/el/4/**/i386", "project-el-4-i386", "version")
+      expect(Pkg::Repo).to receive(:create_signed_repo_archive).with("repos/el/5/**/i386", "project-el-5-i386", "version")
+      expect(Pkg::Repo).to receive(:create_signed_repo_archive).with("repos/el/6/**/i386", "project-el-6-i386", "version")
+      Pkg::Repo.create_all_repo_archives("project", "version")
+    end
+  end
+end
```
underscorgan: [nitpick] no newline at end of file

lib/packaging/config/params.rb
```diff
@@ -125,6 +125,7 @@ module Pkg::Params
                   :pe_platforms,
                   :pe_version,
                   :pg_major_version,
+                  :platform_repos,
```
underscorgan: Should this be overrideable by an environment variable?

underscorgan:
[nitpick] You have individually misspelled in your commit message. Commit title is a little misleading since you are adding new functionality not changing the nightly_repo job in the commit.

lib/packaging/repo.rb
McdonaldSeanp: Oh, yeah. Whoops

lib/packaging/repo.rb
McdonaldSeanp: On it

spec/lib/packaging/repo_spec.rb
McdonaldSeanp: On it.

lib/packaging/config/params.rb
McdonaldSeanp: I don't think so. What is the precedent for the pe/foss_platform vars? I don't think you can override those either.

McdonaldSeanp:
@underscorgan @mckern updated

tasks/nightly_repos.rake
```diff
@@ -54,24 +54,13 @@ namespace :pl do
...
-        Dir.chdir(local_target) do
-          if Pkg::Util::File.empty_dir?(path_to_repo)
-            warn "Skipping #{name_of_archive} because it (#{path_to_repo}) has no files"
-          else
-            stdout, _, _ = Pkg::Util::Execution.capture3("#{tar} --owner=0 --group=0 --create --gzip --file #{File.join("repos", "#{name_of_archive}.tar.gz")} #{path_to_repo}")
-            stdout
-          end
-        end
-      end
+    task :pack_all_signed_repos_individually, [:name_of_archive, :versioning] => ["pl:fetch"] do |t, args|
```
shrug: We already have a pack_all_signed_repos task. Naturally.

McdonaldSeanp: yep, that's what the individually was for.

mckern: 6945118-sad-puppy-face

lib/packaging/config/params.rb
underscorgan: You can't override pe_platforms but can for foss_platforms.

tasks/nightly_repos.rake
```diff
@@ -54,24 +54,13 @@ namespace :pl do
...
-          local_target = File.join(Pkg::Config.project, Pkg::Config.ref)
-        elsif versioning == 'version'
-          local_target = File.join(Pkg::Config.project, Pkg::Util::Version.get_dot_version)
-        end
+      Pkg::Repo.create_signed_repo_archive(path_to_repo, name_of_archive, versioning)
+    end
 
-        Dir.chdir(local_target) do
-          if Pkg::Util::File.empty_dir?(path_to_repo)
-            warn "Skipping #{name_of_archive} because it (#{path_to_repo}) has no files"
```
underscorgan: I am curious why the jenkins job fails, it looks like this should just warn if there are no files? Is that just really badly written?

McdonaldSeanp: it would not fail here. I added the failure myself, because I want the jobs to fail when there are not packages, not just warn. Then optionally, you can opt-in to just have packaging warn if there are not packages.

The job should default to raising an error because if there are not packages I don't want to find out something did not internally ship from downstream jobs. I want it to fail here and let me fix it. 

Then, if we need to only ship certain packages (for instance when we are fixing an internal ship) we can opt-in to only having it warn.

McdonaldSeanp: Now that I think about it the internal job *does* fail when packages are missing, so yeah, something is amiss.

underscorgan: ok.

McdonaldSeanp:
This has been tested (beyond the spec tests) and performs correctly with the agent.

PR #628, opened 2017-02-06
(RE-8596) Ensure vanagonized projects end up in the appropriate project folder

mwaggett:
This commit adds `PROJECT` and `REF` environment variables that can be used to override the default build params.
This will allow us to specify what project we are building for when building from a mono repo, like puppet-enterprise-vanagon, so that builds end up in the appropriate place: builds.delivery.puppetlabs.net/<project>/<ref>/artifacts/...


melissa:
I'm very uncomfortable with the idea of setting REF manually

mwaggett:
@melissa I definitely agree. Open for suggestions on other ways to accomplish this.

shrug:
I'm not sure I understand the issue - what's happening?

mwaggett:
@shrug When we ship from puppet-enterprise-vanagon, packages end up in a puppet-enterprise-vanagon directory, rather than in a directory for project we actually built. See https://tickets.puppetlabs.com/browse/RE-8596

melissa:
Do you need to be able to override REF? We should only need to change PROJECT

shrug:
Gotcha, sorry I should have read the ticket. 

pe-r10k seems to do the right thing though, may be something else we're doing wrong?

mwaggett:
@melissa If we don't override REF, it gets the ref from the puppet-enterprise-vanagon project, rather than from the project we're building. I don't think that's what we want..

mwaggett:
@shrug pe-r10k is fine because it's not coming from a mono repo. It has its own build_defaults: https://github.com/puppetlabs/pe-r10k-vanagon/blob/flanders/ext/build_defaults.yaml

melissa:
I feel like you're overcomplicating this. If we have the ref of the puppet-enterprise-vanagon project, that's more than sufficient. That's always unique and we can use that to indicate changes to the repo. As long as packages are separated out based on name, that's fine.

Side question: what happens if we want to build and ship multiple packages at the same time?

underscorgan:
I think pe-r10k works because it gets the version from git. without overriding ref how do we know where to pull the packages from when we're doing the compose? they could be under any sha.

shrug:
We promote a specific ref, the compose pulls from the PE compose repos, it's completely disconnected from builds.

We can split these out to their own repos, that might be the simplest if people are concerned about this.

mwaggett:
@melissa I thought about that question and don't have an answer. Or rather, the answer right now is that everything staged in /output will end up under puppet-enterprise-vanagon, but that's not what we want.

underscorgan:
@shrug mostly I was wondering about how we remember where we put builds, or how to keep them from getting reaped (although I guess after promotion that doesn't really matter since it'll get pulled into enterprise repos)

mwaggett:
okay, our solution is to have separate repos for each project we vanagonize.

PR #631, opened 2017-03-01
(maint) raise errors if you fail the gems

underscorgan:


melissa:
Does this actually fail the job now?

underscorgan:
@melissa yup! moved my gem credentials out of place and tried to ship something and 💥

PR #636, opened 2017-04-11
(maint) Let pez promotions overwrite packages

shrug:


tasks/pe_ship.rake
```diff
@@ -5,13 +5,19 @@ if Pkg::Config.build_pe
       puts "Shipping packages to #{Pkg::Config.yum_target_path}"
       Pkg::Util::File.empty_dir?("pkg/pe/rpm") and fail "The 'pkg/pe/rpm' directory has no packages. Did you run rake pe:deb?"
       Pkg::Util::Execution.retry_on_fail(:times => 3) do
-        Pkg::Util::Net.rsync_to('pkg/pe/rpm/', Pkg::Config.yum_host, Pkg::Config.yum_target_path)
+        if Pkg::Config.pe_feature_branch
+          puts "On a feature branch - overwrites alllowed"
+          Pkg::Util::Net.rsync_to('pkg/pe/rpm/', Pkg::Config.yum_host, Pkg::Config.yum_target_path, extra_flags: [])
```
mwaggett: Can you tell me what `extra_flags` does?

underscorgan: `extra_flags` lets you pass extra flags to rsync. It defaults to `--ignore-existing` because we usually want that option, but that's breaking PEZ if it gets run multiple times with the same build number, and for PEZ we always want to ship the newest packages.

mwaggett: ah okay. This may be too much work and outside the scope of this PR, but I feel like it makes more sense for a default to be `[]` and then explicitly set `--ignore-existing` or whatever when we need it?

underscorgan: :woman_shrugging: I think I'm more comfortable with `--ignore-existing` as the default behavior since by default that is what we want. I'd also definitely be wary of that change since we're using that code all over the place (probably across a few repos and in hand crafted jenkins releases) and it will break things.

mwaggett: yeah, I figured it was. I just feel like it could be clearer, but I understand that that may not be worth the effort.

PR #638, opened 2017-04-19
(RE-8707) Add `platform_data`

mwaggett:
This PR adds a `platform_data` array that maps build targets to where their artifacts and repo_configs (where applicable) can be found. It also includes a function to manipulate `platform_data` for PC1, which will be removed once we stop shipping to PC1. 

This info will get included in the <ref>.yaml file that gets generated for each project. All platforms will be listed for each project, even if we don't build for all platforms in each project.

Note: EL 4, Fedora 23, and Ubuntu 12.04 have been excluded, since they are going EOL.


mwaggett:
Built yaml files for puppet-agent: http://builds.delivery.puppetlabs.net/puppet-agent/456ec32105d126ae2a64540ba734bb889840b5f9/artifacts/456ec32105d126ae2a64540ba734bb889840b5f9.yaml
and puppetserver: http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.04.19T1053/artifacts/5.0.0.master.SNAPSHOT.2017.04.19T1053.yaml

lib/packaging/config.rb
```diff
@@ -46,11 +46,84 @@ def config_from_hash(data = {})
       end
 
       ##
+      #   Array of platforms and relative paths to their artifacts and
+      #   repo_configs (where applicable)
+      #
+      def platform_data
+      [ {"name" => "aix-5.3-power", "artifact" => "./puppet/aix/5.3/ppc/#{self.project}-#{Pkg::Util::Version.git_describe}-#{self.release}.aix5.3.ppc.rpm", "repo_config" => "../repo_configs/rpm/pl-#{self.project}-#{self.ref}-aix-5.3-ppc.repo"},
```
melissa: Can we not generate these paths? I really really don't like the thought of having to maintain these hardcoded paths..... Also, I'm not seeing how they're going to vary when we change PC1 and puppet? 

It should work with all projects that have a yaml file, which have a variety of pathing formats

`el/6/PC1/x86_64/puppet-agent-1.9.3-1.el6.x86_64.rpm`
`sles/12/products/x86_64/pe-r10k-2.5.4.1-1.sles12.x86_64.rpm`
`pe/rpm/el-6-x86_64/pe-razor-server-1.4.0.0-1.pe.el6.noarch.rpm`

shrug: Eventually this should live in the data service, that will make it possible to generate these more dynamically. 

The immediate need is for puppet-agent and beaker, so that's where the focus has been. 

Do you have a suggestion that is both more dynamic and works well for every project, before we get it in the data service?

lib/packaging/config.rb
```diff
@@ -46,11 +46,84 @@ def config_from_hash(data = {})
...
       # Load a yaml file and use its contents to set the values for Pkg::Config
       # class instance variables
       #
       def config_from_yaml(file)
         build_data = Pkg::Util::Serialization.load_yaml(file)
+        if self.apt_repo_name == 'PC1'
+          build_data[:platform_data] = platform_data_pc1
+        else
+          build_data[:platform_data] = platform_data
+        end
```
underscorgan: Adding these lines are what's causing the unit test failures. You at least need to stub out the `Pkg::Util::Version.git_describe` method so `platform_data(_pc1)` can run successfully in a testing environment.

lib/packaging/config.rb
```diff
@@ -46,11 +46,84 @@ def config_from_hash(data = {})
...
+        {"name" => "ubuntu-16.04-amd64", "artifact" => "./puppet/deb/xenial#{self.project}_#{Pkg::Util::Version.git_describe}-#{self.release}xenial_amd64.deb", "repo_config" => "../repo_configs/deb/pl-#{self.project}-#{self.ref}-xenial.list"},
+        {"name" => "ubuntu-16.10-i386", "artifact" => "./puppet/deb/yakkety/#{self.project}_#{Pkg::Util::Version.git_describe}-#{self.release}yakkety_i386.deb", "repo_config" => "../repo_configs/deb/pl-#{self.project}-#{self.ref}-yakkety.list"},
+        {"name" => "ubuntu-16.10-amd64", "artifact" => "./puppet/deb/yakkety/#{self.project}_#{Pkg::Util::Version.git_describe}-#{self.release}yakkety_amd64.deb", "repo_config" => "../repo_configs/deb/pl-#{self.project}-#{self.ref}-yakkety.list"},
+        {"name" => "windows-2012-x64", "artifact" => "./puppet/windows/#{self.project}-#{Pkg::Util::Version.git_describe}-x64.msi"},
+        {"name" => "windows-2012-x86", "artifact" => "./puppet/windows/#{self.project}-#{Pkg::Util::Version.git_describe}-x86.msi"} ]
```
underscorgan: rubocop is not happy about array formatting

lib/packaging/config.rb
underscorgan: What's the expected lifespan for this? Is this just as a stop-gap until the data service exists? I don't love the hardcoding but if it's actually going to be short lived I have less of a strong opinion.

mwaggett: yeah, this is meant to be temporary while we get the data service up and running. I agree that hardcoding isn't great, but it seemed like the simplest thing to do for something that's gonna get thrown out anyway.

shrug: The lifespan is (hopefully) pretty short, but it depends on how we are able to execute with the data service. 

The short version is that we either wait for the data service, and change beaker for the new foss repos, and then again when we add the data service, or add this here, then we only have to change beaker once, and we generate these values from the data service as soon as we can. I'm not in love with either option, this just seemed least bad.

melissa: we should be able to generate this. If the packages are built with the packaging repo, we should be able to extrapolate the data from mock.rake, where we define those paths. If it's from vanagon, we should theoretically have all the packages staged in the `output` directory. We can then query them, and copy the paths from that data. But I did forget about the metadata service. Hopefully that happens quickly, but hard coding, even temporary hard coding, makes me feel all gross and squirmy.


With the above in mind, though, this is going to cause issues. This will only work for the new puppet5 world, where we move things to the new path layout. What happens when we need to test puppet4 world things like new releases for the LTS? Also, do the beaker folks know this is limited to only puppet-agent? Are they expecting it to work for all projects and branches? When this is merged in, everything that ships to builds.delivery.puppetlabs.net will have this data in the yaml file, so beaker will have to know when it can trust it and when it can't...

lib/packaging/config.rb
underscorgan: Looking at the hash you linked this doesn't seem to be working. The puppet-agent yaml is using all the new output despite the fact that it's apt/yum repo names are PC1

mwaggett: oh, I might've linked the wrong one, one sec.

underscorgan: It also looks like the packages for puppetserver aren't right, http://builds.puppetlabs.lan/puppetserver/5.0.0.master.SNAPSHOT.2017.04.19T1053/artifacts/el/6/PC1/x86_64/puppetserver-5.0.0.master.SNAPSHOT.2017.04.19T1053-1.el6.x86_64.rpm doesn't exist but packages are actually at http://builds.puppetlabs.lan/puppetserver/5.0.0.master.SNAPSHOT.2017.04.19T1053/artifacts/el/6/PC1/x86_64/puppetserver-5.0.0.master-0.1SNAPSHOT.2017.04.19T1053.el6.noarch.rpm

Your puppetserver build didn't end up with repo configs (maybe somethign else funny happened?), but it looks like the ones in the hash are right based on other builds.

underscorgan:
@melissa why were you thinking this would only work for p-a?

melissa:
sorry, not for only p-a. It'll only work for new puppet5 package paths, so nothing old-style or pe style.

lib/packaging/config.rb
```diff
@@ -46,11 +46,84 @@ def config_from_hash(data = {})
...
+      end
+
+      ##
       # Load a yaml file and use its contents to set the values for Pkg::Config
       # class instance variables
       #
       def config_from_yaml(file)
         build_data = Pkg::Util::Serialization.load_yaml(file)
+        if self.apt_repo_name == 'PC1'
+          build_data[:platform_data] = platform_data_pc1
```
underscorgan: @melissa I think this should work for PC1, but yeah it won't work for PE packages where you end up with paths like http://builds.puppetlabs.lan/pe-puppet-enterprise-release/2017.3.0.0/artifacts/el/7/products/ and http://builds.puppetlabs.lan/pe-puppet-enterprise-release/2017.3.0.0/artifacts/deb/trusty/

lib/packaging/config.rb
mwaggett: @underscorgan Here's the right puppet-agent yaml: http://builds.delivery.puppetlabs.net/puppet-agent/5e98e39adfff2dfde1d86ad7abd17d9cbe03b5e8/artifacts/5e98e39adfff2dfde1d86ad7abd17d9cbe03b5e8.yaml

Not sure what happened with the server build.. that process is still a little confusing to me. In the meantime, I'll work on fixing that `0.1` thing that apparently lands in the artifact name..

lib/packaging/config.rb
mwaggett: Does it need to?
@kevpl ^ Are you hoping to also use this for PE packages?

mwaggett:
Closing in favor of https://github.com/puppetlabs/packaging/pull/641

PR #641, opened 2017-04-28
(RE-8707) Add `platform_data`

mwaggett:
This PR adds a `platform_data` hash that maps each platform to where its
artifact and repo_config (if applicable) can be found. This info will get added
to the <ref>.yaml file that gets generated when a package is built, allowing
beaker to consume this info without having to figure out the mapping itself.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -69,6 +69,40 @@ module Pkg
       end
 
       ##
+      # For each platform we ship for, find paths to its artifact and repo_config (if applicable).
+      # This is to be consumed by beaker and later replaced with our metadata service.
+      #
+      def platform_data
+        if self.project && self.ref
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -s \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/ -mindepth 2 -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
+            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            package_format = Pkg::Util::Platform.get_attribute(tag, :package_format)
+            case package_format
+            when 'deb'
+              artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and (e.include?("all") || e.include?("#{arch}.deb")) }
+              repo_config = "./repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Util::Platform.get_attribute(tag, :codename)}.list" if artifact
+            when 'rpm'
+              artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
+              repo_config = "./repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
+            when 'swix', 'svr4', 'ips', 'dmg', 'msi'
+              artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
+            else
+              fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
+            end
+            data[tag] = { :artifact => artifact,
+                          :repo_config => repo_config,
+                        } if artifact
+          end
+          data
+        end
+      end
+
+      ##
       # Return a hash of all build parameters and their values, nil if unassigned.
       #
       def config_to_hash
@@ -76,6 +110,7 @@ module Pkg
         Pkg::Params::BUILD_PARAMS.each do |param|
           data.store(param, self.instance_variable_get("@#{param}"))
         end
+        data.store(:platform_data, platform_data)
         data
       end
 
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -505,11 +505,6 @@ namespace :pl do
       project_basedir = "#{Pkg::Config.jenkins_repo_path}/#{Pkg::Config.project}/#{Pkg::Config.ref}"
       artifact_dir = "#{project_basedir}/#{target}"
 
-      # In order to get a snapshot of what this build looked like at the time
-      # of shipping, we also generate and ship the params file
-      #
-      Pkg::Config.config_to_yaml(local_dir)
-
       # For EZBake builds, we also want to include the ezbake.manifest file to
       # get a snapshot of this build and all dependencies. We eventually will
       # create a yaml version of this file, but until that point we want to
@@ -585,8 +580,16 @@ namespace :pl do
         Pkg::Util::Net.rsync_to("#{local_dir}/", Pkg::Config.distribution_server, "#{artifact_dir}/", extra_flags: ["--ignore-existing", "--exclude repo_configs"])
       end
 
+      # In order to get a snapshot of what this build looked like at the time
+      # of shipping, we also generate and ship the params file
+      #
+      Pkg::Config.config_to_yaml(local_dir)
+      Pkg::Util::Execution.retry_on_fail(:times => 3) do
+        Pkg::Util::Net.rsync_to("#{local_dir}/#{Pkg::Config.ref}.yaml", Pkg::Config.distribution_server, "#{artifact_dir}/", extra_flags: ["--exclude repo_configs"])
+      end
+
       # If we just shipped a tagged version, we want to make it immutable
-      files = Dir.glob("#{local_dir}/**/*").select { |f| File.file?(f) }.map do |file|
+      files = Dir.glob("#{local_dir}/**/*").select { |f| File.file?(f) and !f.include? "#{Pkg::Config.ref}.yaml" }.map do |file|
         "#{artifact_dir}/#{file.sub(/^#{local_dir}\//, '')}"
       end
 
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -20,14 +20,12 @@ module Pkg
       },
 
       'debian' => {
-        '6' => { :codename => 'squeeze', :architectures => ['i386', 'amd64'], :repo => true, :package_format => 'deb', },
         '7' => { :codename => 'wheezy', :architectures  => ['i386', 'amd64'], :repo => true, :package_format => 'deb', },
         '8' => { :codename => 'jessie', :architectures  => ['i386', 'amd64', 'powerpc'], :repo => true, :package_format => 'deb', },
         '9' => { :codename => 'stretch', :architectures  => ['i386', 'amd64'], :repo => true, :package_format => 'deb', },
       },
 
       'el' => {
-        '4' => { :architectures => ['i386', 'x86_64'], :repo => false, :package_format => 'rpm', :signature_format => 'v3', },
         '5' => { :architectures => ['i386', 'x86_64'], :repo => true, :package_format => 'rpm', :signature_format => 'v3', },
         '6' => { :architectures => ['i386', 'x86_64', 's390x'], :repo => true, :package_format => 'rpm', :signature_format => 'v4', },
         '7' => { :architectures => ['x86_64', 's390x'], :repo => true, :package_format => 'rpm', :signature_format => 'v4', },
@@ -38,9 +36,6 @@ module Pkg
       },
 
       'fedora' => {
-        'f21' => { :architectures => ['i386', 'x86_64'], :repo => true, :package_format => 'rpm', :signature_format => 'v4', },
-        'f22' => { :architectures => ['i386', 'x86_64'], :repo => true, :package_format => 'rpm', :signature_format => 'v4', },
-        'f23' => { :architectures => ['i386', 'x86_64'], :repo => true, :package_format => 'rpm', :signature_format => 'v4', },
         'f24' => { :architectures => ['i386', 'x86_64'], :repo => true, :package_format => 'rpm', :signature_format => 'v4', },
         'f25' => { :architectures => ['i386', 'x86_64'], :repo => true, :package_format => 'rpm', :signature_format => 'v4', },
       },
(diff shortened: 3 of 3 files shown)


mwaggett:
puppetserver: http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.05.04T1636/artifacts/
puppet-agent: http://builds.delivery.puppetlabs.net/puppet-agent/ac700b4d7e70783701e13079e0c7f3a635f3fe9a/artifacts/

.gitignore
```diff
@@ -1,3 +1,4 @@
 .bundle
 .idea
 Gemfile.lock
+*.swp
```
melissa: You shouldn't need this here. When there's a file with this ending, it means you have the file open somewhere.

mwaggett: I know, but I'm often committing while I have another file open and it kept being added to the commit. It was bothering me. But I can take this out if you don't think it belongs in the repo.

lib/packaging/config/params.rb
```diff
@@ -125,6 +125,7 @@ module Pkg::Params
                   :pe_platforms,
                   :pe_version,
                   :pg_major_version,
+                  :platform_data,
```
melissa: Are you actually using this anywhere? It could be accessed as `Pkg::Config.platform_data`

tasks/ship.rake
```diff
@@ -585,8 +580,16 @@ namespace :pl do
...
+      # of shipping, we also generate and ship the params file
+      #
+      Pkg::Config.config_to_yaml(local_dir)
+      Pkg::Util::Execution.retry_on_fail(:times => 3) do
+        Pkg::Util::Net.rsync_to("#{local_dir}/#{Pkg::Config.ref}.yaml", Pkg::Config.distribution_server, "#{artifact_dir}/", extra_flags: ["--exclude repo_configs"])
+      end
+
       # If we just shipped a tagged version, we want to make it immutable
-      files = Dir.glob("#{local_dir}/**/*").select { |f| File.file?(f) }.map do |file|
+      files = Dir.glob("#{local_dir}/**/*").select { |f| File.file?(f) and !f.include? "yaml" }.map do |file|
```
melissa: should we make `!f.include? "yaml"` more specific to our yaml file? EZbake also produces a yaml file, which we probably don't want to overwrite all the time

lib/packaging/config/params.rb
mwaggett: I think this would be fine without this, but I left it in for consistency with all the other things that end up in that yaml file, even though we're setting it a different way.

lib/packaging/config.rb
```diff
@@ -69,13 +69,43 @@ def config(args = { :target => nil, :format => :hash })
       end
 
       ##
+      # For each platform we ship for, find paths to its artifact and repo_config (if applicable).
+      # This is to be consumed by beaker and later replaced with our metadata service.
+      #
+      def platform_data
+        if self.project && self.ref
```
melissa: Is this just checking to make sure `project` and `ref` are set? Those should be safe assumptions. Did you run into issues where they weren't set

mwaggett: yes

tasks/ship.rake
mwaggett: oh yeah, I was gonna do that but forgot. Will do now.

lib/packaging/config.rb
```diff
@@ -69,13 +69,43 @@ def config(args = { :target => nil, :format => :hash })
...
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -d \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/*/ -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
```
melissa: what happens if we've shipped packages for platforms that we don't have entries for in `Pkg::Util::Platform.platform_tags`? This is not unheard of, where we have vanagon building packages before the packaging repo knows about them.

melissa: Could it make more sense to iterate over the files that you find that aren't in `repo_configs` and interpolate the platform tag from there?

mwaggett: They won't get included. To me, this seems like a better option than the opposite - grabbing the artifacts and then trying to figure out what platforms they're for. I think we just need to be better about adding platforms in all the places until we get the metadata service going.

mwaggett: I think it would involve a lot more grossness to figure out the platform tags from the artifact. especially because the way I'm grabbing artifacts is just grabbing all the files and there are somethings in there that don't map to a platform tag, like the SRPMS for example.

lib/packaging/config.rb
```diff
@@ -69,13 +69,43 @@ def config(args = { :target => nil, :format => :hash })
...
+            if artifact
+              if Pkg::Util::Platform.get_attribute(tag, :package_format) == 'rpm'
+                repo_config = "./repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo"
+              elsif Pkg::Util::Platform.get_attribute(tag, :package_format) == 'deb'
+                repo_config = "./repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Util::Platform.get_attribute(tag, :codename)}.list"
+              end
+            end
+            data[tag] = { :artifact => artifact,
+                          :repo_config => repo_config,
+                        }
```
underscorgan: I feel like it might be good to only add the tag if at least one of artifact/repo_config is set, otherwise there are a bunch of empty tags. I think it's reasonable for anything consuming this to handle missing tags.

.gitignore
melissa: It doesn't belong in the repo

lib/packaging/config.rb
melissa: Yes to the first question or yes to the second question where I forgot the question mark? If the second, can you elaborate on the issues you ran into?  Again, I believe these should be safe assumptions to make. All things we ship should have a project and a ref.

lib/packaging/config/params.rb
melissa: This implies that you'd be able to alter or manipulate this variable in some way outside of the code you've written for it. We're not doing that. It's an unnecessary addition

lib/packaging/config.rb
mwaggett: They have a project and a ref eventually, but there's a point where `config_to_yaml` is called (which calls `config_to_hash`, which calls `platform_data`) before those have been determined.

lib/packaging/config.rb
```diff
@@ -69,13 +69,43 @@ def config(args = { :target => nil, :format => :hash })
...
+      def platform_data
+        if self.project && self.ref
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -d \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/*/ -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
+            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include? arch }
```
underscorgan: This isn't working for noarch debian packages. For rpms, you end up with the arch in the path even for noarch packages (like, http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.05.04T1636/artifacts/el/7/PC1/x86_64/). But, the actual artifact for noarch deb things is something like `http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.05.04T1636/artifacts/deb/xenial/PC1/puppetserver_5.0.0.master-0.1SNAPSHOT.2017.05.04T1636puppetlabs1_all.deb` which isn't going to match the arch. It was picking up the `.changes` file in the server deb which isn't what you want

melissa: Debs don't always include arch. The path to the package doesn't include the architecture, and the package name only includes the architecture when it's an arch dependent package. Otherwise, it just has `all` in place of the architecture. For example, puppetdb: `http://builds.puppetlabs.lan/puppetdb/4.4.0/artifacts/deb/trusty/PC1/puppetdb_4.4.0-1puppetlabs1_all.deb`

lib/packaging/config/params.rb
mwaggett: okay, I'll remove it.

lib/packaging/config.rb
mwaggett: I feel like it's equally reasonable for anything consuming this to skip over platforms with missing artifacts/repo_configs. I can change that if you think it's better, just not sure what's best.

lib/packaging/config.rb
mwaggett: ah you're right, will fix

mwaggett: ahhh will fix

lib/packaging/config.rb
```diff
@@ -69,13 +69,43 @@ def config(args = { :target => nil, :format => :hash })
...
+      def platform_data
+        if self.project && self.ref
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -d \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/*/ -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
+            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and (e.include?("all") || e.include?(arch)) }
```
underscorgan: Have you run with this? I'm curious what's going to happen for deb noarches where there's a .changes for i386 but you really want the 'all' .deb for both x86_64 and i386

mwaggett: @underscorgan I'm building right now, so we'll see. I'm hoping it checks the 'all' first and just goes with that.

mwaggett: oh wait, but it's gonna go through files one by one and if it checks the .changes file first, it'll grab that.. hmm. I'll keep thinking.

mwaggett: I did work though..

underscorgan: that feels fragile.

mwaggett: I know, gonna change it.

underscorgan:
@mwaggett do you have p-a and puppetserver output with the changes? I like looking at the files :)

mwaggett:
@underscorgan Working on it. Still not quite cleaned up.

underscorgan:
cool cool

mwaggett:
updated puppet-agent yaml: http://builds.delivery.puppetlabs.net/puppet-agent/ac700b4d7e70783701e13079e0c7f3a635f3fe9a/artifacts/

updated puppetserver yaml: http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.05.05T1637/artifacts/

lib/packaging/config.rb
```diff
@@ -69,13 +69,42 @@ def config(args = { :target => nil, :format => :hash })
...
+      # This is to be consumed by beaker and later replaced with our metadata service.
+      #
+      def platform_data
+        if self.project && self.ref
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -s \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/*/ -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
```
melissa: Since we're doing this, can you open a PR to winston to update the hash in this repo with the new platform? Or at least make sure that task is there? (I can't remember if it is or not)

lib/packaging/config.rb
melissa: huh, strange

lib/packaging/config.rb
```diff
@@ -69,13 +69,42 @@ def config(args = { :target => nil, :format => :hash })
...
+            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            if Pkg::Util::Platform.get_attribute(tag, :package_format) == 'deb'
+              artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and (e.include?("all") || e.include?("#{arch}.deb")) }
+              repo_config = "./repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Util::Platform.get_attribute(tag, :codename)}.list" if artifact
+            else
```
melissa: Make this an elseif rpm, then make the else error out with a good error message so we know the platform hash needs updating or something

lib/packaging/config.rb
mwaggett: @melissa It's in there:
https://github.com/puppetlabs/winston/blob/master/tasks/platform-addition.rake#L156
https://github.com/puppetlabs/winston/blob/master/tasks/platform-removal.rake#L76

lib/packaging/config.rb
mwaggett: I can do something like that, but this needs to include rpms, dmgs, msis, etc. as well.

lib/packaging/config.rb
```diff
@@ -69,13 +69,42 @@ def config(args = { :target => nil, :format => :hash })
...
+              artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
```
melissa: Where does beaker want this path to start? I was under the impression it should start in the directory where the yaml file lives, (so, the artifacts directory). Then the artifacts path would be `el/7/...` and the repo config path would be `../repo_configs/el/7/...`. I see they both are now `./artifacts` and `./repo_configs`. Did you okay that with the beaker team?

lib/packaging/config.rb
melissa: The point still stands. We should bail with a reasonable error message when we're dealing with a package type we don't know how to deal with. I'd also suggest using a case statement.

Speaking of different package types, is there something we can do to check if there are any unreported packages? This wouldn't be a blocker, it would just be a nice to have, maybe in the next iteration

lib/packaging/config.rb
melissa: saweet!

lib/packaging/config.rb
mwaggett: I didn't, but this was easier for me and doesn't seem like it would be especially difficult for beaker to just know to start in the <project>/<ref> directory.

melissa: Make sure they're on the same page

lib/packaging/config.rb
mwaggett: Do you mean artifacts we found but didn't associate with a platform? Right now, I'm just looking at all files in a subdirectory under artifacts, which includes things like the .changes file. so I think that might be a bit of work and idk if it's worth it right now.

lib/packaging/config.rb
mwaggett: They're cool with this.

lib/packaging/config.rb
mwaggett: @melissa updated

PR #644, opened 2017-05-10
(maint) Clone build defaults instead of using curl

demophoon:
Using curl to download yaml files from raw.githubusercontent.com feels
wrong. We should instead use git to clone the repo so that it makes
local testing easier.


tasks/fetch.rake
```diff
@@ -35,31 +32,31 @@ namespace :pl do
...
-          if url == team_data_url
-            fail "Could not load team extras data from #{url}. This should not normally happen"
-          else
-            puts "No build data file for #{Pkg::Config.project}, skipping load of extra build data."
+
+    begin
+      temp_build_data_dir = Pkg::Util::File.mktemp
+      %x(git clone #{data_repo} #{temp_build_data_dir})
+      status = $?.exitstatus
+      case status
```
haus: weird to do a case with 2 branches. i think `if $?.success?` might be better than `case status`?

haus: same applies for inner loop

haus:
:woman_shrugging:

tasks/fetch.rake
```diff
@@ -35,31 +32,27 @@ namespace :pl do
...
+      temp_build_data_dir = Pkg::Util::File.mktemp
+      %x(git clone #{data_repo} #{temp_build_data_dir})
+      if $?.success?
+        Dir.chdir(temp_build_data_dir) do
+          [team_data_branch, project_data_branch].each do |branch|
+            %x(git checkout #{branch})
+            if $?.success?
+              Pkg::Util::RakeUtils.invoke_task("pl:load_extras", temp_build_data_dir)
+            else
+              puts "Unable to load build_defaults from branch '#{branch}' in '#{data_repo}'. Skipping."
```
haus: puts vs warn. no idea what the convention is/should be for that in the packaging repo

demophoon: :man_shrugging:

shrug: I think "warn" - turns out there are places (in tests) where some things are depending on stdout

stahnma:
I assume curling is faster though. What are the performance implications?

demophoon:
Curling is probably faster, this was just something I came up with to run against a local environment. At least it had been working for me a lot better than pushing changes up to the upstream repo. :man_shrugging:

tasks/fetch.rake
```diff
@@ -35,31 +32,29 @@ namespace :pl do
...
-          Pkg::Util::RakeUtils.invoke_task("pl:load_extras", tempdir)
-        when 22
-          if url == team_data_url
-            fail "Could not load team extras data from #{url}. This should not normally happen"
-          else
-            puts "No build data file for #{Pkg::Config.project}, skipping load of extra build data."
+
+    begin
+      temp_build_data_dir = Pkg::Util::File.mktemp
+      %x(git clone #{data_repo} #{temp_build_data_dir})
```
nwolfe: Regarding performance....we should set the clone depth here so we don't get all the git history; we just need the current state of things.

shrug:
@demophoon Take a look at the clone depth and possibly retarget at the 1.0 branch?

demophoon:
@shrug Depth change added and retargeted at 1.0.x.

shrug:
@nwolfe Look ok to you with the depth change?

PR #645, opened 2017-05-16
(RE-8843) Don't fail if ext/project_data.yaml doesn't exist

mwaggett:
This commit adjusts the loading of default configs to move on if project_data.yaml isn't found. Load will only fail if both project_data.yaml and build_defaults.yaml cannot be found.

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -321,8 +321,8 @@ module Pkg::Params
   DEPRECATIONS = [{ :var => :gem_devel_dependencies, :message => "
     DEPRECATED, 9-Nov-2013: 'gem_devel_dependencies' has been replaced with
     'gem_development_dependencies.' Please update this field in your
-    project_data.yaml" },
+    build_defaults.yaml or project_data.yaml" },
                   { :var => :gpg_name, :message => "
     DEPRECATED, 29-Jul-2014: 'gpg_name' has been replaced with 'gpg_key'.
-                   Please update this field in your project_data.yaml" }]
+                   Please update this field in your build_defaults.yaml" }]
 end
--- a/spec/lib/packaging/config_spec.rb
+++ b/spec/lib/packaging/config_spec.rb
@@ -313,6 +313,13 @@ describe "Pkg::Config" do
   end
 
   describe "#load_default_configs" do
+    before(:each) do
+      @project_root = double('project_root')
+      Pkg::Config.project_root = @project_root
+      @test_project_data = File.join(Pkg::Config.project_root, 'ext', 'project_data.yaml')
+      @test_build_defaults = File.join(Pkg::Config.project_root, 'ext', 'build_defaults.yaml')
+    end
+
     around do |example|
       orig = Pkg::Config.project_root
       example.run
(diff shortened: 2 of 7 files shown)


mwaggett:
puppet-agent build *without* project_data.yaml: http://builds.delivery.puppetlabs.net/puppet-agent/5d63931ee3ac07b7d9f9368dec618278644099bc/artifacts/
puppet-agent build *with* project_data.yaml: http://builds.delivery.puppetlabs.net/puppet-agent/a650e7a4c1d0a9ea68db72bbb12f16c2562e1f07/artifacts/
puppetserver build *without* project_data.yaml: http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.05.16T1033/artifacts/
puppetserver build *with* project_data.yaml: http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.05.16T1109/artifacts/

lib/packaging/config/params.rb
```diff
@@ -321,8 +321,8 @@ module Pkg::Params
   DEPRECATIONS = [{ :var => :gem_devel_dependencies, :message => "
     DEPRECATED, 9-Nov-2013: 'gem_devel_dependencies' has been replaced with
     'gem_development_dependencies.' Please update this field in your
-    project_data.yaml" },
+    build_defaults.yaml" },
```
melissa: This info probably still lives in project_data for most projects that set it, so you'll want to keep the original depreciation warning

lib/packaging/config/params.rb
```diff
@@ -321,8 +321,8 @@ module Pkg::Params
...
                   { :var => :gpg_name, :message => "
     DEPRECATED, 29-Jul-2014: 'gpg_name' has been replaced with 'gpg_key'.
-                   Please update this field in your project_data.yaml" }]
+                   Please update this field in your build_defaults.yaml" }]
```
melissa: This is good though, since we've only historically set gpg_key in build_defaults. Looks like some old copy pasta

lib/packaging/config/params.rb
mwaggett: I could say `build_defaults.yaml or project_data.yaml` ? Or just`project_data.yaml` if you think that's better.

spec/lib/packaging/config_spec.rb
```diff
@@ -330,6 +330,17 @@
...
+    context "given ext/build_defaults.yaml is readable but ext/project_data.yaml is not" do
+      it "should try to load build_defaults.yaml but not project_data.yaml" do
+        Pkg::Config.project_root = File.join(FIXTURES, 'config')
+        test_project_data = nil
+        test_build_defaults = File.join(FIXTURES, 'config', 'ext', 'build_defaults.yaml')
+        expect(Pkg::Config).to_not receive(:config_from_yaml).with(test_project_data)
+        expect(Pkg::Config).to receive(:config_from_yaml).with(test_build_defaults)
+        Pkg::Config.load_default_configs
+      end
+    end
```
melissa: You should add in more tests here

What happens if ext/project_data.yaml is readable, but ext/build_defaults isn't? At this point, that should be a fail case

melissa:
It looks like as is, this PR fails if it can't load either build_defaults or project_data, but it doesn't distinguish between the two. So, if either exist, that's fine. I can't decide how I feel about that. I was originally thinking we'd hard fail if build_defaults didn't exist, and would load project_data if it was there but didn't care if it wasn't. The thought being that we normally have build_defaults in projects, but we don't always have project_data. It's probably fine if we have at least one of those files though. Like I said, I can't decide. This is probably fine though

mwaggett:
@melissa yeah, I thought about that, but wasn't sure either. Since they usually have the same info, I was thinking it would be fine to have either, as long as things like `project`, `ref`, etc. are set somewhere. But I also kind of agree that it feels a little icky to not be opinionated about which we want. Maybe worth some input from @underscorgan and @shrug ?

underscorgan:
I think we should require build_defaults but not project_data. tbh I'm not sure what the original thinking was around separating the files but I think we should be opinionated here. :woman_shrugging:

mwaggett:
That revert got pulled in when I rebased, but I'm keeping it here since it should probably get merged into this branch anyway.

melissa:
I'm 👍, but I agree the tests aught to be updated. That'd also be a super good learning exercise!

PR #646, opened 2017-05-16
Ticket/1.0.x/re 8836 final versioning

melissa:


lib/packaging/rpm/version.rb
```diff
@@ -0,0 +1,18 @@
+module Pkg::Rpm::Version
+  class << self
+
+    # Utility method to return the dist method if this is a redhat box. We use this
+    # in rpm packaging to define a dist macro, and we use it in the pl:fetch task
+    # to disable ssl checking for redhat 5 because it has a certs bundle so old by
+    # default that it's useless for our purposes.
+    def el_version
+      if File.exists?('/etc/fedora-release')
```
shrug: what about sles or aix? I guess we're not using this on aix currently though...

melissa: I have no idea...

This currently is used in `pl:fetch`
```
    touch "#{ENV['HOME']}/.packaging"
    if dist = Pkg::Util::Version.el_version
      if dist.to_i < 6
        flag = "-k"
      end
    end
```

and `package:rpm` (which doesn't get used anywhere anymore)
```
def build_rpm(buildarg = "-bs")
  Pkg::Util::Tool.check_tool('rpmbuild')
  workdir = prep_rpm_build_dir
  if dist = Pkg::Util::Version.el_version
    if dist.to_i < 6
      dist_string = "--define \"%dist .el#{dist}"
    end
  end
```

lib/packaging/util/git.rb
```diff
@@ -13,13 +16,29 @@ def git_commit_file(file, message = "changes")
...
+    def tag(version)
       fail unless Pkg::Util::Version.is_git_repo?
       stdout, _, _ = Pkg::Util::Execution.capture3("#{Pkg::Util::Tool::GIT} tag -s -u #{Pkg::Config.gpg_key} -m '#{version}' #{version}")
       stdout
     end
 
-    def git_bundle(treeish, appendix = Pkg::Util.rand_string, temp = Pkg::Util::File.mktemp)
+    # Check if we are currently working on a tagged commit.
+    def tagged?
+      return git_ref_type = "tag"
```
underscorgan: this should be `==`

melissa:
@underscorgan don't review too closely. It's definitely broken in more than one way

lib/packaging/deb/version.rb
```diff
@@ -0,0 +1,13 @@
+module Pkg::Deb::Version
+  class << self
+    attr_reader :debversion, :origversion
```
haus: Does `attr_reader` work on things that aren't instance vars?

melissa:
This got out of hand. Closing in favor of https://github.com/puppetlabs/packaging/pull/651

PR #648, opened 2017-05-17
(RE-8707) Add `platform_data` to <ref>.yaml

mwaggett:
This commit adds a `platform_data` hash that maps each platform to where its artifact and repo_config (if applicable) can be found. This info will get added to the <ref>.yaml file that gets generated when a package is built, allowing beaker to consume this info without having to figure out the mapping itself.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -69,6 +69,42 @@ module Pkg
       end
 
       ##
+      # For each platform we ship for, find paths to its artifact and repo_config (if applicable).
+      # This is to be consumed by beaker and later replaced with our metadata service.
+      #
+      def platform_data
+        if self.project && self.ref && Pkg::Util::Net.check_host_ssh([self.builds_server]).empty?
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -s \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/ -mindepth 2 -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
+            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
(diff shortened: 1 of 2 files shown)


lib/packaging/config.rb
```diff
@@ -69,13 +69,48 @@ def config(args = { :target => nil, :format => :hash })
       end
 
       ##
+      # For each platform we ship for, find paths to its artifact and repo_config (if applicable).
+      # This is to be consumed by beaker and later replaced with our metadata service.
+      #
+      def platform_data
+        if self.project && self.ref && Pkg::Util::Net.check_host_ssh([self.builds_server]).empty?
```
melissa: wanna add an else here? Something with a little output, like, purposefully skipping this step because of permissions or something? Nothing major, just like, hey, don't worry this is failing, it's fine

lib/packaging/config.rb
```diff
@@ -69,13 +69,50 @@ def config(args = { :target => nil, :format => :hash })
...
+            else
+              fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
+            end
+            data[tag] = { :artifact => artifact,
+                          :repo_config => repo_config,
+                        } if artifact
+          end
+          data
+        else
+          puts "Skipping platform_data collection for now; don't worry, we'll come back to it."
```
mwaggett: @melissa updated

melissa: So, we may not always come back to it. If this is run as a part of running `rake pl:mock` standalone on someone's laptop, we probably won't come back to it.

lib/packaging/config.rb
```diff
@@ -69,13 +69,50 @@ def config(args = { :target => nil, :format => :hash })
...
+            else
+              fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
+            end
+            data[tag] = { :artifact => artifact,
+                          :repo_config => repo_config,
+                        } if artifact
+          end
+          data
+        else
+          puts "Skipping platform_data collection, but don't worry about it."
```
mwaggett: @melissa ¯\_(ツ)_/¯ writing messages is hard

PR #652, opened 2017-05-22
(RE-8707) Add `platform_data` to <ref>.yaml

mwaggett:
This commit adds a `platform_data` hash that maps each platform to where its
artifact and repo_config (if applicable) can be found. This info will get added
to the <ref>.yaml file that gets generated when a package is built, allowing
beaker to consume this info without having to figure out the mapping itself.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -69,6 +69,43 @@ module Pkg
       end
 
       ##
+      # For each platform we ship for, find paths to its artifact and repo_config (if applicable).
+      # This is to be consumed by beaker and later replaced with our metadata service.
+      #
+      def platform_data
+        if self.project && self.ref && Pkg::Util::Net.check_host_ssh([self.builds_server]).empty?
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -s \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/ -mindepth 2 -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
+            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            package_format = Pkg::Util::Platform.get_attribute(tag, :package_format)
+            case package_format
+            when 'deb'
+              artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and (e.include?("all") || e.include?("#{arch}.deb")) }
+              repo_config = "./repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Util::Platform.get_attribute(tag, :codename)}.list" if artifact
+            when 'rpm'
+              artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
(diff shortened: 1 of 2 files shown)


mwaggett:
Only change is at the end of `platform_data` - changed `puts` to `warn` and added explicit returns.

This is the job that failed with my last change, now succeeding: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent_intn-van-sys-suite-puppet-agent-ad-hoc-vmpooler-ad-hoc/7/

mwaggett:
puppetserver build: http://builds.delivery.puppetlabs.net/puppetserver/5.0.0.master.SNAPSHOT.2017.05.22T1132/artifacts/

puppet-agent build: http://builds.delivery.puppetlabs.net/puppet-agent/ca7f9a91bdf33481a9fa2463c37a488ff1c3b5c5/artifacts/

shrug:
yep, let's merge first thing tomorrow

lib/packaging/config.rb
```diff
@@ -69,13 +69,51 @@ def config(args = { :target => nil, :format => :hash })
...
+      #
+      def platform_data
+        if self.project && self.ref && Pkg::Util::Net.check_host_ssh([self.builds_server]).empty?
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -s \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/ -mindepth 2 -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
+          artifacts = artifacts.split("\n")
+          data = {}
+          Pkg::Util::Platform.platform_tags.each do |tag|
+            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
```
haus: nit: i think this can just be `arch = Pkg::Util::Platform.parse_platform_tag(tag).last`

lib/packaging/config.rb
```diff
@@ -69,13 +69,51 @@ def config(args = { :target => nil, :format => :hash })
...
 
       ##
+      # For each platform we ship for, find paths to its artifact and repo_config (if applicable).
+      # This is to be consumed by beaker and later replaced with our metadata service.
+      #
+      def platform_data
+        if self.project && self.ref && Pkg::Util::Net.check_host_ssh([self.builds_server]).empty?
+          dir = "/opt/jenkins-builds/#{self.project}/#{self.ref}"
+          cmd = "if [ -s \"#{dir}/artifacts\" ]; then cd #{dir}; find ./artifacts/ -mindepth 2 -type f; fi"
+          artifacts, _ = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true)
```
haus: nit: can be `artifacts = Pkg::Util::Net.remote_ssh_cmd(self.builds_server, cmd, true).first`

lib/packaging/config.rb
mwaggett: I personally kind of like the _s, for explicitness reasons.. and I think @underscorgan feels the same?

haus: :man_shrugging: :woman_shrugging: 

i get the same info (that there is more info that is being discarded) from seeing `last` called on it, but I don't feel at all strongly here.

mwaggett: but you wouldn't know how much info.. ¯\_(ツ)_/¯

haus: sure. i feel like in the case where you were going to use either of the other elements you would still need to go look at `parse_platform_tag` to see what it returned.

mwaggett: true.

PR #659, opened 2017-06-02
(RE-8864) When determining the ref type we need a clean describe

underscorgan:
To determine ref-type we use `git cat-file -t`. If this is given a
describe that isn't a sha or a tag it will fail. So, add the ability to
have `describe` run just `git describe` instead of `git describe --tags
--dirty` (the default) so we can use that method for determining the ref
type.


underscorgan:
Want to add spec tests here

underscorgan:
`ref_type` is _really_ hard to meaningfully test since it does everything in a block and we want to be testing what's inside the block, so just testing that we can do what we need to do with describe.

PR #660, opened 2017-06-06
(RE-8842) Ship apt, rpm, and downloads to S3

danieldreier:
Add three additional rake tasks to ship all repositories to S3 in addition to their current deployment targets.

This isn't quite ready to merge - it needs input from people more familiar with this process and code base.

For context, the sync script (`/usr/local/bin/s3_repo_sync.sh`) these commands invoke is deployed to weth using puppet, and handles both S3 upload and CDN cache clearing.

For the `:deploy_downloads_to_s3` task, I'm not sure which `Pkg::Config` setting to use. Right now I'm referencing `Pkg::Config.apt_signing_server` - which I assume is weth. Is that assumption correct? Is there a better config setting I can use? Create a new one?

I'm also not sure how to test this without risking accidentally shipping for real.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -265,6 +265,21 @@ namespace :pl do
         end
       end
     end
+
+    desc "Update rsync servers from AWS S3"
+    task :update_rsync_from_s3 => 'pl:fetch' do
+      puts "Really update rsync download servers from AWS S3? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          ['downloadserver-rsync-prod-1.ops.puppetlabs.net', 'downloadserver-rsync-prod-2.ops.puppetlabs.net'].each do |destination_server|
+            ['apt', 'yum'].each do |repo|
+              command = "sudo aws s3 sync --region us-west-2 s3://#{repo}.puppetlabs.com /opt/repository/#{repo}/"
+              Pkg::Util::Net.remote_ssh_cmd(destination_server, command)
+            end
+          end
+        end
+      end
+    end
   end
 
   desc "Ship built gem to rubygems.org, internal Gem mirror, and public file server"
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -215,6 +215,18 @@ namespace :pl do
       end
     end
 
+    desc "Copy signed deb repos from #{Pkg::Config.apt_signing_server} to AWS S3"
+    task :deploy_apt_repo_to_s3 => 'pl:fetch' do
+      puts "Really run S3 sync to deploy Debian repos from #{Pkg::Config.apt_signing_server} to AWS S3? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          destination_server = Pkg::Config.apt_signing_server
+          command = 'sudo /usr/local/bin/s3_repo_sync.sh apt.puppetlabs.com'
+          Pkg::Util::Net.remote_ssh_cmd(destination_server, command)
+        end
+      end
+    end
+
     desc "Copy rpm repos from #{Pkg::Config.yum_staging_server} to #{Pkg::Config.yum_host}"
     task :deploy_yum_repo => 'pl:fetch' do
       puts "Really run remote rsync to deploy yum repos from #{Pkg::Config.yum_staging_server} to #{Pkg::Config.yum_host}? [y,n]"
@@ -226,6 +238,30 @@ namespace :pl do
             Pkg::Config.yum_host,
             ENV['DRYRUN']
           )
+        end
+      end
+    end
+
+    desc "Copy signed RPM repos from #{Pkg::Config.yum_staging_server} to AWS S3"
+    task :deploy_yum_repo_to_s3 => 'pl:fetch' do
+      puts "Really run S3 sync to deploy RPM repos from #{Pkg::Config.yum_staging_server} to AWS S3? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          destination_server = Pkg::Config.apt_signing_server
+          command = 'sudo /usr/local/bin/s3_repo_sync.sh yum.puppetlabs.com'
+          Pkg::Util::Net.remote_ssh_cmd(destination_server, command)
+        end
+      end
+    end
+
+    desc "Sync downloads.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3"
+    task :deploy_downloads_to_s3 => 'pl:fetch' do
+      puts "Really run S3 sync to sync downloads.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          destination_server = Pkg::Config.apt_signing_server
+          command = 'sudo /usr/local/bin/s3_repo_sync.sh downloads.puppetlabs.com'
+          Pkg::Util::Net.remote_ssh_cmd(destination_server, command)
         end
       end
     end
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -265,6 +265,9 @@ namespace :pl do
         remote:deploy_swix_repo
         remote:deploy_msi_repo
         remote:deploy_tar_repo
+        remote:deploy_apt_repo_to_s3
+        remote:deploy_yum_repo_to_s3
+        remote:deploy_downloads_to_s3
       )
 
       if Pkg::Util.boolean_value(Pkg::Config.answer_override) && !Pkg::Config.foss_only
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -268,6 +268,7 @@ namespace :pl do
         remote:deploy_apt_repo_to_s3
         remote:deploy_yum_repo_to_s3
         remote:deploy_downloads_to_s3
+        remote:update_rsync_from_s3
       )
 
       if Pkg::Util.boolean_value(Pkg::Config.answer_override) && !Pkg::Config.foss_only


tasks/ship.rake
```diff
@@ -215,6 +215,18 @@ namespace :pl do
       end
     end
 
+    desc "Copy signed deb repos from #{Pkg::Config.apt_signing_server} to AWS S3"
+    task :deploy_apt_repo_to_s3 => 'pl:fetch' do
+      puts "Really run S3 sync to deploy Debian repos from #{Pkg::Config.apt_signing_server} to AWS S3? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          destination_server = Pkg::Config.apt_signing_server
+          command = '/usr/local/bin/s3_repo_sync.sh apt.puppetlabs.com'
```
underscorgan: So after thinking about this a little more I think we might need to have a few changes to the sync command. I think we're going to need to add something to the rsync for apt to ignore the freight directories, the current rsync just adds excludes when transferring from weth => burjis (https://github.com/puppetlabs/packaging/blob/master/lib/packaging/deb/repo.rb#L213-L214)

Also we might need some differences in link handling, though i don't know how that interacts with s3 (at least last time I used it s3 didn't symlink), but currently for debs we use https://github.com/puppetlabs/packaging/blob/master/lib/packaging/deb/repo.rb#L200-L201 whereas for rpm we have https://github.com/puppetlabs/packaging/blob/master/lib/packaging/rpm/repo.rb#L56-L57 to not have a ton of copies of everything for the el 5 vs 5.0 vs 5.1 etc.

danieldreier: I've changed the shell script on weth (`/usr/local/bin/s3_repo_sync.sh`, in the [aws_downloadserver branch](https://github.com/puppetlabs/puppetlabs-modules/blob/99cd37a834546ebdd2bc68ad7c236ea02944e21a/site/profile/files/delivery/s3_repo_sync.sh)) to handle these issues.

The script already uses rsync to make a local copy of the repos on weth to a staging location, and so I added the excludes you linked in repo.rb. 

Regarding symlinks, S3 lacks any concept of symlinks, so the `aws sync` command follows symlinks and uploads them as real files. That'll mean we waste disk space in S3 (unavoidable) and on the rsync servers (avoidable if we used a different copying mechanism) but I think that it's cheaper to just copy more files around than to optimize for maybe $10-20/mo of storage.

danieldreier:
@underscorgan I've updated `tasks/ship.rake` to use `Pkg::Config.staging_server` instead of `Pkg::Config.apt_signing_server`.

Can you either test this, or tell / show me how to test it? The things I'm not sure about are:

- It looks like this needs a config file but I wasn't able to find an example of one in the repository. Aside from installing gems etc, do I need to do anything (config files, etc) to set up an environment to run this in, or can I just run as-is? I'm hesitant to experiment since a mistake could mean shipping to production.
- Should it just be invoked as `DRYRUN=true rake pl:jenkins:uber_ship`? I see a few places where the `DRYRUN` environment variable is used.

underscorgan:
@danieldreier so testing changes to packaging can be a little bit finicky. Packaging can not be run stand-alone, and needs to be run from within a different project. I'd say DRY_RUN is used inconsistently, and packaging tasks working will depend on a bunch of local setup that I'm not sure we have fully documented, so it's probably easiest if I test this. With regards to testing, is it ok if I kick off the s3 sync or is that bad / essentially shipping to production at this point? Also happy to pair on testing if you want to see how it's done!

tasks/ship.rake
underscorgan: ok, cool. Looks good, thanks!

underscorgan:
@danieldreier so it looks like the aws credentials aren't being loaded properly. If I sudo to root then run `aws s3 sync /opt/repo-s3-stage/repositories/apt.puppetlabs.com s3://apt.puppetlabs.com` it works fine, but `sudo aws s3 sync /opt/repo-s3-stage/repositories/apt.puppetlabs.com s3://apt.puppetlabs.com` returns `fatal error: Unable to locate credentials`

danieldreier:
@underscorgan I've added an `export AWS_CONFIG_FILE=/root/.aws/credentials` line to the `s3_repo_sync.sh` script. When I ran it as `sudo /usr/local/bin/s3_repo_sync.sh` it worked.

underscorgan:
@danieldreier new errors

```
+ aws cloudfront create-invalidation --distribution-id E3UJXLQWC4T3QR --paths '/*'
AWS CLI support for this service is only available in a preview stage.

However, if you'd like to use the "aws cloudfront" commands with the
AWS CLI, you can enable this service by adding the following to your CLI
config file:

    [preview]
    cloudfront=true

or by running:

    aws configure set preview.cloudfront true
```

underscorgan:
Also, FYI, when I was testing the sync it ended up coinciding with a puppet run which caused weth to almost run out of swap. We might want to give that box some more RAM :)

danieldreier:
@underscorgan I added an `aws configure set preview.cloudfront true` line to the `/usr/local/bin/s3_repo_sync.sh` script. That should resolve the `AWS CLI support for this service is only available in a preview stage.` error. It adds 0.5 seconds but that's not a huge problem given how long the rest of the process takes.

Can you test again? I ran it using your user account on weth, because it had previously worked for me and I wanted to confirm it'd work for you, but I'd feel more comfortable if you ran through it as well since you've found several issues that got past me.

tasks/ship.rake
```diff
@@ -231,6 +243,30 @@ namespace :pl do
     end
   end
 
+  desc "Copy signed RPM repos from #{Pkg::Config.apt_signing_server} to AWS S3"
```
melissa: I thought I read a comment this was changed to `Pkg::Config.staging_server`. Was this changed back? Did it fail for some reason?

danieldreier: @melissa you're right, I just pushed a fix. For this particular one I used `Pkg::Config.yum_staging_server` to be consistent with the other RPM-related entries.

tasks/ship.rake
```diff
@@ -231,6 +243,30 @@ namespace :pl do
     end
   end
```
melissa: You want your additions to be before this `end`, that way they'll be under the `remote` task namespace

tasks/ship.rake
```diff
@@ -79,6 +79,8 @@ namespace :pl do
         end
       end
     end
+
+
```
melissa: I don't think you want these extra newlines

danieldreier:
@melissa it looks like when I updated the [s3 sync script](https://github.com/puppetlabs/puppetlabs-modules/blob/04a453a46f664e6738380f1757a8453ec4320852/site/profile/files/delivery/s3_repo_sync.sh), I omitted the trailing slash on the rsync source, so it copied the directory instead of directory contents. That's why it took so long for you to run - it was copying the whole repo. I updated it with the trailing slash, and also updated the s3 sync command itself to use --delete, which deletes stuff from the S3 bucket that we no longer have on disk on weth, so we don't have unmanaged old files. 

I'm re-running it on weth now. I'll update the PR when it's ready to be tested again.

danieldreier:
@melissa when you have time, I think this is ready to be tested again.

melissa:
There's a PA release going out tomorrow. How would you feel if I waited until then?

danieldreier:
@melissa that's fine, I'm realistically not going to switch over the download servers today anyway, since I won't be available tomorrow and it'd be reckless to do that right before a release when I can't support it.

melissa:
I'm getting errors when I run deploy_downloads_to_s3. But the first two, apt and yum, were successful.
```
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/dashboard/puppet-dashboard-1.0.4.tgz' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/activemq-info-provider-5.5.0-1.el5.noarch.rpm' because the file cannot
 be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/tanukiwrapper-3.5.9-1.el5.i386.rpm' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/tanukiwrapper-3.5.9-1.el5.x86_64.rpm' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/activemq-5.5.0-1.el6.noarch.rpm' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/activemq-5.5.0-1.el5.noarch.rpm' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/tanukiwrapper-3.5.9-1.el6.i686.rpm' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/tanukiwrapper-3.5.9-1.el6.x86_64.rpm' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/mcollective/activemq-info-provider-5.5.0-1.el6.noarch.rpm' because the file cannot
 be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/training/archive/current-pm-vm.zip.md5' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/training/archive/current-pm-vm.zip' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/training/archive/current-pm-vm-ovf.zip.md5' because the file cannot be read
ERROR:root:skipping '/opt/repo-s3-stage/repositories/downloads.puppetlabs.com/training/archive/current-pm-vm-ovf.zip' because the file cannot be read
```

tasks/ship.rake
```diff
@@ -229,6 +241,30 @@ namespace :pl do
         end
       end
     end
+
+    desc "Copy signed RPM repos from #{Pkg::Config.yum_staging_server} to AWS S3"
+    task :deploy_rpm_repo_to_s3 => 'pl:fetch' do
```
mwaggett: I feel like we say `yum_repo` instead of `rpm_repo`, for consistency with apt..

melissa:
This doesn't seem to be sufficient to clear the cache with the current setup. Am I missing something, or is that the case? I know apt.pl.com and yum.pl.com are both behind the CDN, but downloads.pl.com isn't. That being said, I was under the impression the first two rake tasks (`deploy_rpm_repo_to_s3` and `deploy_apt_repo_to_s3`) should do everything we need for the way things are currently set up.

danieldreier:
@melissa it (intentionally) does not clear the cache for our current setup, because our current setup uses a CloudFront distribution that won't exist after this way of doing things goes live. Right now we have a single CloudFront distribution in front of burji3 for both apt and yum, but once we have those in S3 we'll need to use different distributions for apt, yum, and downloads. I've been figuring it didn't make sense to add that functionality to this PR, because this PR being open is what's blocking me from decommissioning that particular CloudFront distribution.

melissa:
Ah, alright, that was my mistake then. I didn't quite understand that. Let me know when you've looked into those issues, and I'd be glad to merge this

danieldreier:
@melissa I fixed a few incorrect symlinks and deleted two others (`/opt/downloads/training/archive/current-pm-vm-ovf.zip` and `/opt/downloads/training/archive/current-pm-vm-ovf.zip.md5`) that didn't resolve. The errors you saw no longer happen. Since those symlinks have apparently been broken for a while with no complaints, I figured there was no harm in removing them.

underscorgan:
hm, one problem I'm realizing now is I don't think the edu/training VMs ship through weth, they currently ship directly to the burjis, so files under /opt/downloads/training are going to be out of date.

We'll need to resync weth to look like the burjis for probably all of /opt/downloads, and once the cutover has happened we can update the training/edu vm shipping automation to ship to weth instead of the burjis, these ship infrequently enough it's probably okay to wait to update that until we do the cutover, but should coordinate with @samuelson

danieldreier:
@melissa @underscorgan I added a new rake task here to run the AWS S3 sync from the new rsync servers. It's kind of terrible and has the servers hardcoded; do you have a better idea? Also, I coded this blind because I still don't have a local environment set up to run this.

Can one of you add something to the README or a confluence page to describe how to set up a local environment to run this? The only bit I'm really missing is that you both appear to have a valid config file on your laptops, and I don't know where to get that. Alternatively we could set up a small server whose only purpose is to run this, so each person didn't need to run it locally.

PR #663, opened 2017-06-09
(maint) Remove Ruby 1.9.3 and 2.0.0 from Travis test matrix

mwaggett:


.travis.yml
```diff
@@ -4,17 +4,8 @@ sudo: false
 notifications:
   email: false
 rvm:
-  - 1.9.3
-  - 2.0.0
   - 2.1.1
```
underscorgan: Can we add some more recent rubies while we're in here? maybe a 2.3 and a 2.4?

mwaggett: sure! Do we want to check rubocop on everything? Before we were just checking the latest version.

underscorgan: no, one rubocop is enough, those won't vary based on ruby version.

mwaggett: @underscorgan updated!

PR #666, opened 2017-06-20
(RE-8727) Add support to define a nonfinal gpg key

melissa:
This commit adds the ability for users to specify a nonfinal gpg key.
This is an alternative to the main gpg key and will be used to sign any
package versions that are flagged as nonfinal versions. The idea is that
this automation allows us to set up and maintain separate
nonfinal/beta/nightly/testing repos using our main shipping automation
(this repo). This will allow us to get a little bit closer to automating
the release process.

Note, the debian repo configs still specify the main gpg key as the
signing key. This is due to the fact that there was no easy way to
branch on keys for freight. The signing key is actually defined in the
freight config file that lives on the internal repo staging host. We
could add multiple of these config files for each key we sign with. But
it seemed more logical to simply continue signing our apt repos with the
main key.


spec/lib/packaging/util/gpg_spec.rb
```diff
@@ -1,16 +1,37 @@
...
+      expect(Pkg::Util::Gpg.key).to eq(gpg_key)
+    end
+
+    it 'uses the final key if the nonfinal key is not defined for a nonfinal version' do
+      allow(Pkg::Config).to receive(:version).and_return('1.0.0-658-gabc1234')
+      allow(Pkg::Config).to receive(:gpg_nonfinal_key).and_return(nil)
+      expect(Pkg::Util::Gpg.key).to eq(gpg_key)
+    end
+
+    it 'uses the nonfinal key if building a nonfinal version' do
```
melissa: yes
```
    # Determines if the version we are working with is or is not final
    #
    # The version here does not include the release version. Therefore, we
    # assume that any version that includes a `-\d+` was not built from a tag
    # and is a non-final version.
    # Examples:
    # Final
    #   - 5.0.0
    #   - 2016.5.6.7
    # Nonfinal
    #   - 4.99.0-22
    #   - 1.0.0-658-gabc1234
    #   - 5.0.0.master.SNAPSHOT.2017.05.16T1357
    #   - 5.9.7-rc4
    #   - 4.99.0-56-dirty
    #
    def final?(version = Pkg::Config.version)
      case version
      when /rc/
        false
      when /SNAPSHOT/
        false
      when /-dirty/
        false
      when /g[a-f0-9]{7}$/
        false
      when /^(\d+\.)+\d+-\d+$/
        false
      else
        true
      end
    end
```

melissa: Would you prefer if I added in a test for snapshot to this file?

shrug: No need, imo, assuming we test the final? method for it

melissa: yup
```
    final_versions = [
      '1.0.0',
      '2017.6.5.3',
      '0.6.8',
      '2068.532.6',
      '96.5'
    ]

    non_final_versions = [
      '4.99.0-22',
      '1.0.0-658-gabc1234',
      '5.0.0.master.SNAPSHOT.2017.05.16T1357',
      '5.9.7-rc4',
      '4.99.0-56-dirty'
    ]

    final_versions.each do |version|
      it "returns true when given #{version}" do
        allow(Pkg::Config).to receive(:version) { nil }
        expect(Pkg::Util::Version.final?(version)).to be true
      end
    end

    non_final_versions.each do |version|
      it "returns false when given #{version}" do
        allow(Pkg::Config).to receive(:version) { nil }
        expect(Pkg::Util::Version.final?(version)).to be false
      end
    end
```

demophoon:
This makes sense to me 👍

PR #668, opened 2017-06-22
(maint) Update training symlinks to match what actually exists now

underscorgan:


underscorgan:
@samuelson I think this matches the updates you had me doing to symlinks last week, but if you could verify that would be a+

samuelson:
Yes I think that looks good @underscorgan

PR #670, opened 2017-06-26
(RE-8953) Update automation to support creating symlinks for repos

underscorgan:
This will let us have 'puppet' which currently links to puppet5, but will be updated to puppet6 when that time comes.


underscorgan:
see also, https://github.com/puppetlabs/puppetlabs-modules/pull/7253

tasks/ship.rake
```diff
@@ -37,6 +58,39 @@ namespace :pl do
         __GPG_KEY__: Pkg::Util::Gpg.key
       }
 
+      # We want to run this for both final/nonfinal paths. This won't set up the links correctly with
+      # nonfinal builds since in that case repo_name gets set to nonfinal_repo_name, but it also won't
```
mwaggett: If you want to use the correct repo name, it's set here: https://github.com/puppetlabs/packaging/blob/1.0.x/lib/packaging/paths.rb#L51
so you can use Pkg::Paths.repo_name instead of Pkg::Config.repo_name

underscorgan: Hmm, actually, I think the issues I was referring to in that comment were from using Pkg::Paths.repo_name, I'll need to get my test env up again so I can check if using Pkg::Config.repo_name fixes that issue / update the comment.

underscorgan: Ok, yup, I want to be using `Pkg::Config.repo_name` and that fixed the oddities. Cleaned up the comment

tasks/ship.rake
```diff
@@ -37,6 +58,37 @@ namespace :pl do
         __GPG_KEY__: Pkg::Util::Gpg.key
       }
 
+      # We want to run this for both final/nonfinal paths.
```
mwaggett: Is there a reason this has comments throughout but the yum equivalent doesn't? I just like as much consistency and clarity as possible. 😊

underscorgan: Comments copied into the yum block.

tasks/ship.rake
```diff
@@ -14,6 +14,33 @@ namespace :pl do
...
+      { Pkg::Config.repo_name => Pkg::Config.repo_link_target, Pkg::Config.nonfinal_repo_name => Pkg::Config.nonfinal_repo_link_target }.each do |name, target|
+        if target
```
melissa: what if `name` isn't set, like if we don't have `Pkg::Config.nonfinal_repo_name` but we do have `Pkg::Config.nonfinal_repo_link_target` set for some reason

underscorgan: Hm, that doesn't seem too likely but probably best to add a check, that could end badly for sure

underscorgan:
@melissa updated to be `if name && target` and tested to make sure that did what I expected it to.

melissa:
@mwaggett this is a good example of complicated bash. Make sure you understand what is going on here. http://mywiki.wooledge.org/BashGuide/TestsAndConditionals is a good reference.

PR #673, opened 2017-06-28
(maint) Add override to make tagged but dirty trees final

underscorgan:
When testing automation changes or fixing last-minute packaging errors,
it can be desirable to check out a tagged version of a project and
modify build_defaults to use an alternate branch of packaging. This
allows you to pass the ALLOW_DIRTY_TREE environment variable to return
cause final? to return true if you're on a tagged build with a dirty
tree.

### DIFF
--- a/lib/packaging/util/version.rb
+++ b/lib/packaging/util/version.rb
@@ -98,12 +98,12 @@ module Pkg::Util::Version
         false
       when /SNAPSHOT/
         false
+      when /g[a-f0-9]{7}/
+        false
+      when /^(\d+\.)+\d+-\d+/
+        false
       when /-dirty/
-        false
-      when /g[a-f0-9]{7}$/
-        false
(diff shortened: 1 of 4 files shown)


lib/packaging/util/version.rb
```diff
@@ -98,12 +98,16 @@ def final?(version = Pkg::Config.version)
...
-      when /-dirty/
-        false
-      when /g[a-f0-9]{7}$/
+      when /g[a-f0-9]{7}/
         false
-      when /^(\d+\.)+\d+-\d+$/
+      when /^(\d+\.)+\d+-\d+/
         false
+      when /-dirty/
+        if Pkg::Config.allow_dirty_tree
```
mwaggett: this could just return `Pkg::Config.allow_dirty_tree`

underscorgan:
@mwaggett done.

PR #677, opened 2017-06-30
(RE-8943) Update shipping automation to deal with 'latest' symlinks

mwaggett:
This commit adds a function for finding the latest version of something and creating a 'latest' symlink to it. We do this for puppet-agent (and its components) for mac and windows. We now do this whenever we ship a dmg or msi.

### DIFF
--- a/lib/packaging/util/ship.rb
+++ b/lib/packaging/util/ship.rb
@@ -70,4 +70,23 @@ module Pkg::Util::Ship
       end
     end
   end
+
+  def create_latest_symlink(package_name, dir, platform_ext, excludes = [], arch = nil)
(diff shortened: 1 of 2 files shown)


mwaggett:
I tested the function in irb, replacing the `remote_ssh_command` part with `puts`, just to see what the resulting command would've been. Then, on weth-dev, I navigated to the windows and mac directories and ran the `ls -1 . . . tail -1` part of the command to see if the latest version was found. Everything worked as expected.

melissa:
This should be targeted at master, as it is not limited to the puppet5 change

mwaggett:
woops okay, I'm gonna close this one and open a new one because the retargeting feature kinda sucks,

PR #678, opened 2017-06-30
(RE-8943) Update shipping automation to deal with 'latest' symlinks

mwaggett:
This commit adds a function for finding the latest version of something and creating a 'latest' symlink to it. We do this for puppet-agent (and its components) for mac and windows. We now do this whenever we ship a dmg or msi.

### DIFF
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -280,6 +280,25 @@ module Pkg::Util::Net
       Pkg::Util::Net.remote_ssh_cmd(host, "sudo chattr +i #{files.join(" ")}")
     end
 
+    def remote_create_latest_symlink(package_name, dir, platform_ext, excludes = [], arch = nil)
+      cmd = "if [ -d '#{dir}' ] ; then "
+      cmd << "pushd #{dir} ; "
+      cmd << "ln -sf `\ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep #{package_name} "
+      if arch
+        cmd << "| grep #{arch} "
+        package_name << "-#{arch}"
+      end
+      excludes.each do |excl|
+        cmd << "| grep -v #{excl} "
+      end
+      cmd << "| sort --version-sort | tail -1` #{package_name}-latest.#{platform_ext} ; "
+      cmd << "popd ; "
+      cmd << "fi"
+
+      _, err = Pkg::Util::Net.remote_ssh_command(Pkg::Config.staging_server, cmd, true)
+      $stderr.puts err
+    end
+
     def escape_html(uri)
       require 'cgi'
       CGI.escapeHTML(uri)
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -280,16 +280,18 @@ module Pkg::Util::Net
       Pkg::Util::Net.remote_ssh_cmd(host, "sudo chattr +i #{files.join(" ")}")
     end
 
-    def remote_create_latest_symlink(package_name, dir, platform_ext, excludes = [], arch = nil)
+    def remote_create_latest_symlink(package_name, dir, platform_ext, options = {})
       cmd = "if [ -d '#{dir}' ] ; then "
       cmd << "pushd #{dir} ; "
       cmd << "ln -sf `\ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep #{package_name} "
-      if arch
-        cmd << "| grep #{arch} "
-        package_name << "-#{arch}"
+      if options[:arch]
+        cmd << "| grep #{options[:arch]} "
+        package_name << "-#{options[:arch]}"
       end
-      excludes.each do |excl|
-        cmd << "| grep -v #{excl} "
+      if options[:excludes]
+        options[:excludes].each do |excl|
+          cmd << "| grep -v #{excl} "
+        end
       end
       cmd << "| sort --version-sort | tail -1` #{package_name}-latest.#{platform_ext} ; "
       cmd << "popd ; "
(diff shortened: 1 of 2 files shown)


mwaggett:
I tested the function in irb, replacing the remote_ssh_command part with puts, just to see what the resulting command would've been. Then, on weth-dev, I navigated to the windows and mac directories and ran the `ls -1 . . . tail -1` part of the command to see if the latest version was found. Everything worked as expected.

tasks/ship.rake
```diff
@@ -367,6 +367,10 @@ namespace :pl do
         end
       end
     end
+    Pkg::Util::Net.remote_create_latest_symlink('puppet', '/opt/downloads/mac', 'dmg', ['agent', 'hiera'])
+    Pkg::Util::Net.remote_create_latest_symlink('hiera', '/opt/downloads/mac', 'dmg', ['puppet'])
+    Pkg::Util::Net.remote_create_latest_symlink('facter', '/opt/downloads/mac', 'dmg')
+    Pkg::Util::Net.remote_create_latest_symlink('puppet-agent', '/opt/downloads/mac', 'dmg')
```
melissa: There are no puppet-agent packages in /opt/downloads/mac. Those packages are all in subdirectories. How does this handle that?

mwaggett: @melissa It creates a link with no target. I was basing this all on the existing script (which you can see [here](https://tickets.puppetlabs.com/browse/RE-8943)), but I can just remove that line or adjust it to the appropriate directory if that's better.

melissa: It seems like we don't want links with no targets...

mwaggett: @melissa I agree. Should I just remove this line? I'm not sure why it was in the original script if puppet-agent doesn't live in that directory. Is there a different directory where those links should be made?

underscorgan: If anywhere, these links could make sense under the individual macos platforms (10.10, 10.11, 10.12, etc). It also seems like they might not be needed since it's been broken for forever. I think the windows symlinks mostly need to exist for beaker/testing.

tasks/ship.rake
```diff
@@ -367,6 +367,12 @@ namespace :pl do
         end
       end
     end
+    Pkg::Util::Net.remote_create_latest_symlink('puppet', '/opt/downloads/mac', 'dmg', ['agent', 'hiera'])
+    Pkg::Util::Net.remote_create_latest_symlink('hiera', '/opt/downloads/mac', 'dmg', ['puppet'])
+    Pkg::Util::Net.remote_create_latest_symlink('facter', '/opt/downloads/mac', 'dmg')
+    Pkg::Platforms.PLATFORM_INFO['osx'].keys.each do |version|
+      Pkg::Util::Net.remote_create_latest_symlink('puppet-agent', "/opt/downloads/mac/#{version}/#{Pkg::Config.yum_repo_name}/x86_64", 'dmg')
+    end
```
mwaggett: @underscorgan @melissa Thoughts on this update?

underscorgan: This looks like it will do the right thing, mostly just wondering if this will actually be useful :woman_shrugging:

mwaggett: yeah, I'm happy to take it out, but I figure better to have it and no one use it than remove and have people get sad?  ¯\_(ツ)_/¯

tasks/ship.rake
```diff
@@ -427,6 +433,10 @@ namespace :pl do
         end
       end
     end
+    Pkg::Util::Net.remote_create_latest_symlink('puppet', '/opt/downloads/windows', 'msi', ['agent', 'x64'])
+    Pkg::Util::Net.remote_create_latest_symlink('puppet', '/opt/downloads/windows', 'msi', ['agent'], 'x64')
```
underscorgan: Both of the puppet arches are x64

mwaggett: no, the first one is excluding x64 and the second is including it as an arch.

mwaggett: since some have x64 specified and others don't

melissa: hmm, this syntax is pretty confusing. Is there a way you could make it more user friendly for when we try to remember how this works in the future?

mwaggett: I agree, especially from here, but do you think it's still too confusing if you go back to the method definition?
My thoughts are to 1. add comments above the method definition, though that doesn't help *here*,
and/or 2. create optional flags for the excludes and arch, though that could make things messier.
Will definitely think on this more and very open to thoughts and suggestions. 😊

underscorgan: Does this work?
```
Pkg::Util::Net.remote_create_latest_symlink('puppet', '/opt/downloads/windows', 'msi', ['agent'], 'x86')
Pkg::Util::Net.remote_create_latest_symlink('puppet', '/opt/downloads/windows', 'msi', ['agent'], 'x64')
```

That's consistent with how it's being done below for the agent, and significantly less confusing I think.

mwaggett: @underscorgan puppet doesn't have a x86 variety - just x64 and plain. I did some refactoring that I think will make things clearer though.

lib/packaging/util/net.rb
```diff
@@ -280,6 +280,27 @@ def remote_set_immutable(host, files)
       Pkg::Util::Net.remote_ssh_cmd(host, "sudo chattr +i #{files.join(" ")}")
     end
 
+    def remote_create_latest_symlink(package_name, dir, platform_ext, options = {})
```
underscorgan: Can you add docs to this method? Is this reasonable to test?

mwaggett: @underscorgan Added documentation.

Last week we decided that it didn't really make sense to write a test for this method, since the behavior we want is the creation of a symlink on a remote server. We considered just testing that the command came back as expected, but it would just involve a lot of copy pasting and any time we updated the command, we'd have to update the test in the same way, which kind of defeats the purpose of having a test.

shrug:
Should this go to master or 1.0?

mwaggett:
@shrug We decided master, since this isn't related to puppet5.

shrug:
👍

PR #682, opened 2017-06-30
(WHO-451) rsync directly to rsync servers

danieldreier:
Ship apt and yum directly from weth to rsync servers.

Without this change we don't have a way to ship symlinks, which would blow up our users when we switch because they'd all download hundreds of gb extra data.

I know that hardcoding these servers is horrible. I'm willing to do it a different way but don't know where to put it in config files. I don't think we'll ever have more or fewer rsync servers so this seems fine, albeit gross.


tasks/ship.rake
```diff
@@ -266,6 +266,23 @@ namespace :pl do
       end
     end
 
+    desc "Sync yum and apt from #{Pkg::Config.staging_server} to rsync servers"
+    task :deploy_to_rsync_server => 'pl:fetch' do
+      puts "Really run rsync to sync apt and yum from #{Pkg::Config.staging_server} to rsync servers? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        destination_server = Pkg::Config.staging_server
```
mwaggett: I feel like this is a confusing variable name, since I'm imagining it as the server we're rsyncing to, rather than the destination of the rsync command. Since it's only used once, I feel like it would be fine to just have `Pkg::Config.staging_server` as an argument to the remote_ssh_cmd.

danieldreier: good idea, I've pushed a change to simplify it

tasks/ship.rake
```diff
@@ -266,6 +266,22 @@ namespace :pl do
       end
     end
 
+    desc "Sync yum and apt from #{Pkg::Config.staging_server} to rsync servers"
+    task :deploy_to_rsync_server => 'pl:fetch' do
+      puts "Really run rsync to sync apt and yum from #{Pkg::Config.staging_server} to rsync servers? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          ['downloadserver-rsync-prod-1.ops.puppetlabs.net', 'downloadserver-rsync-prod-2.ops.puppetlabs.net'].each do |rsync_server|
```
underscorgan: I think it would be pretty straight forward to add an `rsync_servers` config param and then add that in the release branch of https://github.com/puppetlabs/build-data to get around the hard coding.

danieldreier: I made [PR 104](https://github.com/puppetlabs/build-data/pull/104) on build-data (I saw you already responded) and just pushed an update to this codebase to use that configuration instead of hardcoding it. Would you please re-test?

tasks/ship.rake
```diff
@@ -266,6 +266,22 @@ namespace :pl do
...
+              command = "sudo su - rsync --command 'rsync --verbose -a /opt/repository/#{repo}/ rsync@#{rsync_server}:/opt/repository/#{repo}'"
+              Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.staging_server, command)
+            end
+          end
+        end
+      end
+    end
+
+
     desc "Update rsync servers from AWS S3"
```
underscorgan: Does the new code replace this part?

tasks/ship.rake
```diff
@@ -266,6 +266,22 @@ namespace :pl do
...
     end
 
+    desc "Sync yum and apt from #{Pkg::Config.staging_server} to rsync servers"
+    task :deploy_to_rsync_server => 'pl:fetch' do
+      puts "Really run rsync to sync apt and yum from #{Pkg::Config.staging_server} to rsync servers? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          Pkg::Config.rsync_servers.each do |rsync_server|
+            ['apt', 'yum'].each do |repo|
+              command = "sudo su - rsync --command 'rsync --verbose -a /opt/repository/#{repo}/ rsync@#{rsync_server}:/opt/repository/#{repo}'"
```
underscorgan: I think you need the excludes we used elsewhere in here, this is pulling in all the freight cruft like `/opt/repository/apt/dists/xenial-20161109104933300783045`

underscorgan: Which also probably means we need to copy links as directories for apt and links as links for yum.

danieldreier:
@underscorgan I've removed the old code for syncing the rsync servers from s3 - thanks - and have also updated the rsync to copy from `/opt/repo-s3-stage/repositories/` so that we get symlinks for yum and resolved links for yum.

underscorgan:
@danieldreier great, thanks! I'll run through more tests this afternoon!

tasks/ship.rake
```diff
@@ -266,15 +266,16 @@ namespace :pl do
...
+      # This task must run after the S3 sync has run, or else /opt/repo-s3-stage won't be up-to-date
+      puts "Really run rsync to sync apt and yum from #{Pkg::Config.staging_server} to rsync servers? Only say yes if the S3 sync task has run. [y,n]"
       if Pkg::Util.ask_yes_or_no
         Pkg::Util::Execution.retry_on_fail(:times => 3) do
-          ['downloadserver-rsync-prod-1.ops.puppetlabs.net', 'downloadserver-rsync-prod-2.ops.puppetlabs.net'].each do |destination_server|
+          Pkg::Config.rsync_servers.each do |rsync_server|
             ['apt', 'yum'].each do |repo|
-              command = "sudo aws s3 sync --exclude '*.html' --region us-west-2 --delete s3://#{repo}.puppetlabs.com /opt/repository/#{repo}/"
-              Pkg::Util::Net.remote_ssh_cmd(destination_server, command)
+              command = "sudo su - rsync --command 'rsync --verbose -a /opt/repo-s3-stage/repositories/#{repo}/ rsync@#{rsync_server}:/opt/repository/#{repo}'"
```
underscorgan: I think this should be
```
command = "sudo su - rsync --command 'rsync --verbose -a /opt/repo-s3-stage/repositories/#{repo}.puppetlabs.com/ rsync@#{rsync_server}:/opt/repository/#{repo}'"
```

tasks/ship.rake
```diff
@@ -266,15 +266,16 @@ namespace :pl do
...
-      puts "Really update rsync download servers from AWS S3? [y,n]"
+    desc "Sync yum and apt from #{Pkg::Config.staging_server} to rsync servers"
+    task :deploy_to_rsync_server => 'pl:fetch' do
+      # This task must run after the S3 sync has run, or else /opt/repo-s3-stage won't be up-to-date
+      puts "Really run rsync to sync apt and yum from #{Pkg::Config.staging_server} to rsync servers? Only say yes if the S3 sync task has run. [y,n]"
       if Pkg::Util.ask_yes_or_no
         Pkg::Util::Execution.retry_on_fail(:times => 3) do
-          ['downloadserver-rsync-prod-1.ops.puppetlabs.net', 'downloadserver-rsync-prod-2.ops.puppetlabs.net'].each do |destination_server|
+          Pkg::Config.rsync_servers.each do |rsync_server|
             ['apt', 'yum'].each do |repo|
```
underscorgan: I think we used to have dowloads available via rsync, but not sure if that was just an artifact of how it used to be done rather than something that was actually useful for people. Any thoughts @shrug ?

shrug: Yeah, I have not heard anyone complaining about not being able to sync downloads, only yum and/or apt

underscorgan: ok, cool. We can probably just leave this as is and start syncing downloads if it comes up then 👍

tasks/ship.rake
danieldreier: you're right! Fixing, thanks.

tasks/jenkins.rake
```diff
@@ -269,6 +269,7 @@ namespace :pl do
         remote:deploy_yum_repo_to_s3
         remote:deploy_downloads_to_s3
         remote:update_rsync_from_s3
```
underscorgan: This task was deleted, so it should be removed here too.

tasks/jenkins.rake
```diff
@@ -269,6 +269,7 @@ namespace :pl do
...
+        remote:deploy_to_rsync_server
```
underscorgan: The task name will also need to be updated at line 297

underscorgan:
@danieldreier there were still a few failures, but this might just need to be cleaned up on the rsync servers manually

```
cannot delete non-empty directory: puppet
could not make way for new symlink: puppet
cannot delete non-empty directory: el/50
could not make way for new symlink: el/50
cannot delete non-empty directory: el/51
could not make way for new symlink: el/51
cannot delete non-empty directory: el/52
could not make way for new symlink: el/52
cannot delete non-empty directory: el/53
could not make way for new symlink: el/53
cannot delete non-empty directory: el/54
could not make way for new symlink: el/54
cannot delete non-empty directory: el/55
could not make way for new symlink: el/55
cannot delete non-empty directory: el/56
could not make way for new symlink: el/56
cannot delete non-empty directory: el/57
could not make way for new symlink: el/57
cannot delete non-empty directory: el/58
could not make way for new symlink: el/58
cannot delete non-empty directory: el/59
could not make way for new symlink: el/59
cannot delete non-empty directory: el/5Client
could not make way for new symlink: el/5Client
```

tasks/jenkins.rake
```diff
@@ -268,7 +268,7 @@ namespace :pl do
         remote:deploy_apt_repo_to_s3
         remote:deploy_yum_repo_to_s3
         remote:deploy_downloads_to_s3
-        remote:update_rsync_from_s3
+        remote:deploy_to_rsync_server
```
underscorgan: @danieldreier can you update the task name on line 296? And do you have any thoughts about cleaning up the rsync servers? I'm really excited to get this work in so we can decommission those hosts.

danieldreier: I just pushed a fix for line 296 - thanks for reminding me - and I'm currently in the process of re-syncing the rsync servers. I nuked and redid apt and yum on `downloadserver-rsync-prod-1`, yum on `downloadserver-rsync-prod-2`, and am waiting for apt to finish syncing.

underscorgan: great, thank you!

I can give this another test once that's done then we should be good to get this in I think!

danieldreier: @underscorgan the re-sync has finished on both rsync servers. I think this is ready to re-test.

PR #687, opened 2017-07-07
(maint) Update host for the s3 download sync

underscorgan:
The ask_yes_or_no messages were using `staging_host` but the actual
command was running on `apt_signing_host`. While these are the same
right now, this is dangerous, so we should be consistent.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -259,7 +259,7 @@ namespace :pl do
       puts "Really run S3 sync to sync downloads.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3? [y,n]"
       if Pkg::Util.ask_yes_or_no
         Pkg::Util::Execution.retry_on_fail(:times => 3) do
-          destination_server = Pkg::Config.apt_signing_server
+          destination_server = Pkg::Config.staging_server
           command = 'sudo /usr/local/bin/s3_repo_sync.sh downloads.puppetlabs.com'
           Pkg::Util::Net.remote_ssh_cmd(destination_server, command)
         end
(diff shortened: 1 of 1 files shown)


tasks/ship.rake
```diff
@@ -259,7 +259,7 @@ namespace :pl do
       puts "Really run S3 sync to sync downloads.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3? [y,n]"
       if Pkg::Util.ask_yes_or_no
         Pkg::Util::Execution.retry_on_fail(:times => 3) do
-          destination_server = Pkg::Config.apt_signing_server
+          destination_server = Pkg::Config.staging_server
```
mwaggett: I said this on one of Daniel's PRs, but I feel like `destination_server` is a confusing variable name, since I'm imagining it as the server we're rsyncing to, rather than the destination of the rsync command. Since it's only used once, I feel like it would be fine to just have `Pkg::Config.staging_server` as an argument to the remote_ssh_cmd.

mwaggett:
also while we're at it, we should update line 250 of tasks/ship.rake as well.

underscorgan:
@mwaggett updated.

mwaggett:
Actually nvm I decided I don't care about squashing these.

tasks/ship.rake
```diff
@@ -259,9 +257,8 @@ namespace :pl do
       puts "Really run S3 sync to sync downloads.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3? [y,n]"
       if Pkg::Util.ask_yes_or_no
         Pkg::Util::Execution.retry_on_fail(:times => 3) do
-          destination_server = Pkg::Config.apt_signing_server
           command = 'sudo /usr/local/bin/s3_repo_sync.sh downloads.puppetlabs.com'
-          Pkg::Util::Net.remote_ssh_cmd(destination_server, command)
+          Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.staging_server, command)
```
melissa: should this be `tar_staging_server`? Or just `staging_server`?

PR #688, opened 2017-07-07
(RE-8978) Make regexes more specific

underscorgan:
Our initial regexes for determining platform were pretty open ended.
This lead to issues for instance with packages that matched 'el' (such
as puppet-release. woops), and could have easily been encountered with
codenames. This makes the regexes more anchored and adds some more test
cases.

### DIFF
--- a/spec/lib/packaging/paths_spec.rb
+++ b/spec/lib/packaging/paths_spec.rb
@@ -10,9 +10,12 @@ describe 'Pkg::Paths' do
       'pkg/deb/trusty/pe-r10k_2.5.4.3-1trusty_amd64.deb' => 'ubuntu-14.04-amd64',
       'pkg/pe/rpm/el-6-i386/pe-puppetserver-2017.3.0.3-1.el6.noarch.rpm' => 'el-6-i386',
       'pkg/pe/deb/xenial/pe-puppetserver_2017.3.0.3-1puppet1_all.deb' => 'ubuntu-16.04-i386',
+      'pkg/pe/deb/xenial/super-trusty-package_1.0.0-1puppet1_all.deb' => 'ubuntu-16.04-i386',
+      'artifacts/fedora/24/puppet/x86_64/puppet-release-1.0.0-1.fc24.noarch.rpm' => 'fedora-24-x86_64',
       'artifacts/deb/wheezy/PC1/puppetdb_4.3.1-1puppetlabs1_all.deb' => 'debian-7-i386',
       'pkg/el/7/PC1/x86_64/puppetdb-4.3.1-1.el7.noarch.rpm' => 'el-7-x86_64',
       'pkg/apple/10.11/PC1/x86_64/puppet-agent-1.9.0-1.osx10.11.dmg' => 'osx-10.11-x86_64',
+      'artifacts/mac/10.11/PC1/x86_64/puppet-agent-1.9.0-1.osx10.11.dmg' => 'osx-10.11-x86_64',
       'artifacts/eos/4/PC1/i386/puppet-agent-1.9.0-1.eos4.i386.swix' => 'eos-4-i386',
       'pkg/deb/cumulus/puppet5/puppet-agent_1.4.1.2904.g8023dd1-1cumulus_amd64.deb' => 'cumulus-2.2-amd64',
       'pkg/windows/puppet-agent-1.9.0-x86.msi' => 'windows-2012-x86'
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -11,7 +11,7 @@ module Pkg::Paths
   # Given a path to an artifact, divine the appropriate platform tag associated
   # with the artifact and path
   def tag_from_artifact_path(path)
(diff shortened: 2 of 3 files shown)


melissa:
Should we reorder the PLATFORM_INFO hash to be back in alphabetical order?

underscorgan:
@melissa done

spec/lib/packaging/paths_spec.rb
```diff
@@ -10,6 +10,8 @@
       'pkg/deb/trusty/pe-r10k_2.5.4.3-1trusty_amd64.deb' => 'ubuntu-14.04-amd64',
       'pkg/pe/rpm/el-6-i386/pe-puppetserver-2017.3.0.3-1.el6.noarch.rpm' => 'el-6-i386',
       'pkg/pe/deb/xenial/pe-puppetserver_2017.3.0.3-1puppet1_all.deb' => 'ubuntu-16.04-i386',
+      'pkg/pe/deb/xenial/super-trusty-package_1.0.0-1puppet1_all.deb' => 'ubuntu-16.04-i386',
+      'artifacts/fedora/24/puppet/x86_64/puppet-release-1.0.0-1.fc24.noarch.rpm' => 'fedora-24-x86_64',
       'artifacts/deb/wheezy/PC1/puppetdb_4.3.1-1puppetlabs1_all.deb' => 'debian-7-i386',
       'pkg/el/7/PC1/x86_64/puppetdb-4.3.1-1.el7.noarch.rpm' => 'el-7-x86_64',
       'pkg/apple/10.11/PC1/x86_64/puppet-agent-1.9.0-1.osx10.11.dmg' => 'osx-10.11-x86_64',
```
melissa: Did you figure out how this works for osx? I'm also tempted to say we add in another test case to make sure it works for when dmgs are staged in a `mac` directory, just in case

spec/lib/packaging/paths_spec.rb
```diff
@@ -10,6 +10,8 @@
       'pkg/deb/trusty/pe-r10k_2.5.4.3-1trusty_amd64.deb' => 'ubuntu-14.04-amd64',
       'pkg/pe/rpm/el-6-i386/pe-puppetserver-2017.3.0.3-1.el6.noarch.rpm' => 'el-6-i386',
       'pkg/pe/deb/xenial/pe-puppetserver_2017.3.0.3-1puppet1_all.deb' => 'ubuntu-16.04-i386',
+      'pkg/pe/deb/xenial/super-trusty-package_1.0.0-1puppet1_all.deb' => 'ubuntu-16.04-i386',
+      'artifacts/fedora/24/puppet/x86_64/puppet-release-1.0.0-1.fc24.noarch.rpm' => 'fedora-24-x86_64',
```
mwaggett: Could we maybe also get a fedora test case for a path that has the f prefix?

mwaggett: It'd be great if I could use this logic for the stuff I've been working on for beaker.

spec/lib/packaging/paths_spec.rb
underscorgan: I missed the `?` in the regex at https://github.com/puppetlabs/packaging/blob/1.0.x/lib/packaging/paths.rb#L22

I'll add something similar to that with mac, just for future us's sake.

spec/lib/packaging/paths_spec.rb
underscorgan: The 1.0.x branch doesn't support f-prefixed packages. We kept master compatible with either way, but 1.0.x only works for the new fedora package naming

spec/lib/packaging/paths_spec.rb
```diff
@@ -10,9 +10,12 @@
...
       'artifacts/deb/wheezy/PC1/puppetdb_4.3.1-1puppetlabs1_all.deb' => 'debian-7-i386',
       'pkg/el/7/PC1/x86_64/puppetdb-4.3.1-1.el7.noarch.rpm' => 'el-7-x86_64',
       'pkg/apple/10.11/PC1/x86_64/puppet-agent-1.9.0-1.osx10.11.dmg' => 'osx-10.11-x86_64',
+      'artifacts/mac/10.11/PC1/x86_64/puppet-agent-1.9.0-1.osx10.11.dmg' => 'osx-10.11-x86_64',
```
underscorgan: @melissa ^^ done!

spec/lib/packaging/paths_spec.rb
mwaggett: I still feel like it would be valuable if we could recognize a path with the f as belonging to the appropriate platform tag.  ¯\\_(ツ)_/¯

underscorgan: I feel like special casing for this is a bad idea, if we want to be able to recognize f-prefixed versioned we need to get that into the platform map.

underscorgan: FWIW 1.0.x before these changes also doesn't recognize f-prefixed versions.

PR #689, opened 2017-07-10
(RE-8942) Update training VM shipping to archive old VMs

underscorgan:
If we're shipping a new master or training VM, archive the existing ones
so we only have the latest VM at the top level.


melissa:
@samuelson can you verify that this is the behavior you want moving forward? This will archive everything except for the latest thing that is currently shipping, so there will only be one thing available in the main download directory at a time

samuelson:
LGTM

underscorgan:
This has two approvals and @samuelson confirmed this looks right, so merging!

PR #690, opened 2017-07-10
(maint) Updates requested by beaker team

mwaggett:
This PR updates the `platform_data` hash that beaker will consume. I have changed the artifact and repo_config paths to be relative to the location of the <ref>.yaml file (i.e. the artifacts directory) and removed the f prefix from the fedora keys (in this hash only). 

I also added back the non-f-prefixed versions to the `PLATFORM_INFO` map. This will allow us to locate paths with and without f prefixes. The breakage in the puppet-agent pipeline last week was due to packages on builds not having the f-prefix, but beaker looking for packages *with* the f prefix. I believe this was due to the removal of the f prefix in the puppet-agent vanagon configs, rather than anything packaging-related.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -94,6 +94,9 @@ module Pkg
             else
               fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
             end
+            # Remove the f-prefix from the fedora platform tag keys so that
+            # beaker can rely on consistent keys once we rip out the f for good
+            tag = tag.sub(/fedora-f/, 'fedora-')
             data[tag] = { :artifact => artifact.sub('artifacts/', ''),
                           :repo_config => repo_config,
                         } if artifact
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -85,16 +85,16 @@ module Pkg
             case package_format
             when 'deb'
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and (e.include?("all") || e.include?("#{arch}.deb")) }
-              repo_config = "./repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Util::Platform.get_attribute(tag, :codename)}.list" if artifact
+              repo_config = "../repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Util::Platform.get_attribute(tag, :codename)}.list" if artifact
             when 'rpm'
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
-              repo_config = "./repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
+              repo_config = "../repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
             when 'swix', 'svr4', 'ips', 'dmg', 'msi'
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
             else
               fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
             end
-            data[tag] = { :artifact => artifact,
+            data[tag] = { :artifact => artifact.sub('artifacts/', ''),
                           :repo_config => repo_config,
                         } if artifact
           end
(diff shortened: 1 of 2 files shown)


joshcooper:
@mwaggett could you point me to fedora packages built with this PR, so we can verify existing puppet-agent & beaker can consume them?

mwaggett:
@joshcooper whoops, I should've thrown a label on this. It's not quite done, but should be very shortly, and then I'll definitely link to packages.

mwaggett:
@joshcooper okay, I think this is ready to go! I built puppet-agent packages from a test branch pointing to this branch of packaging: http://builds.delivery.puppetlabs.net/puppet-agent/ff3e02ddf16781f94b50918b995c6e793597f9a9/

lib/packaging/config.rb
```diff
@@ -85,16 +85,17 @@ def platform_data
...
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
-              repo_config = "./repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
+              repo_config = "../repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
             when 'swix', 'svr4', 'ips', 'dmg', 'msi'
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
             else
               fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
             end
-            data[tag] = { :artifact => artifact,
+            tag = tag.sub('f2', '2')
```
underscorgan: Is there a better way we can do this? This feel like the sort of thing we're going to forget about until it bites us (fedora 30, or some weird new platform tag, or something). Maybe something with regex captures?

lib/packaging/config.rb
```diff
@@ -85,16 +85,17 @@ def platform_data
...
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
-              repo_config = "./repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
+              repo_config = "../repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
             when 'swix', 'svr4', 'ips', 'dmg', 'msi'
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
             else
               fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
             end
-            data[tag] = { :artifact => artifact,
+            tag = tag.sub(/fedora-f/, 'fedora-')
```
mwaggett: @underscorgan How does this look?

underscorgan: That looks better. Can we get a comment in the code about why we're doing this?

lib/packaging/config.rb
```diff
@@ -85,16 +85,19 @@ def platform_data
...
-              repo_config = "./repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
+              repo_config = "../repo_configs/rpm/pl-#{self.project}-#{self.ref}-#{tag}.repo" if artifact
             when 'swix', 'svr4', 'ips', 'dmg', 'msi'
               artifact = artifacts.find { |e| e.include? Pkg::Util::Platform.artifacts_path(tag) and e.include?(arch) }
             else
               fail "Not sure what to do with packages with a package format of '#{package_format}' - maybe update PLATFORM_INFO?"
             end
-            data[tag] = { :artifact => artifact,
+            # Remove the f-prefix from the fedora platform tag keys so that
+            # beaker can rely on consistent keys once we rip out the f for good
```
mwaggett: @underscorgan comments added

mwaggett:
still built successfully with the appropriate info in the yaml after small fixes: http://builds.delivery.puppetlabs.net/puppet-agent/f0005a296b2e1f5d6a7541aac005d41550690bc5/

shrug:
@joshcooper Did you want to take another look at this or are we ok to merge?

joshcooper:
Checking now

joshcooper:
Looks good to me. Both paths look correct relative to the build_data.yaml:

```ruby
require 'net/http'
require 'uri'

sha = "ff3e02ddf16781f94b50918b995c6e793597f9a9"
base = URI.parse("http://builds.delivery.puppetlabs.net/puppet-agent/#{sha}/artifacts/#{sha}.yaml")
yaml = Net::HTTP.get(base)
build_data = YAML.load(yaml)
fedora25 = build_data[:platform_data]['fedora-25-x86_64']

Net::HTTP.start(base.host, base.port) do |http|
  puts http.head(base.merge(fedora25[:repo_config])).inspect
  puts http.head(base.merge(fedora25[:artifact])).inspect
end
```

```
$ bx ruby download.rb
#<Net::HTTPOK 200 OK readbody=true>
#<Net::HTTPOK 200 OK readbody=true>
```

PR #693, opened 2017-07-11
(RE-8838) Stage packages in platform tag dir

melissa:


tasks/deb.rake
```diff
@@ -69,7 +69,8 @@ task :build_deb, :deb_command, :cow do |t, args|
     work_dir  = Pkg::Util::File.mktemp
     subdir    = 'pe/' if Pkg::Config.build_pe
     codename = /base-(.*)-(.*)\.cow/.match(cow)[1] unless cow.nil?
-    dest_dir  = File.join(Pkg::Config.project_root, "pkg", "#{subdir}deb", codename, Pkg::Paths.repo_name.to_s)
+    arch = /base-(.*)-(.*)\.cow/.match(cow)[2] unless cow.nil?
+    dest_dir  = File.join(Pkg::Config.project_root, "pkg", "#{Pkg::Platforms.codename_to_platform_version(codename)}-#{arch}")
```
melissa: I really hate this

tasks/mock.rake
```diff
@@ -134,6 +134,25 @@ def mock_el_ver(mock_config)
...
+  else
+    components = mock_config.split('-')
+    if (first == 'el' || first == 'fedora') || (first == 'pl' && second.match(/^\d+$/))
+      arch = components[2]
+    else
+      arch = components[3]
+    end
+  end
+  arch
+end
```
melissa: I also really hate this

melissa: and I have no idea if it works

melissa:
I shipped and created the repos for http://builds.puppetlabs.lan/puppet-agent/f8b38b355fdf54b31b5896029af7e0f16f66892d/

and http://builds.puppetlabs.lan/puppetserver/5.0.1.SNAPSHOT.2017.07.28T0027/

melissa:
@haus this is still definitely a WIP, so the commits and tests are a mess right now. BUT, I'd love to get your thoughts on it!

melissa:
So, confirmed, this is working for both ezbake and vanagon sign/ship, and also building packages for ezbake

melissa:
I broke the rpm repo_configs somehow. The deb repo_configs are fine though

```
melissa@jenkins-builds1:~$ ls /opt/jenkins-builds/puppetserver/5.0.1.SNAPSHOT.2017.07.28T0027/repo_configs/deb/
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-jessie.list
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-xenial.list
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-yakkety.list
```
```
melissa@jenkins-builds1:~$ ls /opt/jenkins-builds/puppetserver/5.0.1.SNAPSHOT.2017.07.28T0027/repo_configs/rpm
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-6-i386.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-6-s390x.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-6-srpms.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-6-x86_64.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-7-ppc64le.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-7-s390x.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-7-srpms.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-el-7-x86_64.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-sles-12-s390x.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-sles-12-srpms.repo
pl-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-puppetserver-5.0.1.SNAPSHOT.2017.07.28T0027-sles-12-x86_64.repo
```

melissa:
Okie, fixed that issue with the repo_config files. Example at http://builds.puppetlabs.lan/puppetserver/5.0.1.SNAPSHOT.2017.07.28T2129/

melissa:
TESTS ARE FIXED AND COMMIT MESSAGES ARE FABULOUS. KINDLY REVIEW.

sarameisburger:
@melissa you're awesome and this looks awesome! unfortunately it all goes completely over my head but i'd love for you to explain to me sometime :)

melissa:
@sarameisburger take a look at the code! I broke it up into smaller bits in the commit messages, so see if you can parse those one at a time! This is a great way to learn!

lib/packaging/deb/repo.rb
```diff
@@ -67,53 +69,72 @@ def retrieve_repo_configs(target = "repo_configs")
...
+    def directories_that_contain_deb_packages(artifact_directory)
+      cmd = "[ -d #{artifact_directory} ] || exit 1 ; "
+      cmd << "pushd #{artifact_directory} > /dev/null ; "
+      cmd << 'find ./ -name "*.deb" | xargs -I {} dirname {} ; '
+      begin
+        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.distribution_server, cmd, true)
+        return stdout.split
+      rescue => e
+        fail "Could not retrieve deb directories"
+      end
```
mckern: You can omit the explicit `begin` and use inline `rescue`:

```ruby
def somefunction
  Object.do_the_thing
rescue ArgumentError
  $stderr.puts "The thing was not done"
end
```

lib/packaging/deb/repo.rb
```diff
@@ -67,53 +69,72 @@ def retrieve_repo_configs(target = "repo_configs")
...
-      cmd << "pushd #{artifact_directory}/artifacts/#{prefix}deb && dists=$(ls) && popd; "
-      # We do one more check here to make sure we actually have distributions
-      # to build for. If deb is empty we want to just exit.
-      #
-      cmd << '[ -n "$dists" ] || exit 1 ; '
-      cmd << "pushd #{artifact_directory} ; "
+    def directories_that_contain_deb_packages(artifact_directory)
+      cmd = "[ -d #{artifact_directory} ] || exit 1 ; "
+      cmd << "pushd #{artifact_directory} > /dev/null ; "
+      cmd << 'find ./ -name "*.deb" | xargs -I {} dirname {} ; '
```
mckern: `find . -name "*.deb" -print0 | xargs -0 -I {} dirname {} ; ` is safer (null-terminated strings will keep the pipeline or xargs from choking if a name has some unexpected characters in it).

mckern: And if you're sure this will only ever hit a GNU version of `xargs`, don't forget `--no-run-if-empty`

lib/packaging/deb/repo.rb
```diff
@@ -46,10 +46,12 @@ def generate_repo_configs(source = "repos", target = "repo_configs")
...
         # We want to skip the base_url, which wget returns as one of the results
         next if "#{url}/" == repo_base
-        dist = url.split('/').last
+        platform_tag = Pkg::Paths.tag_from_artifact_path(url)
+        platform, version, _ = Pkg::Platforms.parse_platform_tag(platform_tag)
+        codename = Pkg::Platforms.codename_for_platform_version(platform, version)
         repoconfig = ["# Packages for #{Pkg::Config.project} built from ref #{Pkg::Config.ref}",
-                      "deb #{url} #{dist} #{Pkg::Paths.repo_name}"]
-        config = File.join("pkg", target, "deb", "pl-#{Pkg::Config.project}-#{Pkg::Config.ref}-#{dist}.list")
+                      "deb #{url} #{codename} #{Pkg::Paths.repo_name.empty? ? 'main' : Pkg::Paths.repo_name}"]
```
mckern: Maybe move the ternary operations out of the string interpolation? An intermediate variable is a reasonable compromise.

lib/packaging/deb/repo.rb
```diff
@@ -67,53 +69,72 @@ def retrieve_repo_configs(target = "repo_configs")
...
+        stdout, stderr = Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.distribution_server, cmd, true)
+        return stdout.split
+      rescue => e
+        fail "Could not retrieve deb directories"
+      end
+    end
+
+    def populate_repo_directory(artifact_parent_directory)
+      cmd = "[ -d #{artifact_parent_directory}/artifacts ] || exit 1 ; "
+      cmd << "pushd #{artifact_parent_directory} ; "
```
mckern: `pushd` is redirected to `/dev/null` above. Should it be here? I think it only matters for consistency but consistency is useful.

lib/packaging/deb/repo.rb
```diff
@@ -67,53 +69,72 @@ def retrieve_repo_configs(target = "repo_configs")
...
+      cmd << "rsync -avxl --ignore-existing artifacts/ repos/ ; "
```
mckern: Long flags (`--archive --verbose --one-file-system --links`)? Also, `--archive` already had `--links` included so you could shorten up the flags a little bit.

lib/packaging/deb/repo.rb
```diff
@@ -67,53 +69,72 @@ def retrieve_repo_configs(target = "repo_configs")
...
+      begin
+        Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.distribution_server, cmd)
+      rescue => e
+        fail "Could not populate repo directory"
+      end
```
mckern: Same comment about redundant `begin`.

lib/packaging/deb/repo.rb
```diff
@@ -67,53 +69,72 @@ def retrieve_repo_configs(target = "repo_configs")
...
+    end
 
+    def repo_creation_command(repo_directory, artifact_paths)
+      cmd = "[ -d #{repo_directory} ] || exit 1 ; "
+      cmd << "pushd #{repo_directory} ; "
```
mckern: Same comment about `/dev/null` redirection

lib/packaging/deb/repo.rb
```diff
@@ -67,53 +69,72 @@ def retrieve_repo_configs(target = "repo_configs")
...
 
       # Make the conf directory and write out our configuration file
       cmd << "rm -rf apt && mkdir -p apt ; pushd apt ; "
-      cmd << %Q(for dist in $dists ; do mkdir -p $dist/conf ; pushd $dist ;
-      echo "
+      artifact_paths.each do |path|
+        platform_tag = Pkg::Paths.tag_from_artifact_path(path)
+        platform, version, _ = Pkg::Platforms. parse_platform_tag(platform_tag)
+        codename = Pkg::Platforms.codename_for_platform_version(platform, version)
+        default_arches = ['i386', 'amd64', 'arm64', 'armel', 'armhf', 'powerpc', 'ppc64el', 'sparc', 'mips', 'mipsel']
```
mckern: This could be moved out of the loop. It's not modified in the loop body.

lib/packaging/deb/repo.rb
```diff
@@ -156,15 +177,19 @@ def sign_repos(target = "repos", message = "Signed apt repository")
       Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
 
       dists = Pkg::Util::File.directories("#{target}/apt")
+      supported_codenames = Pkg::Platforms.codenames('deb')
 
       if dists
         dists.each do |dist|
+          next unless supported_codenames.include?(dist)
+          default_arches = ['i386', 'amd64', 'arm64', 'armel', 'armhf', 'powerpc', 'ppc64el', 'sparc', 'mips', 'mipsel']
```
mckern: Same comment about moving out of the loop body.

lib/packaging/paths.rb
```diff
@@ -11,7 +11,7 @@ module Pkg::Paths
   # Given a path to an artifact, divine the appropriate platform tag associated
   # with the artifact and path
   def tag_from_artifact_path(path)
-    platform = Pkg::Platforms.supported_platforms.find { |p| path =~ /(\/|\.)#{p}/ }
+    platform = Pkg::Platforms.supported_platforms.find { |p| path =~ /(\/|\.)#{p}()[^\.]/ }
```
mckern: Intentional empty capture group?

lib/packaging/rpm/repo.rb
```diff
@@ -206,7 +207,7 @@ def create_repos_from_artifacts
       cmd << "while [ -f .lock ] ; do sleep 1 ; echo -n '.' ; done ; "
       cmd << 'echo "Setting lock" ; '
       cmd << "touch .lock ; "
-      cmd << "rsync -avxl artifacts/ repos/ ; pushd repos ; "
+      cmd << "rsync -avxl --ignore-existing artifacts/ repos/ ; pushd repos ; "
```
mckern: Same comment about long flags

lib/packaging/deb/repo.rb
```diff
@@ -156,15 +177,19 @@ def sign_repos(target = "repos", message = "Signed apt repository")
...
+          arches = Pkg::Platforms.arches_for_codename(dist)
           Dir.chdir("#{target}/apt/#{dist}") do
             File.open("conf/distributions", "w") do |f|
               f.puts "Origin: Puppet Labs
 Label: Puppet Labs
 Codename: #{dist}
-Architectures: i386 amd64 arm64 armel armhf powerpc ppc64el sparc mips mipsel
+Architectures: #{(default_arches + arches).uniq.join(' ')}
```
e-gris: Nice start-of-abstraction. Eventually, we will want to see default_arches moved out of code and into properties.

tasks/mock.rake
e-gris: 1. "I had a problem to solve. So I added a regex. Now I have two problems" :-)
1. Nested if/elses are always code smells for me.

Unfortunately, I'm still too ignorant about the code to be helpful, so I chose snark instead.

tasks/mock.rake
```diff
@@ -206,52 +226,22 @@ def build_rpm_with_mock(mocks)
...
-            when /src\.rpm/
-              FileUtils.cp_r(rpm, "pkg/#{family}/#{version}/#{subdir}/SRPMS", { :preserve => true })
-            when /i.?86/
-              FileUtils.cp_r(rpm, "pkg/#{family}/#{version}/#{subdir}/i386", { :preserve => true })
-            when /x86_64/
-              FileUtils.cp_r(rpm, "pkg/#{family}/#{version}/#{subdir}/x86_64", { :preserve => true })
-            when /noarch/
-              FileUtils.cp_r(rpm, "pkg/#{family}/#{version}/#{subdir}/i386", { :preserve => true })
-              FileUtils.ln("pkg/#{family}/#{version}/#{subdir}/i386/#{File.basename(rpm)}", "pkg/#{family}/#{version}/#{subdir}/x86_64/", :force => true, :verbose => true)
+        %x(mkdir -p pkg/#{family}-#{version}-{srpms,#{arches.join(',')}})
```
e-gris: Yay! Nested case/if is even smellier than nested if/else. 10 points for gryffindor!

tasks/pe_sign.rake
```diff
@@ -1,44 +1,13 @@
...
-          if (family == 'el' and version >= 6) || (family == 'sles' and version >= 12)
-            modern_rpms += FileList[rpm_stagedir] + FileList[srpm_stagedir]
-          else
-            old_rpms += FileList[rpm_stagedir] + FileList[srpm_stagedir]
-          end
-        end
-      end
-      sign_legacy_rpm(old_rpms) unless old_rpms.empty?
-      sign_rpm(modern_rpms) unless modern_rpms.empty?
+      Pkg::Util::RakeUtils.invoke_task("pl:sign_rpms")
```
e-gris: Nice refactor/abstraction!

lib/packaging/deb/repo.rb
melissa: I needed to redirect it to `/dev/null` above because I was collecting output. I'm not here, but I'm definitely not opposed to it for consistency

lib/packaging/deb/repo.rb
melissa: Hopefully we won't need this at all when we switch to artifactory!

melissa:
@mckern I think I responded to most of your comments. I also updated the rpm repo stuff, which was fun.

lib/packaging/deb/repo.rb
```diff
@@ -4,6 +4,13 @@
 module Pkg::Deb::Repo
 
   class << self
+
+    # This is the default set of arches we are using for our reprepro repos. We
+    # take this list and combine it with the list of supported arches for each
+    # given platform to ensure a complete set of architectures. We use this
+    # when we initially create the repos and when we sign the repos.
+    REPREPRO_ARCHES = ['i386', 'amd64', 'arm64', 'armel', 'armhf', 'powerpc', 'ppc64el', 'sparc', 'mips', 'mipsel']
```
shrug: We don't actually have all of these architectures, should we trim it back?

lib/packaging/deb/repo.rb
```diff
@@ -59,61 +70,56 @@ def retrieve_repo_configs(target = "repo_configs")
       wget = Pkg::Util::Tool.check_tool("wget")
       FileUtils.mkdir_p("pkg/#{target}")
       config_url = "#{base_url}/#{target}/deb/"
-      begin
```
shrug: did we drop the begin?

lib/packaging/rpm/repo.rb
```diff
@@ -22,11 +22,22 @@ def ship_repo_configs(target = "repo_configs")
...
     end
 
-    def repo_creation_command(createrepo)
-      cmd = 'for repodir in $(find ./ -name "*.rpm" | xargs -I {} dirname {}) ; do '
-      cmd << "[ -d ${repodir} ] || continue; "
-      cmd << "pushd ${repodir} && #{createrepo} --checksum=sha --checkts --update --delta-workers=0 --database . ; popd ; "
-      cmd << "done "
+    def repo_creation_command(repo_directory, artifact_paths)
+      cmd = "[ -d #{repo_directory} ] || exit 1 ; "
+      cmd << "pushd #{repo_directory} > /dev/null ; "
```
shrug: What happens if this fails?

lib/packaging/rpm/repo.rb
```diff
@@ -22,11 +22,22 @@ def ship_repo_configs(target = "repo_configs")
...
+      cmd << 'echo "Checking for running repo creation. Will wait if detected." ; '
+      cmd << 'while [ -f .lock ] ; do sleep 1 ; echo -n "." ; done ; '
+      cmd << 'echo "Setting lock" ; '
+      cmd << 'touch .lock ; '
```
shrug: What happens if this fails?

lib/packaging/rpm/repo.rb
```diff
@@ -22,11 +22,22 @@ def ship_repo_configs(target = "repo_configs")
...
+      cmd << 'createrepo=$(which createrepo) ; '
+
+      artifact_paths.each do |path|
+        cmd << "[ -d #{path} ] || continue; "
+        cmd << "pushd #{path} ; "
```
shrug: Same question - what if this fails?

tasks/sign.rake
```diff
@@ -79,40 +79,37 @@ namespace :pl do
...
+
+      sig_type = Pkg::Platforms.signature_format_for_tag(platform_tag)
+      case sig_type
+      when 'v3'
+        v3 << rpm
+      when 'v4'
+        v4 << rpm
+      else
+        fail "Cannot find signature type for package '#{rpm}'"
+      end
```
shrug: Thank you for all of this :)

lib/packaging/deb/repo.rb
melissa: I thought about it, but there's no harm in including them. We've always had this giant set of arches, even when it was only `i386` and `amd64` we were building. I can trim it back, but I don't think it's harming anything. I can pair it back if that's what folks want.

lib/packaging/deb/repo.rb
melissa: See comments made by @mckern. Apparently this is a simpler way of rescuing, with inline rescues

lib/packaging/rpm/repo.rb
melissa: well, the first thing we do is check to make sure the directory exists. Are you worried about the redirected output?

lib/packaging/rpm/repo.rb
melissa: I'm not sure what you're getting at with that question? Are you asking for a rescue block? Do you want to make sure we're collecting stderr?

tasks/mock.rake
melissa: This totes works, I tested it, but it's so so so so so so so so so gross.

tasks/sign.rake
melissa: totes magotes!

lib/packaging/rpm/repo.rb
shrug: We check that it exists, but we don't do any error handling if anything fails - for example if the directory permissions don't allow is to cd there, then we'll create .lock in the local directory, etc. The easies thing is probably to chain the shell commands via '&&' instead of having them be discrete commands. At least anything that could potentially fail and leave us in a weird state.

lib/packaging/deb/repo.rb
```diff
@@ -37,6 +44,8 @@ def generate_repo_configs(source = "repos", target = "repo_configs")
       # We want to exclude index and robots files and only include the http: prefixed elements
       repo_urls = stdout.split.uniq.reject { |x| x =~ /\?|index|robots/ }.select { |x| x =~ /http:/ }.map { |x| x.chomp('/') }
 
+      # This is a required field for reprepro, so must never be an empty string
+      local_repo = Pkg::Paths.repo_name.empty? ? 'main' : Pkg::Paths.repo_name
```
haus: can `repo_name` be nil? if so this will blow up.

haus: looks like it can't be nil.

lib/packaging/deb/repo.rb
haus: all methods have an implicit begin

shrug: Yep, just wanted to make sure it was intentional

lib/packaging/deb/repo.rb
melissa: nope, can't be nil
```
  def repo_name
    if Pkg::Util::Version.final?
      Pkg::Config.repo_name || ""
    else
      if Pkg::Config.nonfinal_repo_name
        Pkg::Config.nonfinal_repo_name
      else
        fail "You are attempting to ship a non-final build without specifying a non-final repo destination. Either make sure you are shipping a final version or define `nonfinal_repo_name` in your build_defaults.\nIf this is a test build and you want to allow tagged versions with dirty trees to be final builds, set ALLOW_DIRTY_TREE=true."
      end
    end
  end
```

tasks/mock.rake
underscorgan: should we use `FileUtils.mkdir_p()` here instead of shelling out?

tasks/mock.rake
```diff
@@ -206,52 +226,22 @@ def build_rpm_with_mock(mocks)
...
+        case File.basename(rpm)
+        when /debuginfo/
+          rm_rf(rpm)
```
underscorgan: I can't find where we're defining this, should we be calling the `FileUtils.rm_rf`?

tasks/pe_ship.rake
```diff
@@ -110,17 +110,9 @@ if Pkg::Config.build_pe
             Pkg::Util::Net.rsync_to("pkg/pe/deb/#{dist}/*_#{arch}.deb", Pkg::Config.apt_host, "#{archive_path}/")
           end
 
-          # Ship all-arch debs to same dist-location, but to all known
-          # architectures for this distribution.
-          #
-          # I am not proud of this. MM - 1/3/2014.
```
underscorgan: \o/ cleaning up 3.5 year old hackiness!

tasks/sign.rake
underscorgan: this is great.

lib/packaging/paths.rb
```diff
@@ -29,7 +29,7 @@ def tag_from_artifact_path(path)
     end
     # if we didn't find a platform or a version, probably a codename
     if platform.nil? || version.nil?
-      codename = Pkg::Platforms.codenames('deb').find { |c| path =~ /\/#{c}\// }
+      codename = Pkg::Platforms.codenames('deb').find { |c| path =~ /\/#{c}/ }
```
underscorgan: Hm, is there any set of codename-ending characters we could anchor this with?

lib/packaging/deb/repo.rb
```diff
@@ -59,61 +70,56 @@ def retrieve_repo_configs(target = "repo_configs")
...
-      cmd << "pushd #{artifact_directory} ; "
-
-      cmd << 'echo "Checking for running repo creation. Will wait if detected." ; '
-      cmd << "while [ -f .lock ] ; do sleep 1 ; echo -n '.' ; done ; "
-      cmd << 'echo "Setting lock" ; '
-      cmd << "touch .lock ; "
-      cmd << "rsync -avxl artifacts/ repos/ ; pushd repos ; "
+    def repo_creation_command(repo_directory, artifact_paths)
+      cmd = "[ -d #{repo_directory} ] || exit 1 && "
+      cmd << "pushd #{repo_directory} > /dev/null && "
```
melissa: @shrug better?

lib/packaging/deb/repo.rb
melissa: Mostly I was trying to change as little as possible with this

lib/packaging/paths.rb
melissa: I don't think so. There are some of the paths that we search when finding which directories contain repos that return the url with the codename at the end and nothing else. I suppose we could have an or here, maybe something like `($|\/|-)`?

lib/packaging/platforms.rb
```diff
@@ -191,25 +191,12 @@ def get_attribute(platform_tag, attribute_name)
...
-    type = "by_#{format}".to_sym
-    define_method(type) do
-      by_package_format format
-    end
+    PLATFORM_INFO[platform][version][:package_format]
+  end
+
+  def signature_format_for_tag(platform_tag)
+    platform, version = parse_platform_tag(platform_tag)
+    PLATFORM_INFO[platform][version][:signature_format]
```
melissa: Ruby 1.9.3? No, this means we support Ruby 2.2.0 again :( I was previously testing with Ruby 2.4.0, and apparently 2.2 doesn't like this magic...

melissa: I don't really understand why. But I ran into enough issues with the format_by method, I decided I wanted to kill it.

tasks/mock.rake
haus: don't worry, just involving rake in the process includes fileutils implicitly. but being explicit about where the method is coming from seems good.

```
rirb(main):001:0> require 'rake'
=> true
irb(main):002:0> cd 'acceptance'
cd acceptance
=> nil
irb(main):003:0> `ls`
=> "Gemfile\nGemfile.lock\nREADME.md\nRakefile\nfiles\njunit\nlib\nlog\npre-suite\ntests\ntmp\nvendor\n"
irb(main):004:0> cp 'Gemfile', 'Gemfile1'
cp Gemfile Gemfile1
=> nil
irb(main):005:0> `ls`
=> "Gemfile\nGemfile.lock\nGemfile1\nREADME.md\nRakefile\nfiles\njunit\nlib\nlog\npre-suite\ntests\ntmp\nvendor\n"
irb(main):006:0> rm 'Gemfile1'
rm Gemfile1
=> ["Gemfile1"]
```

lib/packaging/paths.rb
haus: which codenames break with a literal / present?

lib/packaging/platforms.rb
haus: So this would be a breaking change to the available methods, which seems not great to include in a 1.0.x release.

lib/packaging/paths.rb
melissa: When we create the repo configs, we do a wget and pull down the directory structure, which returns a url that ends with a codename, i.e., `http://saturn.puppetlabs.net/deb_repos/1234abcd/repos/apt/xenial`

lib/packaging/platforms.rb
melissa: Oh shoot, you're right. I thought I had done this differently. Fixing now.

melissa: I added back in `by_deb` and `by_rpm`, I just kept out the part where we dynamically create methods

haus: sweet, less metaprogramming

lib/packaging/deb/repo.rb
MosesMendoza: Is this, more broadly speaking, the arches we have for debian packaging, or is it specific to reprepro? If the former, can we rename to something like 'DEBIAN_PACKAGING_ARCHITECTURES' so that if we ever swap out reprepro for something else we don't have to rename and change the API?

lib/packaging/deb/repo.rb
MosesMendoza: Would it work to move this out to its own method or a constant so we don't have to redefine it later? Stylistic choice I suppose.

tasks/pe_ship.rake
MosesMendoza: 👏 👏 👏 👏 👏

lib/packaging/deb/repo.rb
```diff
@@ -4,6 +4,19 @@
 module Pkg::Deb::Repo
 
   class << self
+
+    # This is the default set of arches we are using for our reprepro repos. We
+    # take this list and combine it with the list of supported arches for each
+    # given platform to ensure a complete set of architectures. We use this
+    # when we initially create the repos and when we sign the repos.
+    DEBIAN_PACKAGING_ARCHES = ['i386', 'amd64', 'arm64', 'armel', 'armhf', 'powerpc', 'ppc64el', 'sparc', 'mips', 'mipsel']
```
melissa: @MosesMendoza better?

tasks/pe_ship.rake
mckern: 2014 Moses was my favorite Moses. It was a good vintage.

lib/packaging/deb/repo.rb
```diff
@@ -59,61 +73,53 @@ def retrieve_repo_configs(target = "repo_configs")
...
-
-      # Create the repositories using reprepro. Since these are for acceptance
-      # testing only, we'll just add the debs and ignore source files for now.
-      #
-      cmd << "reprepro=$(which reprepro) ; "
-      cmd << %Q($reprepro includedeb $dist ../../#{prefix}deb/$dist#{"/" unless Pkg::Paths.repo_name.empty?}#{Pkg::Paths.repo_name}/*.deb ; popd ; done ; )
-      cmd << "popd ; popd "
+Codename: #{codename}
+Architectures: #{(DEBIAN_PACKAGING_ARCHES + arches).uniq.join(' ')}
+Components: #{DEBIAN_REPO_NAME}
```
melissa: @MosesMendoza better?

lib/packaging/deb/repo.rb
melissa: @shrug more thoughts on this? Should I just use the list of supported platforms for a given codename only? Or should I keep this list around? I don't know what the better solution would be. I just carried forward what we were doing before

mckern:
rooting for you

lib/packaging/deb/repo.rb
shrug: Just use it as-is, imo, it's what we've been doing and this will largely go away as things migrate into artifactory

lib/packaging/deb/repo.rb
shrug: Yeah,I think so. Were you able to test it?

melissa: yes and no. I manually ran this with a failing case, and it exited as expected. I ran the full code on a case that did exist, and it succeeded as expected

PR #697, opened 2017-07-13
Maint/1.0.x/merge up

melissa:


lib/packaging/config.rb
```diff
@@ -85,16 +85,19 @@ def platform_data
             case package_format
             when 'deb'
               artifact = artifacts.find { |e| e.include? Pkg::Paths.artifacts_path(tag) and (e.include?("all") || e.include?("#{arch}.deb")) }
-              repo_config = "./repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Platforms.get_attribute(tag, :codename)}.list" if artifact
+              repo_config = "../repo_configs/deb/pl-#{self.project}-#{self.ref}-#{Pkg::Util::Platform.get_attribute(tag, :codename)}.list" if artifact
```
mwaggett: I don't think we actually want this, since things got refactored in 1.0.x.

mwaggett: this being the Util part

melissa: fixed

PR #706, opened 2017-08-15
(RE-9133) Add task to packaging for syncing latest file to s3

demophoon:
Because it is useful to have a file which has the latest version of a
released package on s3 for cgi-bin scripts to correctly resolve 'latest'
this commit creates automation around this task.

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -317,5 +317,19 @@ namespace :pl do
       Pkg::Util::RakeUtils.invoke_task("pl:jenkins:prepare_signed_repos", target_host, 'signed', 'version')
       Pkg::Util::RakeUtils.invoke_task("pl:jenkins:deploy_signed_repos_to_s3", target_bucket)
     end
+
+    task :update_release_versions, [:target_bucket] do |t, args|
+      target_bucket = args.target_bucket or fail ":target_bucket is a required argument to #{t}"
+
+      tempdir = Pkg::Util::File.mktemp
+      latest_filepath = File.join(tempdir, "pkg", "#{Pkg::Config.project}")
+      FileUtils.mkdir_p(latest_filepath)
+
+      version = Pkg::Util::Version.get_dot_version
+      latest_filename = File.join(latest_filepath, "LATEST")
+      File.open(latest_filename, 'w') { |file| file.write(version) }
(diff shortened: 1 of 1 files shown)


tasks/nightly_repos.rake
```diff
@@ -317,5 +317,18 @@ namespace :pl do
       Pkg::Util::RakeUtils.invoke_task("pl:jenkins:prepare_signed_repos", target_host, 'signed', 'version')
       Pkg::Util::RakeUtils.invoke_task("pl:jenkins:deploy_signed_repos_to_s3", target_bucket)
     end
+
+    task :update_release_versions, [:target_bucket] do |t, args|
+      target_bucket = args.target_bucket or fail ":target_bucket is a required argument to #{t}"
+
+      tempdir = Pkg::Util::File.mktemp
```
melissa: should we clean this up after we're done? I feel like we never reboot our systems, and things just build up in the tmpdir...

tasks/nightly_repos.rake
```diff
@@ -317,5 +317,19 @@ namespace :pl do
...
+      latest_filepath = File.join(tempdir, "pkg", "#{Pkg::Config.project}")
+      FileUtils.mkdir_p(latest_filepath)
+
+      version = Pkg::Util::Version.get_dot_version
```
melissa: I just want to point out that the version returned with this utility does not always correspond to the package version. If the vanagon project is versioned with a tag, we should be fine. PDK is, so we should be fine. But for other projects, like puppet-enterprise-vanagon, we hard code the version into the project file. In those cases, the version in the package name and `Pkg::Util::Version.get_dot_version` are inconsistent

tasks/nightly_repos.rake
demophoon: Yeah, good call. I've updated the function to clear out that directory after we do the s3 sync.

tasks/nightly_repos.rake
demophoon: Is there a better, more consistent way to get the version of a package in packaging? I just realized that using the `Pkg::Util::Version.get_dot_version` on pdk will not work as the package version has an extra '.0' at the end and the tag in pdk is 'v0.6.0' instead of '0.6.0.0'

shrug: We should probably just be tagging the package version - we use 'git describe' all over the place

PR #707, opened 2017-08-15
Maint/1.0.x/a few improvements

melissa:


tasks/sign.rake
```diff
@@ -67,9 +67,9 @@ namespace :pl do
 
   desc "Detach sign any solaris svr4 packages"
   task :sign_svr4 do
-    unless Dir["pkg/solaris/10/**/*.pkg.gz"].empty?
+    unless Dir["pkg/**/*.pkg.gz"].empty?
```
underscorgan: This will end up including the signing bundle

melissa: Nope, that's `.tar.gz`

melissa:
```
$ find pkg -type f -name *.gz | sort
pkg/puppet-agent-5.0.1-15-g921d464-signing_bundle.tar.gz
```

underscorgan: d'oh. i typoed it when I was testing.

melissa: thank you for testing <3

lib/packaging/util/net.rb
```diff
@@ -293,21 +293,31 @@ def remote_set_immutable(host, files)
...
+      ls_cmd << '| sort --version-sort | tail -1'
+      cmd = <<-CMD
+        if [ ! -d '#{dir}' ] ; then
+          echo "directory '#{dir}' does not exist, not creating latest package link"
+          exit 0
+        fi
+        pushd '#{dir}'
+        link_target=$(#{ls_cmd})
+        if [ -z "$link_target" ] ; then
+          echo "Unable to find a link target for '#{package_name}' in '#{dir}'; skipping link creation"
```
melissa: This error message never actually is reached if the `ls` fails

```
Executing 'set -e;         if [ ! -d '/opt/downloads/windows' ] ; then
          echo "directory '/opt/downloads/windows' does not exist, not creating latest package link"
          exit 0
        fi
        pushd '/opt/downloads/windows'
        link_target=$(ls -1 *.msi | grep -v latest | grep -v rc | grep puppet-agent | grep x86| sort --version-sort | tail -1)
        if [ -z "$link_target" ] ; then
          echo "Unable to find a link target for 'puppet-agent-x86' in '/opt/downloads/windows'; skipping link creation"
          exit 0
        fi
        echo "creating link to '$link_target'"
        ln -sf "$link_target" puppet-agent-x86-latest.msi
' on eyncwxwu533kkk0.delivery.puppetlabs.net
Pseudo-terminal will not be allocated because stdin is not a terminal.
ls: cannot access *.msi: No such file or directory
```

melissa: But at least it fails....

PR #708, opened 2017-08-15
Maint/1.0.x/a few more improvements

melissa:


lib/packaging/paths.rb
```diff
@@ -59,23 +59,88 @@ def repo_name
...
+    end
+  end
+
+  def artifacts_base_path_and_link_path(platform_tag, path_prefix = 'artifacts')
+    platform, version, architecture = Pkg::Platforms.parse_platform_tag(platform_tag)
+    package_format = Pkg::Platforms.package_format_for_tag(platform_tag)
+
+    case package_format
+    when 'rpm'
+      if repo_name == 'PC1'
```
e-gris: Rather than encode PC1 hackery multiple times into the file, do you think it would make more sense to isolate it into its own encapsulated function?

Old school:
```ruby
def path_assemble(...)
  if repo_name == 'PCI'
    File.join(<yada>)
  else
    File.join(<yada>
end
```
then
```ruby
when 'rpm'
  path_assemble(path_prefix, platform, version, repo_name, link_name, platform)
```

There's also an uber-cool solution likely using partial functions that I would have to work through.

Not sure of the value of doing it super-slick rather than just getting it done in this case.

melissa: I'm not sure the best answer here. I'd almost rather just get it done, even though it's ugly. But if you do have a strong opinion, I'm glad to update it

melissa: Also, I'm not wholly sure if I follow you here. Are you saying I should add an individual `path_assemble` method for each repo type that we support?

e-gris: I don't have enough context yet to have a strong opinion. I'm not sure how long the PC1 stuff will be around nor do I have a big picture on the packaging code. So, "get-it-done" is the strongest argument at the moment.

lib/packaging/util/platform.rb
```diff
@@ -25,7 +25,7 @@ def get_attribute(platform_tag, attribute_name)
 
   def artifacts_path(platform_tag, package_url = nil)
```
melissa: So, I can't find an instance where we call this method and pass in `package_url`... I don't know if I've just removed all of them already, or if this method was used in enterprise-dist or some jenkins jobs. I searched ent-dist, and I did a brief look at jenkins jobs, and I didn't see anything....

melissa: Oh right, this was only used for mac packages

```
      when 'dmg'
        # We don't consistently ship OSX artifacts to the same path
        # vanagon ships things under a version number, and standard
        # packaging excludes the version number. We should fix that, but in
        # the interim we can check whether or not the versioned path exists
        # and fail back to the unversioned path if needed.
        version_path = File.join('artifacts', 'apple', version)
        if package_url.nil?
          version_path
        else
          code = Pkg::Util::Net.uri_status_code("#{package_url}/#{version_path}")
          if code == '200'
            version_path
          else
            File.join('artifacts', 'apple')
          end
        end
```
I think this is a safe change to make

tasks/ship.rake
```diff
@@ -372,7 +372,20 @@ namespace :pl do
 
   desc "ship apple dmg to #{Pkg::Config.dmg_staging_server}"
   task ship_dmg: 'pl:fetch' do
-    Pkg::Util::Ship.ship_pkgs(['pkg/**/*.dmg'], Pkg::Config.dmg_staging_server, Pkg::Config.dmg_path, addtl_path_to_sub: '/mac')
```
mwaggett: looks like we could remove the `addtl_path_to_sub` option from that method. I don't think it's used anywhere else and I kind of just find it confusing.

melissa: Yes, we could. I chose to leave it in to minimize the change set here

lib/packaging/util/ship.rb
```diff
@@ -54,11 +54,6 @@ def reorganize_packages(pkgs, tmp, platform_independent = false)
   # @option opts [Array] :excludes File globs to exclude packages from shipping
   # @option opts [Boolean] :chattr Whether or not to make the files immutable
   #   after shipping. Defaults to true.
-  # @option opts [String] :addtl_path_to_sub An additional part of the path
```
melissa: @mwaggett better?

mwaggett: yes 😊

PR #710, opened 2017-08-18
(RE-9133) Require a version be passed to release a package

demophoon:
This commit allows packages to be released by passing in a version to
the update_release_versions task instead of having packaging get this
package from the repo it is in. This needs to be done because tagging is
inconsistent across the various projects we have and our automation
should be able to pass versions down through jenkins jobs.

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -318,14 +318,14 @@ namespace :pl do
       Pkg::Util::RakeUtils.invoke_task("pl:jenkins:deploy_signed_repos_to_s3", target_bucket)
     end
 
-    task :update_release_versions, [:target_bucket] do |t, args|
-      target_bucket = args.target_bucket or fail ":target_bucket is a required argument to #{t}"
+    task :update_release_versions do
+      target_bucket = ENV['TARGET_BUCKET'] or fail "TARGET_BUCKET must be specified to run the 'update_release_versions' task"
+      version = ENV['VERSION'] || Pkg::Util::Version.get_dot_version
 
       tempdir = Pkg::Util::File.mktemp
       latest_filepath = File.join(tempdir, "pkg", "#{Pkg::Config.project}")
       FileUtils.mkdir_p(latest_filepath)
 
-      version = Pkg::Util::Version.get_dot_version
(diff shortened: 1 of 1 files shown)


demophoon:
I've created two jobs for this new task on jenkins-compose:
http://jenkins-compose.delivery.puppetlabs.net/job/Release-to-S3/
http://jenkins-compose.delivery.puppetlabs.net/job/puppetlabs-pdk_s3-release/

tasks/nightly_repos.rake
```diff
@@ -318,14 +318,14 @@ namespace :pl do
       Pkg::Util::RakeUtils.invoke_task("pl:jenkins:deploy_signed_repos_to_s3", target_bucket)
     end
 
-    task :update_release_versions, [:target_bucket] do |t, args|
+    task :update_release_versions, [:target_bucket, :version] do |t, args|
       target_bucket = args.target_bucket or fail ":target_bucket is a required argument to #{t}"
+      version = args.version or fail ":version is a required argument to #{t}"
```
melissa: I'm wondering if we should allow version to be passed in, but default to `Pkg::Util::Version.get_dot_version` if it's not there? Also, in the past we've generally used environment variables for things like this, as that's super easy to pass in for jenkins jobs. Any reason not to do that with this?

demophoon: We can definitely do that here. Thats updated now.

tasks/nightly_repos.rake
```diff
@@ -318,14 +318,14 @@ namespace :pl do
       Pkg::Util::RakeUtils.invoke_task("pl:jenkins:deploy_signed_repos_to_s3", target_bucket)
     end
 
-    task :update_release_versions, [:target_bucket] do |t, args|
-      target_bucket = args.target_bucket or fail ":target_bucket is a required argument to #{t}"
+    task :update_release_versions do
+      target_bucket = ENV['TARGET_BUCKET'] or fail "TARGET_BUCKET must be specified to run the 'update_release_versions' task"
+      version = ENV['VERSION'] || Pkg::Util::Version.get_dot_version
```
melissa: I have no idea what the "best practice" here is, but this is what I'm used to....

PR #712, opened 2017-08-25
(maint) Change AIX arches from 'power' to 'ppc'

mwaggett:
AIX artifacts have 'ppc' as the arch, so we encountered issues when mapping artifact paths to platform tags.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -80,7 +80,8 @@ module Pkg
           artifacts = artifacts.split("\n")
           data = {}
           Pkg::Util::Platform.platform_tags.each do |tag|
-            _, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            platform, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            arch = 'ppc' if platform == 'aix'
             package_format = Pkg::Util::Platform.get_attribute(tag, :package_format)
             case package_format
             when 'deb'


mwaggett:
Test build here: http://builds.delivery.puppetlabs.net/puppet-agent/4bc5d4ce8d0c646ec6507cb66873ee09df837af3/artifacts/

I plan on adding tests when I merge this up into 1.0.x, but because of how we restructured things, there isn't a great place for them on this branch.

melissa:
This may not be a safe change to make. This will change how the package gets staged, and the path different tools have to traverse to access the packages

melissa:
It will likely break old school beaker, pe_repo, jar-jar, etc

mwaggett:
@melissa How will this change the paths? They already use ppc because that's what puppet-agent uses: http://builds.delivery.puppetlabs.net/puppet-agent/1.10.6/artifacts/aix/6.1/PC1/ppc/

And is there a way to test what this breaks (besides merging and seeing what happens)? I can fix this a different way, but this feels the cleanest if we can get away with it.

melissa:
there's not a good way other than merging it in. I'm not willing to do that on a friday right before you leave for a week. I don't think the sudden disruption to all of the aix pipelines would be worth it either. That sounds like not a very clean way to fix this. You need to dig around and find out what exactly the consequences of this change will be. What will change in how we build and stage aix packages because of it. What pieces of data do we make available that other tools use and will expect `power` to be set?

mwaggett:
yeah, I wasn't suggesting that we actually just merge this and see what happens, just wondering what options were. I will do some digging.

mwaggett:
Decided to solve this a different way. It may be less elegant, but since 'power' is technically the name of the architecture, it feels better to not rename it (plus the bonus of not breaking things that depend on this).

Test build here: http://builds.delivery.puppetlabs.net/puppet-agent/e3fd0bf21d0baa0a84548a06575d272cf96e12cb/artifacts/

PR #713, opened 2017-08-29
(maint) Add test cases for AIX artifact paths

mwaggett:
AIX artifacts have 'ppc' as the arch, so we encountered issues when mapping artifact paths to platform tags. This adds AIX test cases to ensure we can figure out the platform tags correctly.


mwaggett:
Apparently this was never broken on the 1.0.x branch? http://builds.delivery.puppetlabs.net/puppet-agent/3f2861295786330e0a7f50187cb5558959d2fc2f/artifacts/
This ensures we have more thorough test coverage though.

PR #714, opened 2017-08-29
(maint) Fix package:deb target

smortex:
With no particular environment / setup, creating .deb packages in a puppetlabs project with…

```
rake package:bootstrap
rake package:deb
```

…fails with `TypeError: no implicit conversion of nil into String`.

This is due to the `codename` variable being *nil* (because `cow` is *nil*), which `File#join` does not accept as a valid argument.

Rely on `#to_s` so that this value is promoted to a valid `''` when *nil*.


shrug:
This looks entirely reasonable, thanks!

PR #716, opened 2017-09-07
1.0.x

melissa:


lib/packaging/config.rb
```diff
@@ -81,6 +81,8 @@ def platform_data
           data = {}
           artifacts.each do |artifact|
             tag = Pkg::Paths.tag_from_artifact_path(artifact)
+            platform, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
+            arch = 'ppc' if platform == 'aix'
```
mwaggett: I think these two lines are unnecessary with how we implemented this method on this branch.

mwaggett: aka you can exclude the 'Change AIX arches . . .' commit

melissa: I'd rather not deviate too much from the branches. If it isn't going to break anything, we should leave this as is to make further merge ups from master easier

mwaggett: I don't think this will break anything, since `arch` isn't used anywhere, but these methods are already pretty different between branches, so I don't see the benefit of including this.
https://github.com/puppetlabs/packaging/compare/master...1.0.x#diff-9c94054f5e25a311dcb8254488c7dfe4

melissa: Every time we do this, have changes in master that aren't carried forward to this branch, merge ups get shittier. We can have this method be different in 1.0.x, and moving forward it should be able to figure out how to handle future merge ups. But only if we keep the change set consistent. If it's not going to break anything, we should keep this change.

melissa: Generally speaking, when doing a merge up, you should never change the code except to resolve merge conflicts. Everything else should be untouched.

underscorgan: Yeah, in general if development is happening in a divergent manner but also in parallel, I tend to explicitly add commits in the 'newer' (1.0.x here) removing the unwanted code changes so git will know how to handle those in future mergeups. I feel like since this case isn't going to break anything it isn't a big deal, but can always be changed later with a cleanup commit.

PR #720, opened 2017-09-28
(maint) Handle an empty nonfinal gpg key string

melissa:


underscorgan:
This looks like a reasonable fix for this issue I think.

I'm mostly concerned about what will happen on promotion/internal ship of a non-final version in a repo with a non-final key defined.

melissa:
that is a valid concern. We decided not to use a nonfinal gpg key, so that isn't currently set anywhere. But I can also pull out all the gpg nonfinal logic

underscorgan:
I think we can just fix this now and maybe ticket removing all of that stuff (or do it in a separate PR)

melissa:
Sounds like a lovely plan, RE-9422 opened for that work

lib/packaging/util/gpg.rb
```diff
@@ -7,7 +7,7 @@ class << self
     # files that are generated with this repo use the default gpg key to
     # reflect that.
     def key
-      if !Pkg::Util::Version.final? && Pkg::Config.gpg_nonfinal_key
+      if !Pkg::Util::Version.final? && Pkg::Config.gpg_nonfinal_key && !Pkg::Config.gpg_nonfinal_key.empty?
```
e-gris: Shouldn't Pkg::Config.gpg_nonfinal_key return something falsey when its empty rather than expect its consumer to check?

melissa: It's just a string, so it's not that smart. That would be ideal though. We'll be ripping this method out, and simply calling `Pkg::Config.gpg_key` in the future though

PR #722, opened 2017-09-28
(maint) Pkg::Config is not available when we load deb/repo.rb

melissa:
We can't access Pkg::Config.apt_repo_name, or any of the other variables
from Pkg::Config, because they haven't been loaded yet. Once we actually
call the method, then they have been loaded and we can use them.


lib/packaging/deb/repo.rb
```diff
@@ -59,6 +44,12 @@ def generate_repo_configs(source = "repos", target = "repo_configs")
       # We want to exclude index and robots files and only include the http: prefixed elements
       repo_urls = stdout.split.uniq.reject { |x| x =~ /\?|index|robots/ }.select { |x| x =~ /http:/ }.map { |x| x.chomp('/') }
 
+      if Pkg::Config.apt_repo_name
+        reprepro_repo = Pkg::Config.apt_repo_name
+      else
+        reprepro_repo = Pkg::Paths.repo_name.empty? ? 'main' : Pkg::Paths.repo_name
+      end
+
```
underscorgan: Can this be broken out into a method so we can call that instead of copying the logic?

mwaggett: +1

melissa: done

lib/packaging/deb/repo.rb
```diff
@@ -11,20 +11,12 @@ class << self
...
-      DEBIAN_REPO_NAME = Pkg::Config.apt_repo_name
-    else
-      DEBIAN_REPO_NAME = Pkg::Paths.repo_name.empty? ? 'main' : Pkg::Paths.repo_name
-    end
+    def reprepro_repo_name
+      if Pkg::Config.apt_repo_name
+        Pkg::Config.apt_repo_name
+      else
+        Pkg::Paths.repo_name.empty? ? 'main' : Pkg::Paths.repo_name
+      end
```
mwaggett: missing `end`

lib/packaging/deb/repo.rb
```diff
@@ -11,19 +11,12 @@ class << self
...
-      DEBIAN_REPO_NAME = Pkg::Config.apt_repo_name
-    else
-      DEBIAN_REPO_NAME = Pkg::Paths.repo_name.empty? ? 'main' : Pkg::Paths.repo_name
+    def reprepro_repo_name
+      if Pkg::Config.apt_repo_name
+        Pkg::Config.apt_repo_name
+      else
+        Pkg::Paths.repo_name.empty? ? 'main' : Pkg::Paths.repo_name
+      end
     end
```
mwaggett: there it is

PR #723, opened 2017-09-28
(RE-9422) Remove the ability to pass in a nonfinal gpg key

mwaggett:
This reverts commit 69925ee16d41990ee2017e4df6e12eb1ac379bce: "(RE-8727) Add support to define a nonfinal gpg key" because we decided that we want to sign everything with the final key.


spec/lib/packaging/util/gpg_spec.rb
```diff
@@ -1,37 +1,16 @@
...
 describe "Pkg::Util::Gpg" do
-  let(:gpg)               { "/local/bin/gpg" }
-  let(:keychain)          { "/usr/local/bin/keychain" }
-  let(:gpg_key)           { "abcd1234" }
-  let(:gpg_nonfinal_key)  { "xyz9876" }
-  let(:target_file)       { "/tmp/file" }
+  let(:gpg)      { "/local/bin/gpg" }
+  let(:keychain) { "/usr/local/bin/keychain" }
+  let(:gpg_key)  { "abcd1234" }
+  let(:target_file) { "/tmp/file" }
```
e-gris: Very minor, feel free to ignore if the impact is too significant:  tests should do their scratch work in the current directory rather than use system /tmp. Making $CWD/tmp and using $CWD/tmp/file would be better practice.

mwaggett: I think that makes sense, but is something for a different PR. This is just reverting previous work.

lib/packaging/util/gpg.rb
```diff
@@ -1,19 +1,6 @@
 module Pkg::Util::Gpg
   class << self
 
-    # Please note that this method is not used in determining what key is used
-    # to sign the debian repos. That is defined in the freight config that
-    # lives on our internal repo staging host. The debian conf/distribution
-    # files that are generated with this repo use the default gpg key to
-    # reflect that.
-    def key
```
melissa: I kind of like keeping this here, that way we can add a check and fail if we're trying to do anything and `Pkg::Config.gpg_key` isn't set. It'd also allow us to add some tests to do some simple validation

mwaggett:
@melissa updated. I didn't add tests or validation because 1. that's not the purpose of this PR and 2. I wasn't entirely sure what you had in mind.

melissa:
Given the state of the packaging repo, we should be adding tests and validations with every pr, so don't let that limit you. And basic validation that it's set before we attempt to start signing is sufficient.

Also, we shouldn't merge this in while we are still shipping things to the nightly repos on apt.puppetlabs.com and yum.puppetlabs.com, as those repos do require the use of the nightly gpg key.

spec/lib/packaging/util/gpg_spec.rb
```diff
@@ -1,34 +1,20 @@
...
-      expect(Pkg::Util::Gpg.key).to eq(gpg_key)
-    end
-
-    it 'uses the nonfinal key if building a nonfinal version' do
-      allow(Pkg::Config).to receive(:version).and_return('1.0.0-658-gabc1234')
-      allow(Pkg::Config).to receive(:gpg_nonfinal_key).and_return(gpg_nonfinal_key)
-      expect(Pkg::Util::Gpg.key).to eq(gpg_nonfinal_key)
+    it "fails if Pkg::Config.gpg_key isn't set" do
+      allow(Pkg::Config).to receive(:gpg_key).and_return(nil)
+      expect { Pkg::Util::Gpg.key }.to raise_error(RuntimeError)
```
mwaggett: @melissa Is this a sufficient test?

mwaggett:
@melissa Is the nightly gpg set in pipelines somewhere? because I didn't see it in puppet-agent or build-data.

melissa:
No, the nightly key shouldn't be set anywhere in any pipelines. We've only been using it manually when shipping to apt.pl.com and yum.pl.com nightly repos

spec/lib/packaging/util/gpg_spec.rb
melissa: what if it's set, but an empty string?

spec/lib/packaging/util/gpg_spec.rb
```diff
@@ -1,34 +1,24 @@
...
-      expect(Pkg::Util::Gpg.key).to eq(gpg_key)
-    end
-
-    it 'uses the nonfinal key if building a nonfinal version' do
-      allow(Pkg::Config).to receive(:version).and_return('1.0.0-658-gabc1234')
-      allow(Pkg::Config).to receive(:gpg_nonfinal_key).and_return(gpg_nonfinal_key)
-      expect(Pkg::Util::Gpg.key).to eq(gpg_nonfinal_key)
+    it "fails if Pkg::Config.gpg_key is an empty string" do
+      allow(Pkg::Config).to receive(:gpg_key).and_return('')
+      expect { Pkg::Util::Gpg.key }.to raise_error(RuntimeError)
```
mwaggett: @melissa updated

shrug:
Wait until we stop shipping manual nightlies

PR #724, opened 2017-09-29
(maint) Fix versioning

melissa:


melissa:
@geoffnichols @McdonaldSeanp @caseywilliams I believe this was the origin of the versioning issues in the promotion

PR #726, opened 2017-10-02
(RE-9299) Ship to artifactory

melissa:


lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,105 @@
+module Pkg::Platforms
+  module_function
+
+  ARTIFACTORY_URL = 'http://artifactory.delivery.puppetlabs.net/artifactory'
```
e-gris: Use https

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,105 @@
...
+  CURL = 'curl'
+
+  default_repo = 'development'
+
+  def artifactory_authorization
+    #"--user #{user_name}:#{api_token}"
+    ''
```
e-gris: Will this be used by jenkins or individual users? If it's the former, then we can put Artifactory's api_token inside a Jenkins secret.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def debian_extras(platform_tag)
+    platform, version, arch = Pkg::Util::Platform.parse_platform_tag(platform_tag)
+    codename = Pkg::Platforms::PLATFORM_INFO[platform][version][:codename]
+    fail "Codename not found for #{platform_tag}; cannot push to correct Artifactory repo." unless codename
+    ";deb.distribution=#{codename};deb.component=#{default_repo};deb.architecture=#{architecture}"
+  end
+
+  # package: full path to the package relative to the current working directory
+  def artifactory_curl_command(package, platform_tag)
```
e-gris: I would take a different approach. Create a Factory that generates objects for each type of platform. The object should have accessor methods for build information about that platform.

It _might_ be worthwhile that rather than leaning on curl, using native ruby gems to do the calls.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,105 @@
...
+  #   * build_metadata.json
+  #   * signing bundle
+  #   * etc
+  # - Are we going to isolate the different builds like we have been on
+  #   builds.delivery.puppetlabs.net, or are all development builds going to be
+  #   shipped to the same repo?
+  # - Are we going to ship development builds to their targeted repo?
+  #   * i.e., PC1 vs. puppet5 vs pupppet8 vs etc
+  #   * This may result in duplicate builds if we ever need to ship one thing
+  #     to multiple repos, which is a thing we have done before
```
e-gris: Artifactory is smart enough to not store an object more than once. If the same file appears in multiple repos, then artifactory only stores the file once. In other words, yes, put the same file in multiple places if needed.

lib/packaging/artifactory.rb
melissa: Both. I assume this will be an environment variable, but I haven't put any time into the authentication yet.

melissa:
@e-gris assuming beaker is the primary consumer of these builds, how are you anticipating beaker will find and install the artifacts it needs?

lib/packaging/artifactory.rb
e-gris: In the case of a user: it should be read from a yaml/JSON file that is protected.
In the case of Jenkins: it can be passed using Jenkins mechanism for handling secrets. I would suggest simplifying the process by just dumping that into a protected yaml/JSON file and behaving like the 'user' case.

e-gris:
I would hope that it would use the same code that generated the deploy. To be really abstract:

Deploy:

```ruby
  deploy = ArtifactoryPackageGem.location_info({:package => 'mypackage', :version => '1.7.4.2-679-g3bee7fb', file => path/to/package-file})
  deploy.deploy
```

Retrieve:

```ruby
  retrieve = ArtifactoryPackageGem.location_info({:package => 'mypackage', :version => '1.7.4.2-679-g3bee7fb')
  retrieve.retrieve
```

melissa:
Beaker doesn't load any automation that lives in beaker, so we won't be able to take advantage of this code when installing via beaker

e-gris:
To put it all together.

1. There should be a new gem that knows how to put things into artifactory and get things out of artifactory.

1. Each type of thing that can be stored and retrieved should be defined declaratively (I'd use yaml but have no strong attachments), not in actual ruby code.

1. The declarative definitions can live inside the gem. It might become useful to be able to update the definitions without revving the gem but I suggest deferring that.

1. The methods in the gem interpret the static code and create objects containing accessors needed for that particular package, version, package type, etc.

melissa:
I'm retargeting this work at the 1.0.x branch of artifactory. We should be able to move everything over to this branch, and it's way easier to do this work with the new utilities.

melissa:
Replaced with https://github.com/puppetlabs/packaging/pull/727

PR #727, opened 2017-10-03
(RE-9299) Deploy and retrieve packages from Artifactory

melissa:


lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,120 @@
...
+  # In this context, packaging_platform == platform_tag, the first is how
+  # beaker labels this variable, the second is how the packaging repo labels
+  # this variable.
+  #
+  # If we are looking for a platform independent package, like a gem or a
+  # tarball, we can set packaging_platform to that instead
+  def location_for(project, project_version, packaging_platform = nil)
+    default_subdirectory_base = File.join(REPO_BASE, project, project_version)
+
+    # This is only allowed for platform independent packages, like gems or tars
```
e-gris: Style remark. Remove the "else"

```ruby
toplevel_repo = 'generic'
repo_subdirectories = default_subdirectory_base
unless  packaging_platform.match(/(gem|tar|tgz|tar\.gz|json|yaml$)/)
  ... lookup work
end
```

I still would consider further abstraction...

``` ruby
ALLOWED_PACKAGES = %w[rpm deb gem tar ...]
def initialize(project, project_version, packaging_platform = nil)
  if packaging_platform in ALLOWED_PACKAGES
    repo_object = Object::const_get("Package::Platforms::#{packaging_platform}").new
  end
end
```

Then later, I would expect this to work:

```ruby
artifactory_repo = Package::Platforms("some-project", "some-version", "rpm")
print artifactory_repo.toplevel_repo
```

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,129 @@
...
+    ARTIFACTORY_URL = 'https://artifactory.delivery.puppetlabs.net/artifactory'
+    CURL = 'curl'
+    REPO_BASE = 'development'
+
+    # @param project [String] The name of the project this package is for
+    # @param project_version [String] The version of the project we want the
+    #   package for. This can be one of three things:
+    #     1) the final tag of the project the packages  were built from
+    #     2) the long git sha the project the packages were built from
+    #     3) the ezbabke generated development sha where the packages live
```
underscorgan: [typo] this is really making me want some babke.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,129 @@
...
+    # @param platform_tag [String] Either the platform tag for the
+    #   platform we want deal with packages for (i.e., el-7-x86_64 or
+    #   ubuntu-16.04-amd64), or the package/archive type that is platform
+    #   independent (i.e., gem, tar, tar.gz, tgz, yaml, json)
```
underscorgan: [doc nit] I think I like 'generic' for platform_type, but if there are going to be more accepted values it should be super clear what the full set is.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,129 @@
...
+    #   * package name to install, i.e., puppet-agent, puppetserver, etc
+    #   * ether the long sha of project or "latest"
+    #
+    # In this context, packaging_platform == platform_tag, the first is how
+    # beaker labels this variable, the second is how the packaging repo labels
+    # this variable.
+    #
+    # If we are looking for a platform independent package, like a gem or a
+    # tarball, we can set packaging_platform to that instead
+    def location_for
```
underscorgan: It looks like this is only used during the initialize and it doesn't seem like it has much use other than setting the class variables, I think we should make this private.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,129 @@
+class Pkg
+  class Artifactory
+
+    ARTIFACTORY_URL = 'https://artifactory.delivery.puppetlabs.net/artifactory'
+    CURL = 'curl'
+    REPO_BASE = 'development'
```
underscorgan: Are these actually going to be globals, or do these make more sense as things optionally set in the initialize (maybe an options hash or something), with these as the defaults?

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,129 @@
...
+      if @package_format == 'deb'
+        ";deb.distribution=#{@codename};deb.component=#{@repo_subdirectories};deb.architecture=#{@architecture}"
+      else
+        ''
+      end
+    end
+
+    # package: full path to the package relative to the current working directory
+    def artifactory_deploy_package(package)
+      Pkg::Util::Execution.capture3("#{CURL} #{artifactory_authorization} --upload-file #{package} '#{ARTIFACTORY_URL}/#{@repo_name}/#{@repo_subdirectories}/#{File.basename(package)}#{curl_extras}'")
```
underscorgan: You might be able to use `Pkg::Util::Net.curl_form_data`, it doesn't save too much but at least you won't have to worry about figuring out curl here? :woman_shrugging:

melissa:
@tvpartytonight would you be willing to take a look at the rspec here?

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,158 @@
...
+      when 'rpm'
+        toplevel_repo = 'rpm'
+        repo_subdirectories = File.join(repo_subdirectories, "#{@platform}-#{@platform_version}-#{@architecture}")
+      when 'deb'
+        toplevel_repo = 'debian__local/pool'
+      when 'swix', 'dmg', 'svr4', 'ips'
+        repo_subdirectories = File.join(repo_subdirectories, "#{@platform}-#{@platform_version}-#{@architecture}")
+      when 'msi'
+        repo_subdirectories = File.join(repo_subdirectories, "#{@platform}-#{@architecture}")
+      end
```
melissa: @e-gris how consistent are these top level repos going to be? Should be allow them to be overwritten, or are they going to be consistent enough that hard coding them here is acceptable?

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,163 @@
...
+        DOC
+      else
+        ''
+      end
+    end
+
+    # @return [String] The curl authorization require to access the artifactory
+    #   repos
+    #
+    # TODO this needs to be smoothed over and actually though about
```
melissa: Note to self, don't forget about this.

melissa:
So, I think this is just about there. I have a few more bad assumptions to work though though.

```
$ rake pl:jenkins:ship_to_artifactory
Skipping platform_data collection, but don't worry about it.

    DEPRECATED, 29-Jul-2014: 'gpg_name' has been replaced with 'gpg_key'.
                   Please update this field in your build_defaults.yaml
touch /home/melissa/.packaging
remote: Counting objects: 714, done.
remote: Compressing objects: 100% (17/17), done.
remote: Total 714 (delta 5), reused 19 (delta 4), pack-reused 693
Receiving objects: 100% (714/714), 104.18 KiB | 0 bytes/s, done.
Resolving deltas: 100% (179/179), done.
Switched to a new branch 'release'
Switched to a new branch 'puppet-agent'
rm -rf /tmp/pkgt0pA9g
rake aborted!
I can't find a codename or platform in pkg/f65f9efbb727c3d2d72d6799c0fc345a726f27b5.yaml, teach me?
/home/melissa/puppet-agent/ext/packaging/lib/packaging/paths.rb:34:in `tag_from_artifact_path'
/home/melissa/puppet-agent/ext/packaging/tasks/ship.rake:606:in `block (4 levels) in <top (required)>'
/home/melissa/puppet-agent/ext/packaging/tasks/ship.rake:605:in `each'
/home/melissa/puppet-agent/ext/packaging/tasks/ship.rake:605:in `block (3 levels) in <top (required)>'
Tasks: TOP => pl:jenkins:ship_to_artifactory
(See full trace by running task with --trace)
```

lib/packaging/paths.rb
```diff
@@ -31,9 +31,23 @@ def tag_from_artifact_path(path)
...
+      end
       platform, version = Pkg::Platforms.codename_to_platform_version(codename)
-      fail "I can't find a platform and version from #{codename}, teach me?" if platform.nil? || version.nil?
+      if platform.nil? || version.nil?
+        if fail_hard
+          fail "I can't find a platform and version from #{codename}, teach me?"
+        else
+          warn "I can't find a platform and version from #{codename}, teach me?"
+          return nil
+        end
```
melissa: I really don't like this. I had done it in a much more elegant way, but that required a lot of additional changes that I didn't feel comfortable making in this PR.

melissa:
```
$ rake pl:jenkins:ship_to_artifactory --trace
Skipping platform_data collection, but don't worry about it.

    DEPRECATED, 29-Jul-2014: 'gpg_name' has been replaced with 'gpg_key'.
                   Please update this field in your build_defaults.yaml
** Invoke pl:jenkins:ship_to_artifactory (first_time)
** Execute pl:jenkins:ship_to_artifactory
** Invoke pl:fetch (first_time)
** Execute pl:fetch
touch /home/melissa/.packaging
remote: Counting objects: 714, done.
remote: Compressing objects: 100% (17/17), done.
remote: Total 714 (delta 5), reused 19 (delta 4), pack-reused 693
Receiving objects: 100% (714/714), 104.18 KiB | 0 bytes/s, done.
Resolving deltas: 100% (179/179), done.
Switched to a new branch 'release'
** Invoke pl:load_extras (first_time)
** Execute pl:load_extras
Switched to a new branch 'puppet-agent'
** Invoke pl:load_extras (first_time)
** Execute pl:load_extras
rm -rf /tmp/pkgwrJXb4
I can't find a codename or platform in pkg/f65f9efbb727c3d2d72d6799c0fc345a726f27b5.yaml, teach me?
I can't find a codename or platform in pkg/f65f9efbb727c3d2d72d6799c0fc345a726f27b5.build_metadata.json, teach me?
I can't find a codename or platform in pkg/puppet-agent-5.3.1-34-signing_bundle.tar.gz, teach me?
```

melissa:
One thing I did notice, is that if you ship something that already exists in the repo, it'll simply overwrite the old package without throwing an error. Is that something we're concerned about?

underscorgan:
@melissa that seems like something we'd want to prevent if possible

melissa:
The other thing is that I used the curl api rather than loading in the artifactory gem, which is totally an option https://github.com/chef/artifactory-client

tasks/ship.rake
```diff
@@ -598,6 +598,17 @@ namespace :pl do
   # server path. That way we can separate out built artifacts from
   # signed/actually shipped artifacts e.g. $path/shipped/ or $path/artifacts.
   namespace :jenkins do
+    desc 'ship pkg directory contents to artifactory'
+    task :ship_to_artifactory, :local_dir do |_t, args|
+      Pkg::Util::RakeUtils.invoke_task('pl:fetch')
+      local_dir = args.local_dir || 'pkg'
+      Dir.glob("#{local_dir}/**/*").reject { |e| File.directory? e }.each do |artifact|
```
mwaggett: @e-gris Don't you have feelings about globs?

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+
+      @repo_name, @repo_subdirectories = location_for
+    end
+
+    # @return [Array] An array containing two items, first being the main repo
+    #   name for the platform_tag, the second being the subdirectories of the
+    #   repo leading to the artifact we want to install
+    def location_for
+      toplevel_repo = 'generic'
```
mwaggett: Can we differentiate between the use of 'generic' for artifacts without platform tags and the use of 'generic' as a repo name? I imagine this being confusing. since it's already kind of confusing for me right now and I think it's generally a good idea to not overload things.

tasks/ship.rake
e-gris: I do. They are a code smell and indicate something broken in the *design*. Unfortunately, resolving that design flaw can be *really hard* but worthwhile IMO.

lib/packaging/artifactory.rb
e-gris: Top level repos are about as set as anything. So, hard-coding is acceptable.

See above comment about 'generic' repo.

lib/packaging/artifactory.rb
e-gris: @mwaggett I'm happy to create a different repo name for generic if that will help.

lib/packaging/paths.rb
mwaggett: I agree that this doesn't really seem ideal.. Perhaps this is worth some team discussion?

tasks/ship.rake
melissa: We're grabbing everything in the directory. I would understand if I was trying to exclude certain things, but is there an issue leaving this as is, since it's purposefully everything in the directory that isn't itself a directory?

lib/packaging/paths.rb
mwaggett: I'm assuming that your only reason for this is so you can set the platform tag to 'generic' in the ship task.

lib/packaging/artifactory.rb
melissa: they kind of are the same thing. Artifacts that aren't tied to a platform tag are both generic packages, and are sent to live in the generic repo.

lib/packaging/paths.rb
mwaggett: I think we need a way to distinguish between expecting and not expecting to find a platform tag, e.g. .gem, .tar, etc -> all good, fedora-420-lmao -> not good.

lib/packaging/artifactory.rb
mwaggett: but like.. 'generic' used for gems vs 'generic' used for windows, e.g. I don't think those should be called the same thing..?

tasks/ship.rake
mwaggett: I agree with @e-gris. We should be doing some kind of validation that we're shipping the right things.

e-gris: I'm don't like the idea of letting automation just grab whatever-is-there and putting it blindly somewhere else. There ought to be, at least, some sort of white-listing. However, if there's a Jira for it, I'm OK with deferring it for later improvement.

mwaggett: I made this a while ago: https://tickets.puppetlabs.com/browse/RE-9309

melissa: @mwaggett that ticket is for uber_ship, this is different automation entirely

melissa: @e-gris as this only used for development builds that are well contained in their own subdirectories, I'm not sure it's worth the effort. Especially if we want to ensure we're keeping the build metadata files we've generated

mwaggett: yeah, but it's automation we could reuse in multiple places.

mwaggett: *should

melissa: Also, pkg is automatically populated by other automation, people shouldn't be randomly dropping things into that directory, which is another reason I'd prefer to leave it open

mwaggett: What about when we forget to clean out the pkg directory between ships?

e-gris: I see this as desirable but not critical. I'm OK with deferring to Melissa on this.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,163 @@
...
+      curl_uri = "#{@artifactory_url}/#{@repo_name}/#{@repo_subdirectories}/#{File.basename(package)}#{deploy_curl_extras}"
+      Pkg::Util::Net.curl_form_data(curl_uri, curl_opts)
+    end
+
+    # @param package [String] optional, the name of the package to be
+    #   retrieved. If the user does not know this information, we can derive it
+    #   from the yaml data. The user can set this either as just the package
+    #   name, or optionally include the path to where they want the package to
+    #   land.
+    # @param retrieval_path [String] Optional, an optional path set to where
```
mwaggett: This param name sounds like it's the path you're retrieving FROM. Maybe change to `output_path`?

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+    end
+
+    # @return [String] The curl authorization require to access the artifactory
+    #   repos
+    #
+    # TODO this needs to be smoothed over and actually though about
+    def authorization
+      # "--user #{user_name}:#{api_token}"
+      ''
```
mwaggett: Can this prompt you the way winston does?

melissa: I haven't had time to give this thought yet.

melissa:
@mwaggett this PR is about to change quiet a bit when I move it away from curl. Could you wait until then to comment?

mwaggett:
@melissa sure, sorry.

lib/packaging/artifactory.rb
mwaggett: I know, it was just a thought I was throwing out there. 😊

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,200 @@
...
+    end
+
+    # @return [String] Any required extra bits that we need for the curl
+    #   command used to deploy packages to artifactory
+    #
+    #   These are a few examples from chef/artifactory-client. These could
+    #   potentially be very powerful, but we should decide how to use them.
+    #     status: 'DEV',
+    #     rating: 5,
+    #     branch: 'master'
```
melissa: I believe @shrug had thoughts about setting certain properties when deploying artifacts. I'd like to get more input from him here.

If anyone else is curious, you can read about them at https://www.jfrog.com/confluence/display/RTF/Properties

tasks/ship.rake
melissa: This should only be called via a jenkins pipeline, so I'm not concerned about people forgetting to clean out the pkg directory.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,200 @@
...
+          'deb.architecture' => @architecture,
+        }
+      end
+      {}
+    end
+
+    # @param package [String] The full relative path to the package to be
+    #   shipped, relative from the current working directory
+    def deploy_package(package)
+      if @package_format == 'deb'
```
e-gris: My old favorites.

* Remove the `else` clause.
* Use constant strings.
* Don't abbreviate

```ruby
subdirectories = @repo_subdirectories
if @package_format == DEBIAN_PACKAGE_FORMAT ## or PACKAGE_FORMATS[:debian]
  subdirectories = File.join('pool', @repo_subdirectories)
end
```

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,200 @@
...
+    #   customization for where the user wants to fetch the package is via the
+    #   download_path parameter.
+    # @param download_path [String] Optional, an optional path set to where
+    #   the user wants the retrieved package to end up. If no path is specified
+    #   this defaults to the pkg directory.
+    def retrieve_package(package = nil, download_path = nil)
+      package ||= package_name
+      download_path ||= File.join('pkg', @repo_subdirectories)
+
+      if @package_format == 'deb'
```
e-gris: Code smell, duplicated code from above.

Think about encapsulating this logic with the intent of gathering other related stuff into a single place and using a single method call to set it.

Maybe something like

```ruby
subdirectories = Pkg::PackagingVariations.new(@package_format).subdirectories
```

lib/packaging/paths.rb
```diff
@@ -31,9 +31,23 @@ def tag_from_artifact_path(path)
     # if we didn't find a platform or a version, probably a codename
     if platform.nil? || version.nil?
       codename = Pkg::Platforms.codenames('deb').find { |c| path =~ /\/#{c}/ }
-      fail "I can't find a codename or platform in #{path}, teach me?" if codename.nil?
+      if codename.nil?
+        if fail_hard
+          fail "I can't find a codename or platform in #{path}, teach me?"
+        else
```
e-gris: Any reason for not always failing hard?

tasks/ship.rake
e-gris: I agree with @melissa on the last point. Discarding individual build info should be left entirely to Jenkins/whatever build system we end up with.

lib/packaging/paths.rb
```diff
@@ -21,7 +21,7 @@ def arch_from_artifact_path(platform, version, path)
 
   # Given a path to an artifact, divine the appropriate platform tag associated
   # with the artifact and path
-  def tag_from_artifact_path(path)
+  def tag_from_artifact_path(path, fail_hard = true)
```
melissa: I created https://tickets.puppetlabs.com/browse/RE-9510 to make this not dumb. It's going to be a big enough change it needs to be in a seperate PR

lib/packaging/paths.rb
melissa: Same as below. It's a poorly written method. I had updated this to handle this warn case with more grace, but that involved quiet a bit of change. I felt uncomfortable including all of that in this PR. I opened a new ticket to deal with that.

lib/packaging/artifactory.rb
e-gris: Also, any string that's used more than once and used in the _same_ context, like 'generic' should be put into a constant.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,199 @@
...
+
+    # @param package [String] optional, the name of the package to be
+    #   retrieved. If the user does not know this information, we can derive it
+    #   from the yaml data. This ignores everything but the package name. Any
+    #   customization for where the user wants to fetch the package is via the
+    #   download_path parameter.
+    # @param download_path [String] Optional, an optional path set to where
+    #   the user wants the retrieved package to end up. If no path is specified
+    #   this defaults to the pkg directory.
+    def retrieve_package(package = nil, download_path = nil)
```
melissa: actually, I didn't test this thoroughly, and I probably should...

Gemfile
```diff
@@ -1,5 +1,7 @@
 source "https://rubygems.org"
```
shrug: We should use artifactory here maybe? Use it as the primary source and keep rubygems as a secondary?

melissa: From http://bundler.io/man/gemfile.5.html
>It is possible, but not recommended as of Bundler 1.7, to add multiple global source lines. Each of these sources MUST be a valid Rubygems repository.

e-gris: IMO, it's a P1 service problem if Artifactory isn't delivering gems.

Gemfile
```diff
@@ -1,4 +1,17 @@
...
+#   2. For implicit gems (dependencies of explicit gems), any source, git, or
+#     path repository declared on the parent. This results in bundler
+#     prioritizing the ActiveSupport gem from the Rails git repository over ones
+#     from rubygems.org
+#   3. The sources specified via global source lines, searching each source in
+#     your Gemfile from last added to first added.
+
+source 'https://rubygems.org'
+source 'http://rubygems.delivery.puppetlabs.net'
+source 'https://artifactory.delivery.puppetlabs.net/artifactory/rubygems'
```
melissa: Pre this change:
```
$ time bi
The dependency win32console (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mingw32. To add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mingw32`.
Fetching gem metadata from https://rubygems.org/........
Fetching version metadata from https://rubygems.org/.
Resolving dependencies...
Using rake 0.9.6
Using artifactory 2.8.2
Using ast 2.3.0
Using bundler 1.15.4
Using coderay 1.1.2
Using diff-lcs 1.3
Using json 1.8.6
Using method_source 0.9.0
Using powerpack 0.0.9
Using rspec-core 2.14.8
Using rspec-mocks 2.14.6
Using ruby-progressbar 1.9.0
Using rainbow 2.2.2
Using parser 2.4.0.0
Using rspec-expectations 2.14.5
Using pry 0.11.2
Using rubocop 0.24.1
Using rspec 2.14.1
Bundle complete! 6 Gemfile dependencies, 18 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.

real    0m1.860s
user    0m1.581s
sys     0m0.209s
```

post this change:
```
$ time bi
The dependency win32console (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mingw32. T
o add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mingw32`.
Fetching source index from https://artifactory.delivery.puppetlabs.net/artifactory/rubygems/
Fetching source index from http://rubygems.delivery.puppetlabs.net/
Fetching source index from https://rubygems.org/
Resolving dependencies...
Using rake 0.9.6
Using artifactory 2.8.2
Using ast 2.3.0
Using bundler 1.15.4
Using coderay 1.1.2
Using diff-lcs 1.3
Using json 1.8.6
Using method_source 0.9.0
Using powerpack 0.0.9
Using rspec-core 2.14.8
Using rspec-mocks 2.14.6
Using ruby-progressbar 1.9.0
Using rainbow 2.2.2
Using parser 2.4.0.0
Using rspec-expectations 2.14.5
Using pry 0.11.2
Using rubocop 0.24.1
Using rspec 2.14.1
Bundle complete! 6 Gemfile dependencies, 18 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.

real    0m31.879s
user    0m30.782s
sys     0m0.569s
```

Gemfile
```diff
@@ -1,4 +1,15 @@
...
+# Source Priority (from http://bundler.io/man/gemfile.5.html#SOURCE-PRIORITY)
+#   1. The source explicitly attached to the gem (using :source, :path, or :git)
+#   2. For implicit gems (dependencies of explicit gems), any source, git, or
+#     path repository declared on the parent. This results in bundler
+#     prioritizing the ActiveSupport gem from the Rails git repository over ones
+#     from rubygems.org
+#   3. The sources specified via global source lines, searching each source in
+#     your Gemfile from last added to first added.
+source 'https://rubygems.org'
+source 'https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems'
```
melissa:
```
$ time bi
The dependency win32console (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mingw32. To add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mingw32`.
Fetching gem metadata from https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/.........
Fetching gem metadata from https://rubygems.org/........
Fetching version metadata from https://artifactory.delivery.puppetlabs.net/artifactory/api/gems/rubygems/..
Fetching version metadata from https://rubygems.org/.
Resolving dependencies...
Using rake 0.9.6
Using artifactory 2.8.2
Using ast 2.3.0
Using bundler 1.15.4
Using coderay 1.1.2
Using diff-lcs 1.3
Using json 1.8.6
Using method_source 0.9.0
Using powerpack 0.0.9
Using rspec-core 2.14.8
Using rspec-mocks 2.14.6
Using ruby-progressbar 1.9.0
Using rainbow 2.2.2
Using parser 2.4.0.0
Using rspec-expectations 2.14.5
Using pry 0.11.2
Using rubocop 0.24.1
Using rspec 2.14.1
Bundle complete! 6 Gemfile dependencies, 18 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.

real    0m2.574s
user    0m1.953s
sys     0m0.190s
```

Gemfile
melissa: Specifying artifactory as a source, even a secondary source, causes the travis tests to fail, since travis can't access internal infrastructure

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,204 @@
...
+    #   platform we want deal with packages for (i.e., el-7-x86_64 or
+    #   ubuntu-16.04-amd64), or 'generic' for those packages or archives
+    #   that are platform independent (i.e., tar, gem, etc)
+    # @option :artifactory_url [String] the uri for the artifactory server.
+    #   This currently defaults to 'https://artifactory.delivery.puppetlabs.net/artifactory'
+    # @option :repo_base [String] The base of all repos, set for consistency.
+    #   This currently defaults to 'development'
+    #
+    # rubocop:disable Metrics/AbcSize
+    def initialize(project, project_version, platform_tag = DEFAULT_REPO_NAME, opts = {})
```
melissa: I'm wondering if it's a bad idea to require the platform_tag when we initialize. We have to create a new object for each platform tag we want to handle. Is that reasonable? Or should we only initialize once, and call out to the same object for each platform tag we are dealing with?

mwaggett: I actually kind of like the idea of having one Artifactory object that knows about urls and paths and stuff, but that we pass it project, version, and platform tag for each thing we're trying to ship?
But I will say that my OO skills are kinda rusty at this point, so I don't have super strong feelings.

mwaggett: or rather ManageArtifactory object, now that I look a little closer

e-gris: Would it be worthwhile to have overloaded instantiators such that a missing platform_tag is permitted, but allows functionality to those methods that don't need it?

(I haven't read this far enough to understand the significance of the platform_tag)

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,226 @@
...
+      @full_artifactory_path = File.join(@repo_name, alternate_subdirectory_path)
+
+      Artifactory.endpoint = @artifactory_uri
+      check_authorization
+    end
+
+    # @return [Array] An array containing two items, first being the main repo
+    #   name for the platform_tag, the second being the subdirectories of the
+    #   repo leading to the artifact we want to install
+    def location_for(format = @package_format)
```
mwaggett: yardoc for the param 😊

lib/packaging/artifactory.rb
mwaggett: @e-gris The platform tag is used to determine the repo / directory structure we're shipping to.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,226 @@
...
+        repo_subdirectories = File.join(repo_subdirectories, "#{@platform}-#{@platform_version}-#{@architecture}")
+      when 'msi'
+        repo_subdirectories = File.join(repo_subdirectories, "#{@platform}-#{@architecture}")
+      end
+
+      [toplevel_repo, repo_subdirectories]
+    end
+
+    # @return [String] A string to the subdirectory. This is only truly needed
+    #   debian repos. When we deploy to artifactory, these repos need to be
```
mwaggett: missing the word 'for'

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,226 @@
...
+    #   do not need the 'pool' prefix.
+    def alternate_subdirectory_path
+      if @package_format == 'deb'
+        subdirectories = File.join('pool', @repo_subdirectories)
+      end
+      subdirectories || @repo_subdirectories
+    end
+
+    # Fetch the yaml file in order to parse the contents
+    def retrieve_yaml_data_file(tmpdir)
```
mwaggett: yardoc for param

mwaggett: I'm assuming that param is the desination dir, in which case, I feel like `dir` rather than `tmpdir` makes sense? but that's just a nitpick, so you can ignore.

e-gris: Warning: Huge Biased Opinion

tmpdir --> temporary_directory
dir --> directory

Concise, english words are easr to rd tn abbrs.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,226 @@
...
+      toplevel_repo, repo_subdirectories = location_for('yaml')
+      artifactory_repo_path = File.join(toplevel_repo, repo_subdirectories)
+      retrieve_package("#{@project_version}.yaml", tmpdir, artifactory_repo_path)
+    end
+
+    # @return [Hash] The data loaded from the retrieved yaml file for the
+    #   given project and version
+    def yaml_platform_data
```
mwaggett: What if we passed this function the file path of an already-downloaded file? I think that could make this function more multi-purpose.

lib/packaging/artifactory.rb
mwaggett: yeah, that's fine with me. I generally agree, although I feel like there are certain abbreviations that are so common that it's basically the same?
My real point was about temp vs. not in the param name. But again, that's pretty nitpicky.

lib/packaging/artifactory.rb
```diff
@@ -0,0 +1,226 @@
...
+        return {
+          'deb.distribution' => @codename,
+          'deb.component' => @repo_subdirectories,
+          'deb.architecture' => @architecture,
+        }
+      end
+      {}
+    end
+
+    # @param package [String] The full relative path to the package to be
```
mwaggett: I know @e-gris has feelings about relative paths vs absolute paths, but I'm not sure if that would make this work annoyingly difficult or not.

mwaggett: but just something to think about

e-gris: I have problems with using cd/chdir/pushd in scripts (that causes statefulness which is hard to debug and attracts other bugs). Relative paths are fine as long as the context is obvious because ./foo/bar  is as authorative as $TOPDIR/foo/bar as long as there are no cd shenanigans.

mwaggett: ahhh, gotcha, that makes sense!

lib/packaging/artifactory.rb
melissa: This is a private method, so it should only be used here and in the context of retrieving the yaml file from artifactory. I can't honestly see a reason to make it more generic

mwaggett: But you already have a different method for retrieving the file. I feel like there are many use-cases for parsing the platform_data in that file. You don't have to make it private. and it doesn't have to live here.

lib/packaging/artifactory.rb
shrug: At least I think the properties should mirror the contents of <sha>.yaml, it should be easy enough to parse that out and set the values? I wouldn't block this pr though, we can add that later

tasks/ship.rake
shrug: At some point, in the relatively near future, I think we should refactor our builds to be very explicit about the artifacts that they output, that the rest of our automation(including this, the signing workflow, etc) can consume. Until then I think this conforms with the rest of the codebase.

lib/packaging/artifactory.rb
melissa: You can absolutely not pass in platform_tag, but it'll assume that you're shipping to the 'generic' repo in that case. It will still allow you access to all the utilities, but assume it's shipping a file that isn't platform specific

underscorgan:
@melissa there are a bunch of unit test failures

lib/packaging/artifactory.rb
```diff
@@ -7,146 +7,126 @@ module Pkg
...
+        platform: platform,
+        platform_version: version,
+        architecture: architecture,
+        codename: codename,
+        package_format: package_format,
+        repo_name: repo_name,
+        repo_subdirectories: repo_subdirectories,
+        alternate_subdirectories: alternate_subdirectories,
+        full_artifactory_path: full_artifactory_path
+      }
```
mwaggett: I take it you expect / are okay with some of these values being nil?

lib/packaging/artifactory.rb
```diff
@@ -180,27 +162,68 @@ def check_authorization
...
-          'deb.component' => @repo_subdirectories,
-          'deb.architecture' => @architecture,
-        }
+    #
+    #   Currently we are including everything that would be included in the yaml
+    #   file that is generated at package build time.
+    def deploy_properties(platform_tag)
+      data = platform_specific_data(platform_tag)
+
+      #properties_hash = Pkg::Config.config_to_hash
```
mwaggett: remove comment

lib/packaging/artifactory.rb
```diff
@@ -180,27 +162,68 @@ def check_authorization
...
+      properties_hash = {}
+      if data[:package_format] == 'deb'
+        properties_hash.merge({
+          'deb.distribution' => data[:codename],
+          'deb.component' => data[:repo_subdirectories],
+          'deb.architecture' => data[:architecture],
+        })
       end
-      {}
+      puts properties_hash
```
mwaggett: probably want to take this out too

lib/packaging/artifactory.rb
```diff
@@ -180,27 +162,68 @@ def check_authorization
...
+    #
+    #   Currently we are including everything that would be included in the yaml
+    #   file that is generated at package build time.
+    def deploy_properties(platform_tag)
+      data = platform_specific_data(platform_tag)
+
+      #properties_hash = Pkg::Config.config_to_hash
+      properties_hash = {}
+      if data[:package_format] == 'deb'
+        properties_hash.merge({
```
mwaggett: Is there a reason you're merging with an empty hash instead of doing what you were doing before?

lib/packaging/artifactory.rb
```diff
@@ -180,27 +162,68 @@ def check_authorization
...
+        File.join(data[:alternate_subdirectories], File.basename(package)),
+        deploy_properties(platform_tag)
+      )
+    rescue
+      raise "Attempt to upload '#{package}' to #{File.join(@artifactory_uri, data[:full_artifactory_path])} failed"
+    end
+
+    # @param directory [String] optional, The directory where the yaml file will
+    #   be downloaded
+    def retrieve_yaml_data_file(directory = nil)
```
mwaggett: yardoc for return

lib/packaging/artifactory.rb
```diff
@@ -180,27 +162,68 @@ def check_authorization
...
+      directory ||= Dir.mktmpdir
+      retrieve_package(DEFAULT_REPO_TYPE, "#{@project_version}.yaml", directory)
+      File.join(directory, "#{@project_version}.yaml")
+    end
+
+    # @return [String] The name of the package for the given project,
+    #   project_version, and platform_tag
+    def package_name(platform_data, platform_tag)
```
mwaggett: yardoc for params

PR #728, opened 2017-10-05
(maint) Do not report unversioned msis

melissa:
Prior to this commit, we had been reporting unversioned msis in the
platform_data hash. This was fine for testing, as the unversioned msi is
simply a link to the versioned one. But this is bad when we go to ship
the packages. If we are fetching packages with FOSS_ONLY=true, it will
query the yaml data looking for foss only platform targets and fetch the
corresponding artifact. This is bad, since it means that for windows, we
were fetching the unversioned artifact for the final ship. This
changes ensures we don't set the unversioned msi when we are initially
writing out the platform_data hash.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -84,6 +84,13 @@ module Pkg
             platform, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
             arch = 'ppc' if platform == 'aix'
             package_format = Pkg::Platforms.get_attribute(tag, :package_format)
+
+            # Skip this if it's an unversioned MSI. We create these to help
+            # beaker install the msi without having to know any version
+            # information, but we should report the versioned artifact in
+            # platform_data
+            next if platform == 'windows' && artifact == "#{self.project}-#{arch}.#{package_format}"
+
             case package_format
             when 'deb'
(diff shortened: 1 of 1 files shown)


lib/packaging/config.rb
```diff
@@ -84,6 +84,13 @@ def platform_data
             platform, _, arch = Pkg::Util::Platform.parse_platform_tag(tag)
             arch = 'ppc' if platform == 'aix'
             package_format = Pkg::Platforms.get_attribute(tag, :package_format)
+
+            # Skip this if it's an unversioned MSI. We create these to help
+            # beaker install the msi without having to know any version
+            # information, but we should report the versioned artifact in
+            # platform_data
+            next if platform == 'windows' && artifact == "#{self.project}-#{arch}.#{package_format}"
```
mwaggett: if this is the approach we want to take, I feel like it would be better to be more general about this.
What if before we even get the tag, we do:
`next if !artifact.include? self.ref`

melissa: `self.ref` doesn't always correspond to the version that is included in the package name

mwaggett: oh yeah, you're right. grr.

mwaggett: we should also skip the .wixpdb packages

melissa: that would be something we can do for all packages, where we check what package_format is set to, and take that. That should be a ticketed bug fix, since it applied to more than just windows

PR #729, opened 2017-10-09
(maint) Remove EOL platforms

mwaggett:


melissa:
Have all of these platforms been fully removed from all the other build/test pipelines? If not, then this will cause pipelines failures.

mwaggett:
@melissa They weren't built for agent 5.3.2, so I assume so?

@McdonaldSeanp @ScottGarman Can you confirm?

shrug:
ping @geoffnichols

McdonaldSeanp:
@shrug @melissa @mwaggett These OSes are no longer built and/or tested anywhere.

These defs can be removed.

PR #733, opened 2017-10-18
Ticket/1.0.x/re 9510 fix platform tag assumptions

melissa:


lib/packaging/paths.rb
```diff
@@ -9,36 +9,48 @@ module Pkg::Paths
...
 
-    Pkg::Platforms.arches_for_platform_version(platform, version).find { |a| path.include?(a) } || Pkg::Platforms.arches_for_platform_version(platform, version)[0]
+    # First check if it's a source package
+    source_formats = Pkg::Platforms.get_attribute_for_platform_version(platform, version, :source_package_formats)
+    if source_formats.find { |fmt| path =~ /#{fmt}$/ }
+      return Pkg::Platforms.get_attribute_for_platform_version(platform, version, :source_architecture)
+    end
+    arches.find { |a| path.include?(a) } || arches[0]
+  rescue
+    arches.find { |a| path.include?(a) } || arches[0]
```
melissa: oh, I should probably change this. The idea is that we want to return this both if there's a failure, and if we don't return the source architecture. I didn't put a lot of thought into the best way to logic that.

lib/packaging/platforms.rb
```diff
@@ -5,71 +5,313 @@
...
+    # This is probably a bad assumption, but I'm assuming if we find a codename,
+    # that's more reliable as it's less likely to give us a false match
+    if codename
+      platform, version = codename_to_platform_version(codename)
+    end
+
+    # There's a possibility that the platform name has a dash in it, in which
+    # case, our assumption that it's an element of the above array is false,
+    # since it would be a combination of elements in that array
+    platform ||= supported_platforms.find { |p| platform_tag =~ /#{p}-/ }
```
melissa: I can add this to the comment, but I don't want to default to this since we have platforms like `el`, that are super easy to get false matches for. But platforms like `cisco-wrlinux` aren't going to play well with matching single elements in the array

lib/packaging/platforms.rb
```diff
@@ -5,71 +5,313 @@
...
+
+    version ||= (platform_elements & versions_for_platform(platform)).first
+
+
+    # For platform names with a dash in them, because everything is special
+    supported_arches = get_attribute_for_platform_version(platform, version, :architectures)
+    architecture = (platform_elements & supported_arches).first
```
melissa: Actually, this is a bad assumption. This should fail if we're passing in an architecture that's not supported. I need to update this.

lib/packaging/paths.rb
melissa: nevermind, Morgan and I couldn't get this to play nice with `ensure`, so it's staying as is

PR #735, opened 2017-10-25
(maint) Only require artifactory gem if we are using artifactory

melissa:
Previously, the fact that we had been requiring the artifactory gem at
the top of this file meant that calling any rake tasks would fail if the
artifactory gem was not installed. This commit moves the requires inside
the init for artifactory, so that we only require that library when we
are actually using it.


melissa:
I hate rspec.

melissa:
So, it looks like this is only failing sometimes, because we run rspec in a random order. This is a flaw in how I wrote my rspec.

PR #736, opened 2017-10-25
(maint) Make needed variables accessible in tasks via `let`

underscorgan:


spec/lib/packaging/artifactory_spec.rb
```diff
@@ -9,31 +9,33 @@
...
+      },
+      'solaris-10-sparc' => {
         :artifact => "./solaris/10/PC1/puppet-agent-5.3.1.34.gf65f9ef-1.sparc.pkg.gz",
         :repo_config => '',
-    },
+      },
+    }
   }
 
   platform_tags = {
```
melissa: can you also update `platform_tags`?

melissa: or does that not work with rspec?

mwaggett: +1

underscorgan: that doesn't work with rspec, we're only accessing platform_tags outside of tests

melissa:
I'm merging this, as it's a trivial change.

PR #737, opened 2017-10-30
(RE-8843) Don't fail if ext/project_data.yaml doesn't exist

underscorgan:
This commit adjusts the loading of default configs to move on if
project_data.yaml isn't found. Load will only fail if both project_data.yaml
and build_defaults.yaml cannot be found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ rescue LoadError
 end
 ```
 
-Also in ext should be two files, build_defaults.yaml and project_data.yaml.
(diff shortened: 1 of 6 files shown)


underscorgan:
Ran into this for something using packaging#master and it was super hard to track down. This change should be safe to be in master.

underscorgan:
Since this is just a cherry-pick of what's in the 1.0.x branch I'd rather leave it as it is.

PR #739, opened 2017-11-02
(maint) If server name includes '@' don't make user assumptions

underscorgan:


sarameisburger:
We all know what happens when you assume...

PR #741, opened 2017-11-08
(maint) Remove project config from synced path

demophoon:
This commit removes the project name from the final synced path as it
when the S3 sync happens we get an extra directory too deep when syncing
files.

### DIFF
--- a/tasks/nightly_repos.rake
+++ b/tasks/nightly_repos.rake
@@ -323,7 +323,7 @@ namespace :pl do
       version = ENV['VERSION'] || Pkg::Util::Version.get_dot_version
 
       tempdir = Pkg::Util::File.mktemp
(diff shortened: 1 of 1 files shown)


mwaggett:
Can you show me what was happening before and what is happening now, once this goes in? I'm a little confused by your commit message.

demophoon:
@mwaggett So before when we did the sync_to_s3 step the file that was synced was `pdk/LATEST` when instead it should have just been `LATEST`. This caused s3 to create a new directory called `pdk` with the `LATEST` file inside of it instead of just puttin the `LATEST` file at the root which is where the latest pdk cgi-bin script will look for it.

mwaggett:
@demophoon ahh. I just want to make sure that it's clear which project this LATEST file belongs to.

demophoon:
Yeah, it will always be uploaded to a specific bucket, as specified by the environment variable `TARGET_BUCKET`.

PR #742, opened 2017-11-08
Mergeup master -> 1.0.x

underscorgan:


README.md
```diff
@@ -11,6 +11,15 @@ flux, and it should not be considered to have a formal API. However, every
 effort is being made to ensure existing tasks/behavior are not broken as we
 continue to iterate and improve upon it.
 
+## Getting Started
```
e-gris: I think this is a fantastic change. It's the best change. Believe me. :)

lib/packaging/osx.rb
```diff
@@ -3,8 +3,13 @@ class << self
     def sign(target_dir = 'pkg')
       use_identity = "-i #{Pkg::Config.osx_signing_ssh_key}" unless Pkg::Config.osx_signing_ssh_key.nil?
 
-      ssh_host_string = "#{use_identity} #{ENV['USER']}@#{Pkg::Config.osx_signing_server}"
-      rsync_host_string = "-e 'ssh #{use_identity}' #{ENV['USER']}@#{Pkg::Config.osx_signing_server}"
+      if Pkg::Config.osx_signing_server =~ /@/
```
e-gris: <insert my normal complaint about "else" statements>

underscorgan: fair, but should probably be fixed in master then merged up since this is just a mergeup

e-gris: Yup, just wanted to comment, not request any actual action.

PR #743, opened 2017-11-08
(maint) Only retrieve <ref>.yaml when using FOSS_ONLY

mwaggett:
This commit refactors the retrieve task to only look for the <ref>.yaml file when using FOSS_ONLY to ship. Previously, this caused problems during internal ships because we were looking for <ref>.yaml in directories other than the 'artifacts' directory.


mwaggett:
@e-gris I know, I was just thinking that. I can work on that, but I am also eager to get a fix in so that puppet-agent passes, because that's blocking a lot of things.

e-gris:
Understood.

PR #744, opened 2017-11-09
(maint) Mistakes and cleanup

mwaggett:
This commit moves the `package_url` variable to higher up in the fetch process it is needed by both foss_only and regular retrieval. I've also begun some cleanup of this file.

### DIFF
--- a/tasks/retrieve.rake
+++ b/tasks/retrieve.rake
@@ -17,6 +17,7 @@ namespace :pl do
       local_target = args.local_target || "pkg"
       Pkg::Util::RakeUtils.invoke_task("pl:fetch")
       mkdir_p local_target
+      package_url = "http://#{Pkg::Config.builds_server}/#{Pkg::Config.project}/#{Pkg::Config.ref}"
       if wget = Pkg::Util::Tool.find_tool("wget")
         if Pkg::Config.foss_only && !Pkg::Config.foss_platforms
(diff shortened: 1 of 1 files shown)


tasks/retrieve.rake
```diff
@@ -44,25 +44,17 @@ namespace :pl do
             sh "#{wget} --quiet --recursive --no-parent --no-host-directories --level=0 --cut-dirs 3 --directory-prefix=#{local_target} --reject 'index*' #{package_url}/#{remote_target}/#{paths[:artifact]}" if Pkg::Config.foss_platforms.include?(platform)
           end
         else
-          # For the next person who needs to look these flags up:
```
shrug: Dropping the comments?

mwaggett: it honestly was making things harder for me to read. I think including the full flags will help make things less mysterious.

underscorgan: bah I figured those were going away since we don't have short flags anymore, but these are still super helpful as someone who doesn't `curl` all the time. @mwaggett can you put these back please (in another PR?)

mwaggett: okay

shrug: If anything we could update them to refer to the full flags, but it's not alwas clear why we're using a certain flag so I kinda like them

mwaggett: I'm working on a refactor of the retrieval anyway, so I'll put that back somewhere useful.

PR #745, opened 2017-11-09
(maint) Using && and || can cause issues

underscorgan:
Care needs to be taken when combining && and || in a long command to
avoid running into operator precedence issues.

### DIFF
--- a/lib/packaging/deb/repo.rb
+++ b/lib/packaging/deb/repo.rb
@@ -83,7 +83,7 @@ module Pkg::Deb::Repo
     end
 
     def repo_creation_command(repo_directory, artifact_paths)
-      cmd = "[ -d #{repo_directory} ] || exit 1 && "
+      cmd = "[ -d #{repo_directory} ] || exit 1 ; "
       cmd << "pushd #{repo_directory} > /dev/null && "
       cmd << 'echo "Checking for running repo creation. Will wait if detected." && '
       cmd << 'while [ -f .lock ] ; do sleep 1 ; echo -n "." ; done && '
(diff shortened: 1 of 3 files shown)


lib/packaging/deb/repo.rb
```diff
@@ -83,7 +83,7 @@ def retrieve_repo_configs(target = "repo_configs")
     end
 
     def repo_creation_command(repo_directory, artifact_paths)
-      cmd = "[ -d #{repo_directory} ] || exit 1 && "
+      cmd = "[ -d #{repo_directory} ] || exit 1 ; "
```
sarameisburger: What was the purpose of having the && in the first place?

underscorgan: Looking back in the original commit all the commands used to end in `;` and were switched to `&&` so a command fails in the chain we don't end up running a command in an unplanned location or end up in a weird state if we failed a prerequisite command or something.

underscorgan: (https://github.com/puppetlabs/packaging/pull/693/commits/b17bc443a5c48454a693da9ddbd2988466405792 for reference)

sarameisburger: Ah, interesting. Would using 'and' be useful at all? Stack Overflow tells me it has lower precedence

underscorgan: I don't think bash has `and` (these are going to be executed in bash via an ssh command)

sarameisburger: Ohh good to know. Thanks for all the info!

PR #748, opened 2017-11-21
(maint) More compatibility work for nightlies/internal rpm shipping

underscorgan:


underscorgan:
merging because @shrug told me to

PR #751, opened 2017-12-04
(RE-9522) Add s3 repo sync task for nightlies

mwaggett:
This commit adds a rake task for syncing nightlies to s3, the same way we sync apt, yum, and downloads.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -210,6 +210,17 @@ namespace :pl do
       end
     end
 
+    desc "Sync nightlies.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3"
+    task :deploy_nightlies_to_s3 => 'pl:fetch' do
+      puts "Really run S3 sync to sync nightlies.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          command = 'sudo /usr/local/bin/s3_repo_sync.sh nightlies.puppetlabs.com'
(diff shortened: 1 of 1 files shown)


tasks/ship.rake
```diff
@@ -210,6 +210,17 @@ namespace :pl do
       end
     end
 
+    desc "Sync nightlies.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3"
```
shrug: should we use puppet.com instead of puppetlabs.com ?

mwaggett: We could. I was just being consistent with what we have for apt, yum, and downloads. They go to the same place.

shrug: Yep, I'm mostly wondering if we should make a concerted effort to update them all, or update as we touch them

mwaggett: I like doing them all at once, personally. because consistency++

tasks/ship.rake
```diff
@@ -210,6 +210,17 @@ namespace :pl do
...
+    task :deploy_nightlies_to_s3 => 'pl:fetch' do
+      puts "Really run S3 sync to sync nightlies.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3? [y,n]"
+      if Pkg::Util.ask_yes_or_no
+        Pkg::Util::Execution.retry_on_fail(:times => 3) do
+          command = 'sudo /usr/local/bin/s3_repo_sync.sh nightlies.puppetlabs.com'
```
stahnma: Curious if we document where this comes from and how it came to be anywhere....

mwaggett: @stahnma https://confluence.puppetlabs.com/display/SRE/Apt+and+Yum+Download+Service+Docs

PR #755, opened 2017-12-06
(RE-9526) Add s3 repo sync to nightlies job

mwaggett:


tasks/jenkins.rake
```diff
@@ -265,6 +265,12 @@ namespace :pl do
       end
     end
 
+    task :ship_nightlies => "pl:fetch" do
+      Rake::Task['pl:jenkins:uber_ship_lite'].invoke
+      Rake::Task['remote:deploy_nightlies_to_s3'].invoke
+    end
+    end
```
e-gris: This looks like either an indentation error or an extraneous ```end```

tasks/ship.rake
```diff
@@ -212,12 +212,10 @@ namespace :pl do
...
     task :deploy_nightlies_to_s3 => 'pl:fetch' do
-      puts "Really run S3 sync to sync nightlies.puppetlabs.com from #{Pkg::Config.staging_server} to AWS S3? [y,n]"
-      if Pkg::Util.ask_yes_or_no
-        Pkg::Util::Execution.retry_on_fail(:times => 3) do
-          command = 'sudo /usr/local/bin/s3_repo_sync.sh nightlies.puppetlabs.com'
-          Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.staging_server, command)
-        end
+      puts "Deploying nightly builds from #{Pkg::Config.staging_server} to AWS S3..."
+      Pkg::Util::Execution.retry_on_fail(:times => 3) do
+        command = 'sudo /usr/local/bin/s3_repo_sync.sh nightlies.puppetlabs.com'
```
e-gris: (acknowledging that this isn't the point of this PR but just feel compelled to ask these questions)

Is there any easy way to make this work without using sudo? Running this as root doesn't seem right to me but there may be bits that I do not understand.

tasks/jenkins.rake
mwaggett: yep, typo, removed it. 😊

tasks/ship.rake
mwaggett: Right now, these are the permissions for that script:
`-rwx------ 1 root staff 1873 Dec  4 09:51 s3_repo_sync.sh`
I'm not entirely sure why that decision was made - I could see it being changed to be owned by the release group and executable by group.

e-gris: An investigation into it I think is worthwhile, but not in the context of this PR.

underscorgan: There was definitely a reason this needed to run as root (at least with the current implementation) but I'm going to need to dig back through memories of setting this up to remember _why_

PR #760, opened 2017-12-18
(maint) Add rake tasks for automated shipping

mwaggett:


tasks/jenkins.rake
```diff
@@ -267,11 +267,17 @@ namespace :pl do
...
     task :ship_nightlies => "pl:fetch" do
       Rake::Task['pl:jenkins:uber_ship_lite'].invoke
-      Rake::Task['pl:remote:update_apt_repo'].invoke
-      Rake::Task['pl:remote:update_yum_repo'].invoke
+      Rake::Task['pl:remote:update_foss_repos'].invoke
       Rake::Task['pl:remote:deploy_nightlies_to_s3'].invoke
     end
 
+    task :ship_final => "pl:fetch" do
+      Rake::Task['pl:jenkins:uber_ship_lite'].invoke
```
e-gris: I understand that this is historic but 'uber_ship_lite' has no discernible meaning. Not something to fix in this PR but wanted to note that a more descriptive name would be nice to have. Maybe creating a ticket that deprecates it for something more understandable?

mwaggett: yeah, that makes sense. I think when we ultimately break packaging down into chunks for signing, shipping, etc. we'll be able to choose better names.

PR #762, opened 2018-01-03
(RE-9886) Update windows keys in platform_data to include `r2`

mwaggett:
This commit adds `r2` to windows versions in platform_data so that beaker can find artifacts based on the same platform tag used in puppet-agent configs.

Related to https://github.com/puppetlabs/beaker-hostgenerator/pull/103

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -113,6 +113,9 @@ module Pkg
             # Remove the f-prefix from the fedora platform tag keys so that
             # beaker can rely on consistent keys once we rip out the f for good
             tag = tag.sub(/fedora-f/, 'fedora-')
+            # Add `r2` to windows versions so that beaker can find artifacts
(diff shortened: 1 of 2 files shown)


mwaggett:
cc @MikaelSmith @joshcooper @melissa

MikaelSmith:
How do you verify that this will work?

MikaelSmith:
Also, looking at what's done for Fedora, it seems like a change might still be needed for https://github.com/puppetlabs/packaging/blob/master/lib/packaging/platforms.rb#L66.

MikaelSmith:
This is more complicated than anticipated. We could change packaging to use 2012r2, which might be a less painful change than trying to change everything to a nonsense version for Windows. I've started an e-mail discussion to consider solutions.

mwaggett:
closing this in favor of https://github.com/puppetlabs/ci-job-configs/pull/3883

PR #763, opened 2018-01-04
(maint) Substitute 'power' for 'ppc' in AIX repo_config

mwaggett:
This commit replaces 'power' with 'ppc' when listing AIX repo_config paths in platform_data, since 'ppc' is the arch used in package names.

PR #764, opened 2018-01-04
(maint) Substitute 'power' for 'ppc' in AIX repo_config

mwaggett:
This commit replaces 'power' with 'ppc' when listing AIX repo_config paths in platform_data, since 'ppc' is the arch used in package names.


mwaggett:
This is the master counterpart to https://github.com/puppetlabs/packaging/pull/763

PR #765, opened 2018-01-05
(RE-9889) Nightly repo ship only generates repo configs, ships no artifacts

mwaggett:
This commit adds an optional parameter `repo_target` to Pkg::Paths.repo_path. Previously, the value of Pkg::Paths.repo_name was used instead. This allows the nightly ship task to find builds using the (final) repo name, rather than the nonfinal repo name, when shipping to ravi. Ideally, this should be reverted once the old nightlies process is obsoleted.

### DIFF
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -176,7 +176,7 @@ module Pkg::Paths
     end
   end
 
-  def repo_path(platform_tag, legacy = false)
+  def repo_path(platform_tag, legacy: false)
     repo_target = repo_name
     repo_target = Pkg::Config.repo_name if legacy
     platform, version, arch = Pkg::Platforms.parse_platform_tag(platform_tag)
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -176,21 +176,27 @@ module Pkg::Paths
     end
   end
 
-  def repo_path(platform_tag)
+  def repo_path(platform_tag, legacy = false)
+    repo_target = repo_name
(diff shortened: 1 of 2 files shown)


mwaggett:
Shipped successfully with this! http://nightlies.puppetlabs.com/puppetserver/5.2.0.master.SNAPSHOT.2018.01.05T1534/repos/

tasks/nightly_repos.rake
```diff
@@ -196,7 +196,7 @@ namespace :pl do
 
       if args.foss_only && Pkg::Config.foss_platforms && !Pkg::Config.foss_platforms.empty?
         Pkg::Config.foss_platforms.each do |platform|
-          include_paths << Pkg::Paths.repo_path(platform)
+          include_paths << Pkg::Paths.repo_path(platform, true)
```
e-gris: _Maybe_, just for the sake of clarity, make the second argument a named one? This helps call out the non-standard case.

```ruby
def repo_path(platform_tag, legacy: false)
  etc ...
end
```

```ruby
Pkg::paths.repo_path(platform, :legacy => true)
```

lib/packaging/paths.rb
```diff
@@ -176,21 +176,27 @@ def artifacts_path(platform_tag, path_prefix = 'artifacts')
     end
   end
 
-  def repo_path(platform_tag)
+  def repo_path(platform_tag, legacy: false)
```
mwaggett: @e-gris updated

PR #766, opened 2018-01-08
(RE-9947) Migrate from stickler to artifactory

sarameisburger:


mwaggett:
This should go into 1.0.x as well.

PR #767, opened 2018-01-09
(maint) Remove named parameters

mwaggett:
This commit switches repo_path to use `options` as a parameter instead of a named parameter, since Ruby < 2.0 does not support named parameters.

### DIFF
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -176,15 +176,15 @@ module Pkg::Paths
     end
   end
 
-  def repo_path(platform_tag, legacy: false)
+  def repo_path(platform_tag, options = { :legacy => false })
     repo_target = repo_name
-    repo_target = Pkg::Config.repo_name if legacy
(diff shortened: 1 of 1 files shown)


lib/packaging/paths.rb
```diff
@@ -176,15 +176,15 @@ def artifacts_path(platform_tag, path_prefix = 'artifacts')
     end
   end
 
-  def repo_path(platform_tag, legacy: false)
```
underscorgan: Do you need to make changes to how this is called also?

mwaggett: I don't think so, I tested in irb:
```
irb(main):001:0> def test(var, opts = {:thing => false})
irb(main):002:1> puts var if opts[:thing]
irb(main):003:1> end
=> :test
irb(main):004:0> test("hi")
=> nil
irb(main):005:0> test("hi", thing: true)
hi
=> nil
```

underscorgan: ok

PR #772, opened 2018-01-16
(RE-9889) `legacy` wasn't maintaining legacy behavior

underscorgan:
In an effort to maintain functionality of current nightlies pipelines, a
`legacy` option was added to Pkg::Paths.repo_path(). However, this did
not maintain behavior from packaging#master
Pkg::Util::Platform.repo_path(), which is causing existing nightly
pipelines to only ship repo configs, not actual repos/packages.

### DIFF
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -178,7 +178,12 @@ module Pkg::Paths
 
   def repo_path(platform_tag, options = { :legacy => false })
     repo_target = repo_name
-    repo_target = Pkg::Config.repo_name if options[:legacy]
+    # in legacy packaging methods, there was no consistent way to determine the
+    # repo name. There were separate variables for apt_repo_name and
+    # yum_repo_name. At times, either or both of these were unset, and they had
+    # different defaults. So, for legacy automation we need to just use a splat
+    # and globbing to find our packages.
+    repo_target = '**' if options[:legacy]
     platform, version, arch = Pkg::Platforms.parse_platform_tag(platform_tag)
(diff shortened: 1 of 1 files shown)


lib/packaging/paths.rb
```diff
@@ -178,7 +178,7 @@ def artifacts_path(platform_tag, path_prefix = 'artifacts')
 
   def repo_path(platform_tag, options = { :legacy => false })
     repo_target = repo_name
-    repo_target = Pkg::Config.repo_name if options[:legacy]
+    repo_target = '**' if options[:legacy]
```
shrug: https://media.giphy.com/media/Fwn1auHR2LUGI/giphy.gif

lib/packaging/paths.rb
```diff
@@ -178,7 +178,12 @@ def artifacts_path(platform_tag, path_prefix = 'artifacts')
 
   def repo_path(platform_tag, options = { :legacy => false })
     repo_target = repo_name
-    repo_target = Pkg::Config.repo_name if options[:legacy]
+    # in legacy packaging methods, there was no consistent way to determine the
+    # repo name. There were separate variables for apt_repo_name and
+    # yum_repo_name. At times, either or both of these were unset, and they had
+    # different defaults. So, for legacy automation we need to just use a splat
+    # and globbing to find our packages.
+    repo_target = '**' if options[:legacy]
```
shrug: hungry_hippos

e-gris: https://gph.is/1KP1eHi

sarameisburger: I really appreciate the descriptive comment!

PR #773, opened 2018-01-18
Fix typo - Rake::Take should be Rake::Task

shrug:


underscorgan:
invoked trivial merge

PR #777, opened 2018-01-25
(PA-1786) Add platform def for redhat-fips

underscorgan:


mwaggett:
@underscorgan Is it work targeting at master and merging up or nah?

underscorgan:
@mwaggett nope, this won't be going into anything other than 5.x and beyond.

PR #783, opened 2018-02-02
(maint) packaging doesn't recognize source platform tags

underscorgan:
When compiling the list of valid architectures to validate a platform
tag, it doesn't include source architectures. This means that things
like `el-7-SRPMS` and `ubuntu-14.04-source` are not recognized as valid
platform tags.

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -291,7 +291,7 @@ module Pkg::Platforms # rubocop:disable Metrics/ModuleLength
 
 
     # For platform names with a dash in them, because everything is special
-    supported_arches = arches_for_platform_version(platform, version)
+    supported_arches = arches_for_platform_version(platform, version, true)
     architecture = platform_tag.sub(/^(#{platform}-#{version}|#{codename})-?/, '')
 
     fail unless supported_arches.include?(architecture) || architecture.empty?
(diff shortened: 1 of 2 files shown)


underscorgan:
Tested by updating puppet-enterprise-vanagon to use packaging 1.0.x and successfully running `bundle exec ship`.

lib/packaging/platforms.rb
```diff
@@ -407,7 +407,16 @@ def codename_to_tags(codename)
   # Given a platform and version, return the arches that we build for that
   # platform
   def arches_for_platform_version(platform, version)
-    get_attribute_for_platform_version(platform, version, :architectures)
+    platform_architectures = get_attribute_for_platform_version(platform, version, :architectures)
+    # get_attribute_for_platform_version will raise an exception if the attribute
```
sarameisburger: I <3 comments

lib/packaging/platforms.rb
```diff
@@ -407,7 +407,16 @@ def codename_to_tags(codename)
...
+    # isn't found. We don't want this to be a fatal error, we just want to append
+    # the source architecture if it's found
+    begin
+      source_architecture = Array(get_attribute_for_platform_version(platform, version, :source_architecture))
+    rescue
+      source_architecture = []
+    end
+    return (platform_architectures + source_architecture).flatten
```
mwaggett: Is there a time when we'd only want platform architectures? I'm worried that `arches_for_platform_version` is/will be used in a way that only expects platform architectures. 

Perhaps there could be a separate `source_arches_for_platform_version` function and then a third function for combining them if you want both? In which case, maybe this one should be called `platform_arches_for_platform_version`?

underscorgan: Looking through all the places this is used, I think the only time you don't want sources included in there is when we're copying noarch packages around, and we were talking yesterday about how that behavior is problematic anyways 🤷‍♀️

underscorgan: Though in the interest of not breaking that even further, I'll at least rework how that bit is happening.

underscorgan:
@mwaggett on second thought, I think this is a lot safer. Running a test to make sure this still fixes the issue I was seeing, but this look better?

underscorgan:
yup, test worked \o/

PR #787, opened 2018-02-07
(maint) Specify path for remote bundle installs

mwaggett:
This commit adds `.bundle/gems` as the path to install gems when executing remote rake tasks. Previously, `bundle install` could fail because the user did not have write permission to the default path.


mwaggett:
Just successfully shipped puppetserver tarballs manually with this change.

PR #788, opened 2018-02-08
(RE-10238) FOSS_ONLY doesn't retrieve tarballs

mwaggett:
This PR adds `'level' => 1` as an option to `foss_only_retrieve` to ensure that all top-level artifacts (including source tarballs) are retrieved before platform-specific artifacts.

### DIFF
--- a/lib/packaging/retrieve.rb
+++ b/lib/packaging/retrieve.rb
@@ -8,9 +8,37 @@ module Pkg::Retrieve
   # --directory-prefix = where to save to disk (defaults to ./)
   # --reject = Reject all hits that match the supplied regex
 
-  def default_wget(local_target, url)
+  def default_wget_command(local_target, url, additional_options = {})
+    default_options = {
+      'quiet' => true,
+      'recursive' => true,
+      'no-parent' => true,
+      'no-host-directories' => true,
+      'level' => 0,
+      'cut-dirs' => 3,
+      'directory-prefix' => local_target,
+      'reject' => "'index*'",
+    }
+    options = default_options.merge(additional_options)
     wget = Pkg::Util::Tool.check_tool('wget')
-    wget_command = "#{wget} --quiet --recursive --no-parent --no-host-directories --level=0 --cut-dirs 3 --directory-prefix=#{local_target} --reject 'index*' #{url}"
+    wget_command = wget
+    options.each do |option, value|
+      next unless value
+      if value.is_a?(TrueClass)
+        wget_command << " --#{option}"
+      else
+        wget_command << " --#{option}=#{value}"
+      end
+    end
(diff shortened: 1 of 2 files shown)


mwaggett:
```
molly:puppetdb molly.waggett$ rake pl:jenkins:retrieve FOSS_ONLY=true
[ . . . ]
Executing /usr/local/bin/wget --quiet --recursive --no-parent --no-host-directories --level=1 --cut-dirs=3 --directory-prefix=pkg --reject='index*' http://builds.delivery.puppetlabs.net/puppetdb/5.1.4/artifacts/ . . .
[ . . . ]
Packages staged in pkg
molly:puppetdb molly.waggett$ ls pkg/
5.1.4.build_metadata.json		puppetdb-5.1.4-packaging-bundle.tar.gz
5.1.4.ezbake.manifest			puppetdb-5.1.4-signing_bundle.tar.gz
5.1.4.yaml				puppetdb-5.1.4.tar.gz
deb					sles
el
```

lib/packaging/retrieve.rb
```diff
@@ -8,9 +8,37 @@ module Pkg::Retrieve
...
+    options.each do |option, value|
+      next unless value
+      if value.is_a?(TrueClass)
+        wget_command << " --#{option}"
+      else
+        wget_command << " --#{option}=#{value}"
+      end
+    end
+    wget_command << " #{url}"
+    return wget_command
```
e-gris: Ruby purists would omit the "return" as redundant. I'm indifferent.

jennaesq: That's good to know... thank you @e-gris

lib/packaging/retrieve.rb
```diff
@@ -8,9 +8,37 @@ module Pkg::Retrieve
...
+    default_options = {
+      'quiet' => true,
+      'recursive' => true,
+      'no-parent' => true,
+      'no-host-directories' => true,
+      'level' => 0,
+      'cut-dirs' => 3,
+      'directory-prefix' => local_target,
+      'reject' => "'index*'",
+    }
```
underscorgan: What's the reasoning behind storing these in a map and translating them into wget parameters rather than just storing them in an array in the format they'll get passed to wget?

lib/packaging/retrieve.rb
mwaggett: I like the explicitness. Probably because of you, @e-gris 😄

lib/packaging/retrieve.rb
mwaggett: I wanted there to be an easy way for people to override existing params without having to retype all the defaults they wanted to keep, so using keys seemed like a good idea.

PR #789, opened 2018-02-12
(PA-1775) Update PLATFORM_INFO to include Fedora 27 (x86_64)

ScottGarman:


mwaggett:
This will need to go into 1.0.x too.

ScottGarman:
@mwaggett should I re-target this to 1.0.x so it can be merged up easier? Also I just realized the ticket number referenced is a duplicate, so I'd like to fix that before this gets merged.

mwaggett:
kk @ScottGarman leaving this targeted at master is fine, but it would be great if you could open a corresponding 1.0.x PR.

ScottGarman:
Ok, this should be ready for merge once Travis passes. I've created a PR targeting 1.0.x here: https://github.com/puppetlabs/packaging/pull/790

PR #794, opened 2018-02-14
Initial support for PE release branches

shrug:
This adds a new config variable for PE release branches,
and updates the ship tasks to ship to the correct locations
depending on release branch (vs feature or mainline)

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -121,6 +121,7 @@ module Pkg::Params
                   :packaging_url,
                   :pbuild_conf,
                   :pe_feature_branch,
+                  :pe_release_branch,
                   :pe_name,
                   :pe_platforms,
                   :pe_version,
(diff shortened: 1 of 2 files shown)


shrug:
Tested package promotions and composes from the command line with this change, they seem to work as expected.

We still need to test pez against the release branches, but that will come in a later PR if updates are needed.

lib/packaging/config/params.rb
```diff
@@ -121,6 +121,7 @@ module Pkg::Params
                   :packaging_url,
                   :pbuild_conf,
                   :pe_feature_branch,
+                  :pe_release_branch,
```
mwaggett: Should we add this as an env var too? There's one for feature branches.

shrug: I don't think the one for feature branches is ever used anywhere, so no I don't think so.

I don't want to remove the one for feature branches in this pr though, trying to keep it additive.

mwaggett:
These could be simplified to something like
```
return <path> if feature_branch
return <other_path> if release_branch
return <other_other_path>
```

mwaggett:
or we could have something like
```
extra_dir = ''
extra_dir = '/release' if release_branch
extra_dir = '/feature' if feature_branch
return "#{repo_path}/#{pe_version}#{extra_dir}/repos
```

shrug:
Any of those would work - which do you find most readable?

sarameisburger:
I like the first one more

mwaggett:
I'm torn. I think the first one is more readable but the second one feels DRYer?
I guess readability is probably most important.

shrug:
Updated: this better? or rather clean it up more? *I* find this more readable than having the conditional on the right, but that's a pretty personal preference.

PR #795, opened 2018-02-15
(maint) Separate nightly repo updates

mwaggett:
This PR modifies how we update apt and yum repos to be intentional about whether we are updating final or nonfinal repos, instead of relying on version magic.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -12,20 +12,11 @@ namespace :pl do
       else
         path = Pkg::Config.nonfinal_yum_repo_path || Pkg::Config.yum_repo_path
       end
-      yum_whitelist = {
-        __REPO_NAME__: Pkg::Paths.repo_name,
-        __REPO_PATH__: path,
-        __REPO_HOST__: Pkg::Config.yum_staging_server,
-        __GPG_KEY__: Pkg::Util::Gpg.key
-      }
+      command = Pkg::Config.yum_repo_command || 'rake -f /opt/repository/Rakefile mk_repo'
 
       $stdout.puts "Really run remote repo update on '#{Pkg::Config.yum_staging_server}'? [y,n]"
       if Pkg::Util.ask_yes_or_no
-        if Pkg::Config.yum_repo_command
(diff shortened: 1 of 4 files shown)


tasks/ship.rake
```diff
@@ -7,61 +7,37 @@ namespace :pl do
...
+    task update_nightlies_yum_repo: 'pl:fetch' do
+      command = Pkg::Config.yum_repo_command || 'rake -f /opt/repository/Rakefile mk_repo'
       $stdout.puts "Really run remote repo update on '#{Pkg::Config.yum_staging_server}'? [y,n]"
       if Pkg::Util.ask_yes_or_no
-        if Pkg::Config.yum_repo_command
-          Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.yum_staging_server, Pkg::Util::Misc.search_and_replace(Pkg::Config.yum_repo_command, yum_whitelist))
-        else
-          Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.yum_staging_server, 'rake -f /opt/repository/Rakefile mk_repo')
-        end
+        Pkg::Repo.update_yum_repo(Pkg::Config.nonfinal_repo_name, Pkg::Config.nonfinal_yum_repo_path, Pkg::Config.yum_staging_server, command)
```
underscorgan: What happens if Pkg::Config.nonfinal_yum_repo_path is nil?

tasks/ship.rake
```diff
@@ -7,61 +7,37 @@ namespace :pl do
...
-            Pkg::Config.apt_signing_server,
-            Pkg::Util::Misc.search_and_replace(
-              cmd,
-              apt_whitelist
-            )
-          )
-        else
-          warn %(Pkg::Config#apt_repo_command returned something unexpected, so no attempt will be made to update remote repos)
-        end
+        Pkg::Repo.update_apt_repo(Pkg::Config.nonfinal_repo_name, Pkg::Config.nonfinal_apt_repo_path, Pkg::Config.apt_host, Pkg::Config.apt_repo_url, Pkg::Config.nonfinal_apt_repo_command)
```
underscorgan: What happens if Pkg::Config.nonfinal_apt_repo_path is nil?

mwaggett:
@demophoon I added tests (good call!) and @underscorgan I updated to explicitly fail if any params are nil.

PR #796, opened 2018-02-15
(maint) Only fail repo updates if *required* params are missing

mwaggett:
This PR updates the update_repo methods to not fail if nil arguments are not required by the repo command.

### DIFF
--- a/lib/packaging/repo.rb
+++ b/lib/packaging/repo.rb
@@ -56,32 +56,22 @@ module Pkg::Repo
       repo_command.include?("__#{argument_name.upcase}__")
     end
 
-    def update_yum_repo(repo_name, repo_path, repo_host, command)
-      method(__method__).parameters.each do |_, param|
-        fail "Missing required argument #{param}, update your build_defaults?" if argument_required?(param.to_s, command) && binding.local_variable_get(param).to_s.empty?
-      end
-      yum_whitelist = {
-        __REPO_NAME__: repo_name,
-        __REPO_PATH__: repo_path,
-        __REPO_HOST__: repo_host,
-        __GPG_KEY__: Pkg::Util::Gpg.key
-      }
-      Pkg::Util::Net.remote_ssh_cmd(repo_host, Pkg::Util::Misc.search_and_replace(command, yum_whitelist))
-    end
+    def update_repo(remote_host, command, options = {})
+      fail_message = "Missing required argument '%s', update your build_defaults?"
+      fail fail_message % 'repo_name' if argument_required?('repo_name', command) && !options[:repo_name]
+      fail fail_message % 'repo_path' if argument_required?('repo_path', command) && !options[:repo_path]
+      fail fail_message % 'repo_host' if argument_required?('repo_host', command) && !options[:repo_host]
+      fail fail_message % 'repo_url' if argument_required?('repo_url', command) && !options[:repo_url]
 
-    def update_apt_repo(repo_name, repo_path, repo_host, repo_url, command)
-      method(__method__).parameters.each do |_, param|
-        fail "Missing required argument #{param}, update your build_defaults?" if argument_required?(param.to_s, command) && binding.local_variable_get(param).to_s.empty?
-      end
-      apt_whitelist = {
-        __REPO_NAME__: repo_name,
-        __REPO_PATH__: repo_path,
-        __REPO_HOST__: repo_host,
-        __REPO_URL__: repo_url,
+      whitelist = {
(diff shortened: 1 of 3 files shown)


lib/packaging/repo.rb
```diff
@@ -52,28 +52,25 @@ def populate_repo_directory(artifact_parent_directory)
...
     end
 
-    def update_apt_repo(repo_name, repo_path, repo_host, repo_url, command)
-      fail "At least one of your arguments is nil, update your build_defaults?" unless repo_name && repo_path && repo_host && repo_url && command
-      apt_whitelist = {
+    def update_repo(remote_host, command, repo_name = nil, repo_path = nil, repo_host = nil, repo_url = nil) # rubocop:disable Style/ParameterLists
+      fail_message = "Missing required argument '%s', update your build_defaults?"
+      fail fail_message % 'repo_name' if argument_required?('repo_name', command) && !repo_name
+      fail fail_message % 'repo_path' if argument_required?('repo_path', command) && !repo_path
+      fail fail_message % 'repo_url' if argument_required?('repo_url', command) && !repo_url
```
underscorgan: are you intentionally not checking for repo_host?

lib/packaging/repo.rb
```diff
@@ -52,28 +52,25 @@ def populate_repo_directory(artifact_parent_directory)
...
-      }
-      Pkg::Util::Net.remote_ssh_cmd(repo_host, Pkg::Util::Misc.search_and_replace(command, yum_whitelist))
+    def argument_required?(argument_name, repo_command)
+      repo_command.include?("__#{argument_name.upcase}__")
     end
 
-    def update_apt_repo(repo_name, repo_path, repo_host, repo_url, command)
-      fail "At least one of your arguments is nil, update your build_defaults?" unless repo_name && repo_path && repo_host && repo_url && command
-      apt_whitelist = {
+    def update_repo(remote_host, command, repo_name = nil, repo_path = nil, repo_host = nil, repo_url = nil) # rubocop:disable Style/ParameterLists
```
underscorgan: Would this be better as `def update_repo(remote_host, command, options = {})` ?

underscorgan: I feel like that's a bit cleaner

mwaggett: yeah, I was going to do that and I forget why I changed my mind. I'll try again.

lib/packaging/repo.rb
mwaggett: oops, no, I added that param last minute, will fix.

mwaggett:
@underscorgan updated

lib/packaging/repo.rb
```diff
@@ -52,28 +52,26 @@ def populate_repo_directory(artifact_parent_directory)
...
-    def update_apt_repo(repo_name, repo_path, repo_host, repo_url, command)
-      fail "At least one of your arguments is nil, update your build_defaults?" unless repo_name && repo_path && repo_host && repo_url && command
-      apt_whitelist = {
-        __REPO_NAME__: repo_name,
-        __REPO_PATH__: repo_path,
-        __REPO_HOST__: repo_host,
-        __REPO_URL__: repo_url,
+    def update_repo(remote_host, command, options = {})
+      fail_message = "Missing required argument '%s', update your build_defaults?"
+      fail fail_message % 'repo_name' if argument_required?('repo_name', command) && !options[:repo_name]
```
e-gris: Suggest using a loop. I haven't tested it but should work something like this.,,

```ruby
[:repo_name, :repo_path, :repo_host, :repo_url].each do |repo|
  fail fail_message % repo.to_s if argument_required?(repo.to_s, command) && !options[repo]
end
```

mwaggett: ooh, I like that, will update.

lib/packaging/repo.rb
```diff
@@ -52,28 +52,25 @@ def populate_repo_directory(artifact_parent_directory)
...
-      apt_whitelist = {
-        __REPO_NAME__: repo_name,
-        __REPO_PATH__: repo_path,
-        __REPO_HOST__: repo_host,
-        __REPO_URL__: repo_url,
+    def update_repo(remote_host, command, options = {})
+      fail_message = "Missing required argument '%s', update your build_defaults?"
+      [:repo_name, :repo_path, :repo_host, :repo_url].each do |option|
+        fail fail_message % option.to_s if argument_required?(option.to_s, command) && !options[option]
+      end
```
mwaggett: @e-gris updated

PR #798, opened 2018-02-16
(maint) Update apt repos on signing server, not staging server

mwaggett:
This was another typo / copy-pasta mistake. I'm sorry. :(


underscorgan:
@mwaggett have you been able to test this?

mwaggett:
@underscorgan I've been testing with the nightlies job after merge. which I realize is not the best, but I would otherwise have to build a new puppet-agent pointed at my branch and it would've taken forever. In the future, I will do that for big refactors, but for these little typos, it seems like a waste of time. Sorry I didn't do it in the first place. :/

PR #800, opened 2018-02-26
(RE-9325) update_yum_repo updates ALL repos, not just ones we're shipping to

mwaggett:
This commit adds separate rake tasks for updating ALL (final or nonfinal) yum
repos. Previously, the existing update_yum_repo and its nightly counterpart
would update all repos, but changes to the yum_repo_command change this
behavior. Usually we will only want to update the repo we're shipping to, but
this maintains ths ability to update all repos, should we want to.


mwaggett:
@jennaesq repo updates are in progress, but they seem to be proceeding as they should. 😊

mwaggett:
aaand completed without error!

PR #801, opened 2018-02-27
(maint) Use ssh for cloning build-data repo

mwaggett:
This commit changes the default build_data_repo value to be the ssh version
rather than the https version, due to Github's recent https changes (see
https://github.com/blog/2498-weak-cryptographic-standards-removal-notice).

### DIFF
--- a/lib/packaging/config/params.rb
+++ b/lib/packaging/config/params.rb
@@ -304,7 +304,7 @@ module Pkg::Params
               { :var => :random_mockroot,         :val => true },
               { :var => :keychain_loaded,         :val => false },
               { :var => :foss_only,               :val => false },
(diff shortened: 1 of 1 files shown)


mwaggett:
without this change:
```
** Execute pl:fetch
touch /var/lib/jenkins/.packaging
fatal: unable to access 'https://github.com/puppetlabs/build-data.git/': Peer reports incompatible or unsupported protocol version.
rm -rf /tmp/pkg8meyjg
rake aborted!
There was an error attempting to fetch the build-data repo. Maybe you do not have the correct permissions?
```

Was able to ship successfully by overriding to this value with an environment variable.

mwaggett:
We updated manasa (the box where this original failure was encountered). Closing this since making updates on our machines seems like a better solution than worrying about ssh key pairs.

PR #802, opened 2018-03-01
(maint) Add and use separate functions for shipping different package types

mwaggett:
This PR adds wrapper functions that use `ship_pkgs` and various link creation functions to ship different types of packages (rpms, debs, msis, etc.) and uses those in the various rake tasks for shipping.

This will be useful for [RE-10284](https://tickets.puppetlabs.com/browse/RE-10284).

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -485,6 +485,14 @@ module Pkg::Platforms # rubocop:disable Metrics/ModuleLength
     platform_tags
   end
 
+  # Return a supported platform tag for the given platform, not caring about
+  # version or architecture
+  def generic_platform_tag(platform)
+    version = versions_for_platform(platform).first
+    arch = arches_for_platform_version(platform, version).first
+    return "#{platform}-#{version}-#{arch}"
+  end
+
   # @method by_deb
   # @return [Array] An Array of Strings, containing all platforms
   #   that use .deb packages
--- a/lib/packaging/util/ship.rb
+++ b/lib/packaging/util/ship.rb
@@ -105,15 +105,7 @@ module Pkg::Util::Ship
   def ship_rpms(local_staging_directory, remote_path, opts = {})
     ship_pkgs(["#{local_staging_directory}/**/*.rpm", "#{local_staging_directory}/**/*.srpm"], Pkg::Config.yum_staging_server, remote_path, opts)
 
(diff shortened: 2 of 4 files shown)


lib/packaging/util/ship.rb
```diff
@@ -102,6 +102,111 @@ def ship_pkgs(pkg_exts, staging_server, remote_path, opts = {})
...
   end
 
+  def ship_rpms(local_staging_directory, remote_path, opts = {})
+    ship_pkgs(["#{local_staging_directory}/**/*.rpm", "#{local_staging_directory}/**/*.srpm"], Pkg::Config.yum_staging_server, remote_path, opts)
+
+    # I really don't care which one we grab, it just has to be some supported
+    # version and architecture from the `el` hash. So here we're just grabbing
+    # the first one, parsing out some info, and breaking out of the loop. Not
+    # elegant, I know, but effective. [written by Melissa, copied by Molly]
+    Pkg::Platforms::PLATFORM_INFO['el'].each do |key, value|
```
shrug: Why do we loop over 'each' and then break instead of just getting the first?

mwaggett: I'm not sure, I just copied this from Melissa's work. I can look into making this nicer.

mwaggett: so yeah, we could do
```
version = Pkg::Platforms::PLATFORM_INFO['el'].keys.first
arch = Pkg::Platforms::PLATFORM_INFO['el'][version][:architectures[0]]
generic_platform_tag = "el-#{version}-#{arch}"
```
but I'm not sure if that's better?

e-gris: Much, much, better IMO. Breaking a loop upon first iteration isn't a loop. So, the code shouldn't be looping.

lib/packaging/platforms.rb
```diff
@@ -485,6 +485,14 @@ def platform_tags_for_package_format(format)
...
   end
 
+  # Return a supported platform tag for the given platform, not caring about
+  # version or architecture
+  def generic_platform_tag(platform)
+    version = versions_for_platform(platform).first
+    arch = arches_for_platform_version(platform, version).first
+    return "#{platform}-#{version}-#{arch}"
+  end
+
```
mwaggett: @shrug @e-gris Added this, for use below.

PR #803, opened 2018-03-01
(RE-10284) Separate final and nightly ship tasks

mwaggett:
This commit adds separate rake tasks for shipping nightly packages. Previously,
we had single rake tasks for each package type that would switch between
shipping to the final repos and the nightly repos, based on the version of the
packages. This allows us to be intentional about what we're shipping where.

Note: I didn't add nightly tasks for svr4 and p5p since they are PE-only platforms right now.

### DIFF
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -265,8 +265,24 @@ namespace :pl do
       end
     end
 
+    task :stage_nightlies => "pl:fetch" do
+      tasks = %w(
+        jenkins:retrieve
+        jenkins:sign_all
+        ship_nightly_rpms
+        ship_nightly_debs
+        ship_nightly_dmg
+        ship_nightly_swix
+        ship_nightly_msi
+      )
+      tasks.map { |t| "pl:#{t}" }.each do |t|
+        puts "Running #{t} . . ."
+        Rake::Task[t].invoke
+      end
+    end
+
     task :ship_nightlies => "pl:fetch" do
-      Rake::Task['pl:jenkins:uber_ship_lite'].invoke
+      Rake::Task['pl:jenkins:stage_nightlies'].invoke
       Rake::Task['pl:remote:update_nightly_repos'].invoke
       Rake::Task['pl:remote:deploy_nightlies_to_s3'].invoke
     end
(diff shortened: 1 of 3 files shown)


mwaggett:
https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_nightly-repo_packaging_ship/3293/

tasks/jenkins.rake
```diff
@@ -265,8 +265,24 @@ namespace :pl do
...
+      tasks = %w(
+        jenkins:retrieve
+        jenkins:sign_all
+        ship_nightly_rpms
+        ship_nightly_debs
+        ship_nightly_dmg
+        ship_nightly_swix
+        ship_nightly_msi
+      )
+      tasks.map { |t| "pl:#{t}" }.each do |t|
```
e-gris: This is the correct way to do this but it still makes me wince.

tasks/jenkins.rake
```diff
@@ -265,8 +265,24 @@ namespace :pl do
...
+        puts "Running #{t} . . ."
+        Rake::Task[t].invoke
+      end
+    end
+
     task :ship_nightlies => "pl:fetch" do
-      Rake::Task['pl:jenkins:uber_ship_lite'].invoke
+      Rake::Task['pl:jenkins:stage_nightlies'].invoke
```
e-gris: Win! "uber_ship" has no meaning to me. "stage_nightlies" does.

tasks/ship.rake
```diff
@@ -435,11 +431,14 @@ namespace :pl do
...
       path = Pkg::Config.msi_path
     end
-    path = Pkg::Config.nonfinal_msi_path if Pkg::Config.nonfinal_msi_path && !Pkg::Util::Version.final?
-
     Pkg::Util::Ship.ship_msi('pkg', path, excludes: ["#{Pkg::Config.project}-x(86|64).msi"])
   end
 
+  desc "Ship nightly MSI packages to #{Pkg::Config.msi_staging_server}"
+  task ship_nightly_msi: 'pl:fetch' do
+    Pkg::Util::Ship.ship_msi('pkg', Pkg::Config.nonfinal_msi_path, excludes: ["#{Pkg::Config.project}-x(86|64).msi"])
```
shrug: What's the exclude?

mwaggett: The unversioned msis we build. We didn't have a comment about it before, but I can add one if you think that would be helpful.

PR #808, opened 2018-03-06
(RE-9425) uber_ship automation does not work for untagged ezbake builds

mwaggett:
This PR updates our signing automation to not rely on package versions and to fail explicitly if signing goes wrong. See individual commit messages for details.


lib/packaging/sign.rb
```diff
@@ -0,0 +1,49 @@
+module Pkg::Sign
+  module_function
+
+  def sign_rpm(rpm, sign_flags = nil)
+
+    # To enable support for wrappers around rpm and thus support for gpg-agent
+    # rpm signing, we have to be able to tell the packaging repo what binary to
+    # use as the rpm signing tool.
+    #
+    rpm_cmd = ENV['RPM'] || Pkg::Util::Tool.find_tool('rpm')
```
e-gris: Can we use 'command' rather than 'cmd' throughout?

tasks/sign.rake
```diff
@@ -1,77 +1,33 @@
...
 
   desc "Sign the Arista EOS swix packages, defaults to PL key, pass GPG_KEY to override or edit build_defaults"
   task :sign_swix do
     packages = Dir["pkg/**/*.swix"]
-    unless packages.empty?
-      Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
-      packages.each do |swix_package|
-        Pkg::Util::Gpg.sign_file swix_package
-      end
+    fail "No swix packages found. Did you retrieve?" if packages.empty?
```
mwaggett: I want to keep this, but right now, in the `sign_all` task (below) we choose to attempt swix signing if we're operating on a vanagon project. As far as I know, puppet-agent is the only project that ships swix packages, so this may fail for other vanagon projects. 

I have some ideas, but I think they'll require some big refactors, so I'm not sure what the best path for the immediate time being is.

lib/packaging/sign.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    # If we're using the gpg agent for rpm signing, we don't want to specify the
+    # input for the passphrase, which is what '--passphrase-fd 3' does. However,
+    # if we're not using the gpg agent, this is required, and is part of the
+    # defaults on modern rpm. The fun part of gpg-agent signing of rpms is
+    # specifying that the gpg check command always return true
+    #
```
e-gris: Give gpg_check_command and input_flag explicit initializations (even if they are empty strings).
Let the conditional override them.

tasks/retrieve.rake
```diff
@@ -24,7 +24,11 @@ namespace :pl do
       else
         Pkg::Retrieve.retrieve_all(build_url, build_path, remote_target, local_target)
       end
-      puts "Packages staged in #{local_target}"
+      if Dir["#{local_target}/*"].empty?
+        fail "Uh oh, looks like we didn't find anything to retrieve!"
+      else
```
e-gris: 'else' is not needed:

```ruby
if Dir["#{local_target}/*"].empty?
  fail "Uh oh, looks like we didn't find anything to retrieve!"
end

puts "Packages staged in #{local_target}"

File: tasks/sign.rake [hunk 5]
```diff
@@ -1,77 +1,33 @@
-def sign_rpm(rpm, sign_flags = nil)
```
e-gris: Yay! Out of rake and into .rb!

tasks/sign.rake
```diff
@@ -1,77 +1,33 @@
...
-  sign_program = "-p'gpg --use-agent --no-tty'" if ENV['RPM_GPG_AGENT']
-  sh "debsign #{sign_program} --re-sign -k#{Pkg::Config.gpg_key} #{file}"
-end
-
 namespace :pl do
   desc "Sign the tarball, defaults to PL key, pass GPG_KEY to override or edit build_defaults"
   task :sign_tar do
     unless Pkg::Config.vanagon_project
-      File.exist?("pkg/#{Pkg::Config.project}-#{Pkg::Config.version}.tar.gz") or fail "No tarball exists. Try rake package:tar?"
+      tars_to_sign = Pkg::Util::Ship.collect_packages(['pkg/*.tar.gz'], excludes: ['signing_bundle', 'packaging-bundle'])
```
e-gris: Maybe "tarballs_to_sign"?

lib/packaging/sign.rb
mwaggett: This was a copy+paste, but I can change that.

tasks/retrieve.rake
mwaggett: oh yeah, I was gonna have them both be `puts` and then changed this and forgot to update. Will do.

tasks/sign.rake
mwaggett: sure

lib/packaging/sign.rb
mwaggett: again, more copy+paste, but I can refactor. 😊

tasks/sign.rake
```diff
@@ -216,13 +175,14 @@ namespace :pl do
       signing_bundle = ENV['SIGNING_BUNDLE']
       rpm_sign_task = Pkg::Config.build_pe ? "pe:sign_rpms" : "pl:sign_rpms"
       deb_sign_task = Pkg::Config.build_pe ? "pe:sign_deb_changes" : "pl:sign_deb_changes"
-      sign_tasks    = [rpm_sign_task, deb_sign_task]
+      sign_tasks    = [rpm_sign_task]
```
mwaggett: This causes issues when we try to ship things like puppet, which is just gems and tars.
Thinking about a good approach.

mwaggett:
Closing this in favor of smaller PRs.

PR #809, opened 2018-03-08
(maint) Fail retrieval if destination directory is empty

mwaggett:
This commit adds a failure to the retrieve task if the local target (usually 'pkg') is empty at the end of the retrieve. Previously, the retrieve would appear successful, but future tasks would have nothing to operate on.

### DIFF
--- a/tasks/retrieve.rake
+++ b/tasks/retrieve.rake
@@ -24,6 +24,7 @@ namespace :pl do
       else
         Pkg::Retrieve.retrieve_all(build_url, build_path, remote_target, local_target)
       end
+      fail "Uh oh, looks like we didn't find anything to retrieve from #{build_url}!" if Dir["#{local_target}/*"].empty?
       puts "Packages staged in #{local_target}"
     end
   end


tasks/retrieve.rake
```diff
@@ -24,6 +24,7 @@ namespace :pl do
       else
         Pkg::Retrieve.retrieve_all(build_url, build_path, remote_target, local_target)
       end
+      fail "Uh oh, looks like we didn't find anything to retrieve!" if Dir["#{local_target}/*"].empty?
```
underscorgan: Might be useful to include the build url, so it's at least clear where it's trying to retrieve from

mwaggett: I was going to do that, but the retrieve tasks above already record that info, so I felt like it was kind of redundant.
```
Executing /usr/local/bin/wget --quiet --recursive --no-parent --no-host-directories --level=0 --cut-dirs=3 --directory-prefix=pkg --reject='index*' http://builds.delivery.puppetlabs.net/puppetserver/83bae0c9f8b3e6e5de5b890f2a2af7db4724d7ba/artifacts/ . . .
rake aborted!
Uh oh, looks like we didn't find anything to retrieve!
```

underscorgan: I'm wondering if both those messages end up sent to stderr, if not the repetition could be useful so output filtering doesn't leave you with an error with not enough context to debug

mwaggett: ooh, not sure. I can add the url to the fail message.

tasks/retrieve.rake
```diff
@@ -24,6 +24,7 @@ namespace :pl do
       else
         Pkg::Retrieve.retrieve_all(build_url, build_path, remote_target, local_target)
       end
+      fail "Uh oh, looks like we didn't find anything to retrieve from #{build_url}!" if Dir["#{local_target}/*"].empty?
```
mwaggett: @underscorgan updated! (will squash once you 👍)

PR #810, opened 2018-03-08
(maint) Add Pkg::Sign module

mwaggett:
This PR moves a bunch of signing functionality into its own module (almost exclusively copy pasta).


lib/packaging/sign.rb
```diff
@@ -2,34 +2,31 @@ module Pkg::Sign
...
-      gpg_check_cmd = "--define '%__gpg_check_password_cmd /bin/true'"
+      gpg_check_command = "--define '%__gpg_check_password_command /bin/true'"
     else
       input_flag = "--passphrase-fd 3"
     end
 
     # Try this up to 5 times, to allow for incorrect passwords
     Pkg::Util::Execution.retry_on_fail(:times => 5) do
-      # This definition of %__gpg_sign_cmd is the default on modern rpm. We
+      # This definition of %__gpg_sign_command is the default on modern rpm. We
```
underscorgan: `%__gpg_sign_cmd` is a pre-defined rpm macro, I don't think changing it to `%__gpg_sign_command` is going to work the way you're expecting

lib/packaging/sign.rb
```diff
@@ -2,34 +2,31 @@ module Pkg::Sign
...
     # input for the passphrase, which is what '--passphrase-fd 3' does. However,
     # if we're not using the gpg agent, this is required, and is part of the
     # defaults on modern rpm. The fun part of gpg-agent signing of rpms is
     # specifying that the gpg check command always return true
-    #
+    gpg_check_command = ''
+    input_flag = ''
     if Pkg::Util.boolean_value(ENV['RPM_GPG_AGENT'])
-      gpg_check_cmd = "--define '%__gpg_check_password_cmd /bin/true'"
+      gpg_check_command = "--define '%__gpg_check_password_command /bin/true'"
```
underscorgan: I don't think you can change the name of `%__gpg_check_password_cmd` to `%__gpg_check_password_command`, `%__gpg_check_password_cmd` is an expected RPM macro http://rpm5.org/docs/api/config_macros.html

underscorgan:
also it looks like by commenting on individual commits my comments are showing under 'show outdated', will re-add them inline on the final PR...

lib/packaging/sign.rb
```diff
@@ -0,0 +1,219 @@
...
+
+    # If we're using the gpg agent for rpm signing, we don't want to specify the
+    # input for the passphrase, which is what '--passphrase-fd 3' does. However,
+    # if we're not using the gpg agent, this is required, and is part of the
+    # defaults on modern rpm. The fun part of gpg-agent signing of rpms is
+    # specifying that the gpg check command always return true
+    gpg_check_command = ''
+    input_flag = ''
+    if Pkg::Util.boolean_value(ENV['RPM_GPG_AGENT'])
+      gpg_check_command = "--define '%__gpg_check_password_command /bin/true'"
```
underscorgan: I don't think you can change the name of %__gpg_check_password_cmd to %__gpg_check_password_command, %__gpg_check_password_cmd is an expected RPM macro http://rpm5.org/docs/api/config_macros.html

lib/packaging/sign.rb
```diff
@@ -0,0 +1,219 @@
...
+    else
+      input_flag = "--passphrase-fd 3"
+    end
+
+    # Try this up to 5 times, to allow for incorrect passwords
+    Pkg::Util::Execution.retry_on_fail(:times => 5) do
+      # This definition of %__gpg_sign_command is the default on modern rpm. We
```
underscorgan: %__gpg_sign_cmd is a pre-defined rpm macro, I don't think changing it to %__gpg_sign_command is going to work the way you're expecting

mwaggett: ahh good to know, I'll change it back.

mwaggett:
@underscorgan re: different files for each packaging type - that's what we had originally. My thoughts for changing it was that it felt unintuitive to me that, for example, the osx signing method was in Pkg::OSX with nothing else.. I want to see Pkg::Sign and know I'll find all the signing methods there. 
If we had lots of osx-specific methods in there (and corresponding ones for other platforms), I would be more inclined to do it the other way, but right now, it seemed more useful to consolidate. 
I also was thinking about creating an overall signing method that switches on package format and having all the functions it would call in the same place felt nice.

All that being said, I'm definitely willing to be convinced that the original separation is a better idea.

underscorgan:
@mwaggett I wasn't thinking the initial separation so much as having something like Pkg::Sign::OSX and Pkg::Sign::RPM or something like that, it just feels like a natural separation to me, probably at least in part since I almost never need to see all the signing tasks, usually if I'm looking for something related to that I need the information for a specific platform, so splitting it up would make it easier to find the specific things I'm looking for. For future there could still be something generic in Pkg::Sign that switches and calls the right methods.

lib/packaging/sign.rb
```diff
@@ -0,0 +1,219 @@
...
+
+    # If we're using the gpg agent for rpm signing, we don't want to specify the
+    # input for the passphrase, which is what '--passphrase-fd 3' does. However,
+    # if we're not using the gpg agent, this is required, and is part of the
+    # defaults on modern rpm. The fun part of gpg-agent signing of rpms is
+    # specifying that the gpg check command always return true
+    gpg_check_command = ''
+    input_flag = ''
+    if Pkg::Util.boolean_value(ENV['RPM_GPG_AGENT'])
+      gpg_check_command = "--define '%__gpg_check_password_cmd /bin/true'"
```
mwaggett: @underscorgan changed these back

mwaggett:
ahh, so something like lib/packaging/sign/osx.rb, lib/packaging/sign/msi.rb, etc.? And then potentially a lib/packaging/sign.rb with the overall thing?

underscorgan:
@mwaggett yup!

PR #811, opened 2018-03-09
(CPR-521) Create HFS+ images

ccaviness:
When building on macOS 10.13, hdiutil creates APFS images by default.
This is unnecessary and can cause issues on older releases, so specify
HFS+ instead.

### DIFF
--- a/tasks/apple.rake
+++ b/tasks/apple.rake
@@ -86,6 +86,7 @@ end
 def build_dmg
   # Local Variables
   dmg_format_code   = 'UDZO'
+  dmg_filesystem    = 'HFS+'
   zlib_level        = '9'
   dmg_format_option = "-imagekey zlib-level=#{zlib_level}"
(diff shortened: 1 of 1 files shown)


tasks/apple.rake
```diff
@@ -86,6 +86,7 @@ end
 def build_dmg
   # Local Variables
   dmg_format_code   = 'UDZO'
+  dmg_fs            = 'HFS+'
```
mwaggett: Do you think you could change this to `dmg_filesystem` so it's clear for future spelunkers?

ccaviness: Done.

ccaviness:
Ok, but be warned there's a good chance I'll screw it up entirely. Gimme a minute.

ccaviness:
Ok, I think I did it. Also changed to reference correct ticket (CPR-521).

shrug:
Great, thanks for the contribution!

PR #819, opened 2018-03-13
 Revert "Revert "(maint) Remove `remote_target` param from `retrieve_all`""

mwaggett:
The issue was a missing trailing slash..

### DIFF
--- a/lib/packaging/retrieve.rb
+++ b/lib/packaging/retrieve.rb
@@ -48,24 +48,24 @@ module Pkg::Retrieve
     unless Pkg::Config.foss_platforms
       fail "FOSS_ONLY specified, but I don't know anything about FOSS_PLATFORMS. Retrieve cancelled."
     end
-    default_wget(local_target, "#{build_url}/artifacts/", { 'level' => 1 })
+    default_wget(local_target, "#{build_url}", { 'level' => 1 })
     yaml_path = File.join(local_target, "#{Pkg::Config.ref}.yaml")
     unless File.readable?(yaml_path)
       fail "Couldn't read #{Pkg::Config.ref}.yaml, which is necessary for FOSS_ONLY. Retrieve cancelled."
     end
     platform_data = Pkg::Util::Serialization.load_yaml(yaml_path)[:platform_data]
     platform_data.each do |platform, paths|
-      default_wget(local_target, "#{build_url}/artifacts/#{paths[:artifact]}") if Pkg::Config.foss_platforms.include?(platform)
+      default_wget(local_target, "#{build_url}/#{paths[:artifact]}") if Pkg::Config.foss_platforms.include?(platform)
     end
   end
 
-  def retrieve_all(build_url, rsync_path, remote_target, local_target)
+  def retrieve_all(build_url, rsync_path, local_target)
     if Pkg::Util::Tool.find_tool("wget")
-      default_wget(local_target, "#{build_url}/#{remote_target}/")
+      default_wget(local_target, build_url)
     else
       warn "Could not find `wget` tool. Falling back to rsyncing from #{Pkg::Config.distribution_server}."
       begin
-        Pkg::Util::Net.rsync_from("#{rsync_path}/#{remote_target}/", Pkg::Config.distribution_server, "#{local_target}/")
+        Pkg::Util::Net.rsync_from(rsync_path, Pkg::Config.distribution_server, "#{local_target}/")
       rescue => e
         fail "Couldn't rsync packages from distribution server.\n#{e}"
       end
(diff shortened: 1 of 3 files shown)


mwaggett:
Tested here: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_nightly-repo_packaging_ship/3376/

lib/packaging/retrieve.rb
```diff
@@ -48,24 +48,24 @@ def foss_only_retrieve(build_url, local_target)
...
-  def retrieve_all(build_url, rsync_path, remote_target, local_target)
+  def retrieve_all(build_url, rsync_path, local_target)
     if Pkg::Util::Tool.find_tool("wget")
-      default_wget(local_target, "#{build_url}/#{remote_target}/")
+      default_wget(local_target, build_url)
     else
       warn "Could not find `wget` tool. Falling back to rsyncing from #{Pkg::Config.distribution_server}."
       begin
-        Pkg::Util::Net.rsync_from("#{rsync_path}/#{remote_target}/", Pkg::Config.distribution_server, "#{local_target}/")
+        Pkg::Util::Net.rsync_from(rsync_path, Pkg::Config.distribution_server, "#{local_target}/")
```
underscorgan: Are you still missing a trailing slash after rsync_path?

mwaggett: ooh yeah, if we only want the contents of the rsync path (and not just whole directory itself), then we want the slash there too. I'll add that.

tasks/retrieve.rake
```diff
@@ -17,14 +17,14 @@ namespace :pl do
       remote_target = args.remote_target || "artifacts"
       local_target = args.local_target || "pkg"
       mkdir_p local_target
-      build_url = "http://#{Pkg::Config.builds_server}/#{Pkg::Config.project}/#{Pkg::Config.ref}"
-      build_path = "#{Pkg::Config.jenkins_repo_path}/#{Pkg::Config.project}/#{Pkg::Config.ref}"
+      build_url = "http://#{Pkg::Config.builds_server}/#{Pkg::Config.project}/#{Pkg::Config.ref}/#{remote_target}/"
+      build_path = "#{Pkg::Config.jenkins_repo_path}/#{Pkg::Config.project}/#{Pkg::Config.ref}/#{remote_target}/"
```
mwaggett: @underscorgan Added the trailing slash here.

lib/packaging/retrieve.rb
```diff
@@ -48,24 +48,24 @@ def foss_only_retrieve(build_url, local_target)
...
-      default_wget(local_target, "#{build_url}/artifacts/#{paths[:artifact]}") if Pkg::Config.foss_platforms.include?(platform)
+      default_wget(local_target, "#{build_url}/#{paths[:artifact]}") if Pkg::Config.foss_platforms.include?(platform)
     end
   end
 
-  def retrieve_all(build_url, rsync_path, remote_target, local_target)
+  def retrieve_all(build_url, rsync_path, local_target)
     if Pkg::Util::Tool.find_tool("wget")
-      default_wget(local_target, "#{build_url}/#{remote_target}/")
+      default_wget(local_target, build_url)
```
underscorgan: Moving the trailing slashes from being included here to needing to be passed in the parameters seems like a confusing change. While reviewing it was difficult to trace through and figure out if all the places these are getting called now have the needed trailing slash, or if that's missing still.

I think it would be easier to reason about if we went back to having the trailing slashes added in the retrieve_all and foss_only_retrieve methods.

mwaggett: I'm not sure I agree.
For rsync, I definitely don't agree - in some cases we may want the slash and in other cases we might not. I think whatever is calling that should make that decision.
For wget, I feel similarly, but since we do default to using `--no-parent`, I can see the argument for including it in the function. 

I do understand what you're saying about it being hard to review, but I'd rather do the "right" thing than do something for the sake of easy review.

underscorgan: I don't want it changing in the `default_wget` and `rsync_from` methods, I want it to change in how we call those methods from `retrieve_all` and `retrieve_foss_only`, since I would otherwise need to read through the methods to figure out how it's retrieving where or when I may need slashes. That doesn't feel like the 'right' thing.

mwaggett: okay, that is fair.

mwaggett:
@underscorgan updated

underscorgan:
want to squash?

mwaggett:
Squashed! (I wanted to leave the revert commit separate from the slashes one so it's clear what the history there was.)

underscorgan:
Yes, don't squash that. It's fine for now, but in the future if you could have each revert as a separate commit that makes it much easier when trying to figure out which reverts have been reverted, rather than having to read through full commit messages as they get squashed together (I'm thinking of say, working on changelogs and trying to figure out what's actually there....)

mwaggett:
That's fair. I grouped them together because that work should never have been split in the first place (my bad, typos and missing some function changes). For example, if we needed to revert again, I would want them all reverted together, but I understand how that could be confusing later on from a changelog perspective.

PR #820, opened 2018-03-13
(RE-10396) Fail explicitly when signing fails

mwaggett:
This commit changes the local signing functions to use `capture3` rather than `%x` so that failures are captured and reported.


mwaggett:
The other signing functions use `remote_ssh_cmd` which raises errors when things fail.

mwaggett:
Tested signing locally and everything went smoothly. Not sure the best way to induce failure..

PR #825, opened 2018-03-15
(RE-10413) Updates for legacy compatibility

underscorgan:


underscorgan:
throwing a don't merge on here until I'm able to successfully test

underscorgan:
I've tested legacy/master ezbake builds, puppet-agent builds, internal-tools builds

underscorgan:
The main functionality change here is on packaging 1.0.x unset values for repo_name would default to the new packaging layout, but most of the more legacy repos where these tend to maybe be unset anyways were still on packaging master. The new puppet*(-nightly)? repos are the only ones using the new layout anyways, so let's be more explicit here.

PR #826, opened 2018-03-19
(RE-10428) Paths were wrong for nightlies

underscorgan:
Since the nightly ship tasks have been separated from the final ship
tasks, we can pass a variable through when we're working on a nightly
ship so we don't need to rely on the magic checking of whether or not
this is a final version.

### DIFF
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -58,15 +58,11 @@ module Pkg::Paths
   # shipped to the development/beta/non-final repo, if there is one defined.
   # Otherwise, default to the final repo name. We use this for more than just
   # shipping to the final repos, so we need this to not fail.
-  def repo_name
-    if Pkg::Util::Version.final?
-      Pkg::Config.repo_name || ""
+  def repo_name(nonfinal = false)
+    if nonfinal && Pkg::Config.nonfinal_repo_name
+      Pkg::Config.nonfinal_repo_name
     else
-      if Pkg::Config.nonfinal_repo_name
-        Pkg::Config.nonfinal_repo_name
-      else
-        Pkg::Config.repo_name || ""
-      end
+      Pkg::Config.repo_name || ""
     end
   end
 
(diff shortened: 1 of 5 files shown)


underscorgan:
I've tested this at least with shipping a nightly RPM to a test VM, did not test more extensively as it is very very slow to run retrieve over the vpn.

```
[root@o536rhrkr766tu4 ~]# cd /opt/repository-nightlies/yum/
[root@o536rhrkr766tu4 yum]# ll
total 0
drwxr-xr-x 3 root root 15 Mar 19 23:08 puppet6-nightly
lrwxrwxrwx 1 root root 45 Mar 19 23:08 puppet-nightly -> /opt/repository-nightlies/yum/puppet6-nightly
[root@o536rhrkr766tu4 yum]# find . -type f
./puppet6-nightly/el/7/x86_64/puppet-agent-5.4.0.571.g625403c-1.el7.x86_64.rpm
```

lib/packaging/paths.rb
```diff
@@ -82,14 +78,22 @@ def is_legacy_repo?(repo_name)
   # Method to determine the yum repo name. Maintains compatibility with legacy
   # projects, where `Pkg::Config.yum_repo_name` is set instead of
   # `Pkg::Config.repo_name`. Defaults to 'products' if nothing is set.
-  def yum_repo_name
+  def yum_repo_name(nonfinal = false)
+    if nonfinal
+      return Pkg::Config.nonfinal_repo_name || Pkg::Config.repo_name || Pkg::Config.yum_repo_name || 'products'
```
mwaggett: I think I want to fail if nonfinal_repo_name isn't set and you're explicitly trying to get the nonfinal repo name. Do you have a good reason for not doing that?

underscorgan: My reasoning for this was it's the same logic that is (and has been) in Pkg::Paths.repo_name

mwaggett: I think I'd rather switch to failing in that case. 
My reasons are:
1. We won't end up shipping things to the wrong repos again.
2. If something fails, it's not that big a deal because it's nightlies.

underscorgan:
@mwaggett updated

PR #830, opened 2018-03-23
(RE-10422) Make separate signing step before shipping to builds

sarameisburger:
This change modifies the packaging ship rake task to work with signing before shipping to builds. It takes a repo_dir that a user wants it to sign, copies the files into mozart, transfers them into the /pkg dir (this part i'm not 100% on because testing is hard in mozart... ideally it would just sign in whatever dir you give it but in case there is not an updated version of packaging pulled in it will be looking to sign files that are in the pkg/ dir so that's what i have it doing for now) signs them, and sends them back to your local pkg/. This commit relates to https://github.com/puppetlabs/vanagon/pull/549 where you can see packages that have gotten signed. Working on getting more tests for different package types.

### DIFF
--- a/tasks/sign.rake
+++ b/tasks/sign.rake
@@ -11,8 +11,9 @@ namespace :pl do
   end
 
   desc "Sign the Arista EOS swix packages, defaults to PL key, pass GPG_KEY to override or edit build_defaults"
-  task :sign_swix do
-    packages = Dir["pkg/**/*.swix"]
+  task :sign_swix, :root_dir do |_t, args|
+    swix_dir = args.root_dir || "pkg"
+    packages = Dir["#{swix_dir}/**/*.swix"]
     unless packages.empty?
       Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
       packages.each do |swix_package|
@@ -22,10 +23,11 @@ namespace :pl do
   end
 
   desc "Detach sign any solaris svr4 packages"
-  task :sign_svr4 do
-    unless Dir["pkg/**/*.pkg.gz"].empty?
+  task :sign_svr4, :root_dir do |_t, args|
+    svr4_dir = args.root_dir || "pkg"
+    unless Dir["#{svr4_dir}/**/*.pkg.gz"].empty?
       Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
-      Dir["pkg/**/*.pkg.gz"].each do |pkg|
+      Dir["#{svr4_dir}/**/*.pkg.gz"].each do |pkg|
         Pkg::Util::Gpg.sign_file pkg
       end
     end
@@ -97,13 +99,15 @@ namespace :pl do
   end
 
   desc "Sign ips package, uses PL certificates by default, update privatekey_pem, certificate_pem, and ips_inter_cert in build_defaults.yaml to override."
-  task :sign_ips do
-    Pkg::Sign::Ips.sign unless Dir['pkg/**/*.p5p'].empty?
+  task :sign_ips, :root_dir do |_t, args|
+    ips_dir = args.root_dir || "pkg"
+    Pkg::Sign::Ips.sign(ips_dir) unless Dir["#{ips_dir}/**/*.p5p"].empty?
   end
 
   desc "Sign built gems, defaults to PL key, pass GPG_KEY to override or edit build_defaults"
-  task :sign_gem do
-    gems = FileList["pkg/*.gem"]
+  task :sign_gem, :root_dir do |_t, args|
+    gems_dir = args.root_dir || "pkg"
+    gems = FileList["#{gems_dir}/*.gem"]
     gems.each do |gem|
       puts "signing gem #{gem}"
       Pkg::Util::Gpg.sign_file(gem)
@@ -111,9 +115,10 @@ namespace :pl do
   end
 
   desc "Check if all rpms are signed"
-  task :check_rpm_sigs do
+  task :check_rpm_sigs, :root_dir do |_t, args|
+    rpm_dir = args.root_dir || "pkg"
     signed = TRUE
-    rpms = Dir["pkg/**/*.rpm"]
+    rpms = Dir["#{rpm_dir}/**/*.rpm"]
     print 'Checking rpm signatures'
     rpms.each do |rpm|
       if Pkg::Sign::Rpm.has_sig? rpm
@@ -128,12 +133,13 @@ namespace :pl do
   end
 
   desc "Sign generated debian changes files. Defaults to PL Key, pass GPG_KEY to override"
-  task :sign_deb_changes do
+  task :sign_deb_changes, :root_dir do |_t, args|
     begin
-      change_files = Dir["pkg/**/*.changes"]
+      deb_dir = args.root_dir || "pkg"
+      change_files = Dir["#{deb_dir}/**/*.changes"]
       unless change_files.empty?
         Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
-        Pkg::Sign::Deb.sign_changes("pkg/**/*.changes")
+        Pkg::Sign::Deb.sign_changes("#{deb_dir}/**/*.changes")
       end
     ensure
       Pkg::Util::Gpg.kill_keychain
@@ -141,13 +147,15 @@ namespace :pl do
   end
 
   desc "Sign OSX packages"
-  task :sign_osx => "pl:fetch" do
-    Pkg::Sign::Dmg.sign unless Dir['pkg/**/*.dmg'].empty?
+  task :sign_osx, [:root_dir] => "pl:fetch" do |_t, args|
+    dmg_dir = args.root_dir || "pkg"
+    Pkg::Sign::Dmg.sign(dmg_dir) unless Dir["#{dmg_dir}/**/*.dmg"].empty?
   end
 
   desc "Sign MSI packages"
-  task :sign_msi => "pl:fetch" do
-    Pkg::Sign::Msi.sign unless Dir['pkg/**/*.msi'].empty?
+  task :sign_msi, [:root_dir] => "pl:fetch" do |_t, args|
+    msi_dir = args.root_dir || "pkg"
+    Pkg::Sign::Msi.sign(msi_dir) unless Dir["#{msi_dir}/**/*.msi"].empty?
   end
 
   ##
@@ -156,8 +164,10 @@ namespace :pl do
   #
   namespace :jenkins do
     desc "Sign all locally staged packages on #{Pkg::Config.signing_server}"
-    task :sign_all => "pl:fetch" do
-      Dir["pkg/*"].empty? and fail "There were files found in pkg/. Maybe you wanted to build/retrieve something first?"
+    task :sign_all, :root_dir do |_t, args|
+      Pkg::Util::RakeUtils.invoke_task('pl:fetch')
+      root_dir = args.root_dir || "pkg"
+      Dir["#{root_dir}/*"].empty? and fail "There were no files found in #{root_dir}. Maybe you wanted to build/retrieve something first?"
 
       # Because rpms and debs are laid out differently in PE under pkg/ they
       # have a different sign task to address this. Rather than create a whole
@@ -172,7 +182,7 @@ namespace :pl do
       rpm_sign_task = Pkg::Config.build_pe ? "pe:sign_rpms" : "pl:sign_rpms"
       deb_sign_task = Pkg::Config.build_pe ? "pe:sign_deb_changes" : "pl:sign_deb_changes"
       sign_tasks    = [rpm_sign_task]
-      sign_tasks    << deb_sign_task unless Dir['pkg/**/*.changes'].empty?
+      sign_tasks    << deb_sign_task unless Dir["#{root_dir}/**/*.changes"].empty?
       sign_tasks    << "pl:sign_tar" if Pkg::Config.build_tar
       sign_tasks    << "pl:sign_gem" if Pkg::Config.build_gem
       sign_tasks    << "pl:sign_osx" if Pkg::Config.build_dmg || Pkg::Config.vanagon_project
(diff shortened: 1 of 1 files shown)


tasks/sign.rake
```diff
@@ -97,23 +99,25 @@ namespace :pl do
...
       Pkg::Util::Gpg.sign_file(gem)
     end
   end
 
   desc "Check if all rpms are signed"
-  task :check_rpm_sigs do
+  task :check_rpm_sigs, :root_dir do
     signed = TRUE
-    rpms = Dir["pkg/**/*.rpm"]
+    rpms = Dir["#{gems_dir}/**/*.rpm"]
```
underscorgan: Missing a variable here

tasks/sign.rake
```diff
@@ -182,7 +191,9 @@ namespace :pl do
       sign_tasks    << "pl:sign_msi" if Pkg::Config.build_msi || Pkg::Config.vanagon_project
       remote_repo   = Pkg::Util::Net.remote_bootstrap(Pkg::Config.signing_server, 'HEAD', nil, signing_bundle)
       build_params  = Pkg::Util::Net.remote_buildparams(Pkg::Config.signing_server, Pkg::Config)
-      Pkg::Util::Net.rsync_to('pkg', Pkg::Config.signing_server, remote_repo)
+      Pkg::Util::Net.rsync_to(root_dir, Pkg::Config.signing_server, remote_repo)
+      # move our files into the /pkg directory for signing
+      Pkg::Util::Net.remote_ssh_cmd(Pkg::Config.signing_server, "cp -a #{remote_repo}/#{root_dir}/. #{remote_repo}/pkg")
```
underscorgan: So I think instead of having this `cp`, you can just have it rsync directly to pkg with something like 
`Pkg::Util::Net.rsync_to("#{root_dir}/", Pkg::Config.signing_server, "#{remote_repo}/pkg")` (untested though)

underscorgan: I'm curious if we can have it stay at the directory you send it to, I'm going to see if I can't figure out a reasonable test.

sarameisburger: That would definitely be ideal! But yeah, hard to test so I went with a hacky work around

tasks/sign.rake
```diff
@@ -128,26 +132,29 @@ namespace :pl do
...
       Pkg::Util::Gpg.kill_keychain
     end
   end
 
   desc "Sign OSX packages"
-  task :sign_osx => "pl:fetch" do
-    Pkg::Sign::Dmg.sign unless Dir['pkg/**/*.dmg'].empty?
+  task :sign_osx, [:root_dir] => "pl:fetch" do |_t, args|
+    osx_dir = args.root_dir || "pkg"
+    Pkg::Sign::Dmg.sign unless Dir['#{osx_dir}/**/*.dmg'].empty?
```
mwaggett: this needs double quotes instead of single quotes, otherwise the variable won't get interpolated.

tasks/sign.rake
sarameisburger: Whoops!

tasks/sign.rake
```diff
@@ -97,23 +99,25 @@ namespace :pl do
   end
 
   desc "Sign ips package, uses PL certificates by default, update privatekey_pem, certificate_pem, and ips_inter_cert in build_defaults.yaml to override."
-  task :sign_ips do
-    Pkg::Sign::Ips.sign unless Dir['pkg/**/*.p5p'].empty?
+  task :sign_ips, :root_dir do |_t, args|
+    ips_dir = args.root_dir || "pkg"
+    Pkg::Sign::Ips.sign unless Dir['#{ips_dir}/**/*.p5p'].empty?
```
mwaggett: You'll need to pass `ips_dir` to the sign method, since it assumes 'pkg' if nothing is specified.
Same for the dmg and msi ones.

tasks/sign.rake
```diff
@@ -128,26 +132,29 @@ namespace :pl do
...
+    osx_dir = args.root_dir || "pkg"
+    Pkg::Sign::Dmg.sign unless Dir['#{osx_dir}/**/*.dmg'].empty?
   end
 
   desc "Sign MSI packages"
-  task :sign_msi => "pl:fetch" do
-    Pkg::Sign::Msi.sign unless Dir['pkg/**/*.msi'].empty?
+  task :sign_msi, [:root_dir] => "pl:fetch" do |_t, args|
+    msi_dir = args.root_dir || "pkg"
+    Pkg::Sign::Msi.sign unless Dir['#{msi_dir}/**/*.msi'].empty?
```
mwaggett: same about the quotes

tasks/sign.rake
```diff
@@ -182,7 +191,9 @@ namespace :pl do
       sign_tasks    << "pl:sign_msi" if Pkg::Config.build_msi || Pkg::Config.vanagon_project
```
mwaggett: You'll want to pass `root_dir` to all these tasks.

tasks/sign.rake
jennaesq: Good catch... I tend to favor double quotes...  still new to ruby though; so these 'gotchas' are nice to read about.

Are single/double quotes favored in ruby?  Curious as to best practice.

mwaggett: I think if interpolation isn't needed, single quotes are preferred? But I don't think it matters that much. Consistency over everything, imo. 😄

underscorgan:
@sarameisburger \o/ I found a good test, if you want to stop copying things directly over to `pkg`. So the trick is it's going to use the Gemfile that was created at build time for the project, so I made these changes to the Gemfile in a test project
```
-gem 'vanagon', *location_for(ENV['VANAGON_LOCATION'] || '~> 0.15.5')
-gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || '~> 0.99')
+gem 'vanagon', *location_for(ENV['VANAGON_LOCATION'] || 'git@github.com:underscorgan/vanagon#add_sign_task')
+gem 'packaging', *location_for(ENV['PACKAGING_LOCATION'] || 'git@github.com:sarameisburger/packaging#add-signing-before')
```

and then it was using this packaging when trying to sign! I think having things in whatever directory they start out in would be good, want to give this a try?

tasks/sign.rake
sarameisburger: Yeah that's just how the file was when I got it so I'm assuming both ways work but +1 to consistency.

mwaggett: If there's string interpolation going on, only double quotes work.

sarameisburger:
Oo awesome, I'll try that out! @underscorgan

sarameisburger:
Updated so it's now signing in the desired root_dir on mozart

tasks/sign.rake
```diff
@@ -182,21 +192,21 @@ namespace :pl do
...
+      Pkg::Util::Net.rsync_to(root_dir, Pkg::Config.signing_server, remote_repo)
       rake_command = <<-DOC
 cd #{remote_repo} ;
 bundle_prefix= ;
 if [[ -r Gemfile ]]; then
   source /usr/local/rvm/scripts/rvm; rvm use ruby-2.3.1; bundle install --path .bundle/gems;
   bundle_prefix='bundle exec';
 fi ;
-$bundle_prefix rake #{sign_tasks.join(' ')} PARAMS_FILE=#{build_params}
+$bundle_prefix rake #{sign_tasks.join('#{root_dir} ')} PARAMS_FILE=#{build_params}
```
sarameisburger: @mwaggett updated

sarameisburger:
Gonna test with el 7

tasks/sign.rake
mwaggett: need double quotes for string interpolation. But I don't think that's going to work anyway. 
see https://robots.thoughtbot.com/how-to-use-arguments-in-a-rake-task

tasks/sign.rake
```diff
@@ -172,7 +182,7 @@ namespace :pl do
       rpm_sign_task = Pkg::Config.build_pe ? "pe:sign_rpms" : "pl:sign_rpms"
       deb_sign_task = Pkg::Config.build_pe ? "pe:sign_deb_changes" : "pl:sign_deb_changes"
       sign_tasks    = [rpm_sign_task]
-      sign_tasks    << deb_sign_task unless Dir['pkg/**/*.changes'].empty?
+      sign_tasks    << deb_sign_task unless Dir['#{root_dir}/**/*.changes'].empty?
```
mwaggett: need double quotes

sarameisburger:
sara@sara:~/puppetlabs/pl-build-tools-vanagon$ rpm -qpi pkg/el/7/x86_64/pl-rust-1.18.0-2.el7.x86_64.rpm
warning: pkg/el/7/x86_64/pl-rust-1.18.0-2.el7.x86_64.rpm: NOKEY, key ID ef8d349f
Name        : pl-rust                      Relocations: (not relocatable)
Version     : 1.18.0                            Vendor: Puppet Labs
Release     : 2.el7                         Build Date: Fri Mar 23 12:54:03 2018
Install Date: (not installed)            Build Host: mu693o0cex2rpv4.delivery.puppetlabs.net
Group       : System Environment/Base       Source RPM: pl-rust-1.18.0-2.el7.src.rpm
Size        : 360140621                        License: Apache 2.0 and MIT
Signature   : RSA/SHA512, Fri Mar 23 12:57:19 2018, Key ID 7f438280ef8d349f
URL         : https://www.puppetlabs.com
Summary     : Puppet Labs Rust
Architecture: x86_64
Description :
Puppet Labs Rust

mwaggett:
@sarameisburger I would try testing with a (non-default) root dir specified, to make sure these all work. Also, I'd try signing puppet-agent packages, since they have all the platforms. I think you're still missing arguments for the Ips, Dmg, and Msi sign methods - you'll want to pass the directory var to that method, otherwise it'll default to pkg.

sarameisburger:
@mwaggett I have been testing it with the 'output' directory that's being passed in via the vanagon file I linked `$bundle_prefix rake pl:sign_rpms[output] pl:sign_osx[output] pl:sign_swix[output] pl:sign_svr4[output] pl:sign_ips[output] pl:sign_msi PARAMS_FILE=/tmp/25175010380574314/1170357e3c0fc5d760bd45a9a2dac014fb4d8b15.yaml` the signed packages end up in pkg/ once signed, is that what you mean? I am currently working on testing other platforms, but building takes time so I wanted to make sure the logic was there before I built larger things, but other platforms will be tested. Maybe it's my Friday brain but I'm not sure what you mean about the ips, dmg, and msi methods, could you clarify?

mwaggett:
ahh gotcha, sounds good!

About the ips/dmg/msi stuff - the methods used inside those rake tasks, like `Pkg::Sign::Ips` take an optional argument for the target_dir. When not specified, it defaults to 'pkg'. Since we're now allowing different directories to be passed to the rake task, we should pass down that argument to the underlying method. So, for example, `Pkg::Sign::Ips.sign(ips_dir)`.

tasks/sign.rake
```diff
@@ -182,21 +192,21 @@ namespace :pl do
...
+      Pkg::Util::Net.rsync_to(root_dir, Pkg::Config.signing_server, remote_repo)
       rake_command = <<-DOC
 cd #{remote_repo} ;
 bundle_prefix= ;
 if [[ -r Gemfile ]]; then
   source /usr/local/rvm/scripts/rvm; rvm use ruby-2.3.1; bundle install --path .bundle/gems;
   bundle_prefix='bundle exec';
 fi ;
-$bundle_prefix rake #{sign_tasks.join(' ')} PARAMS_FILE=#{build_params}
+$bundle_prefix rake #{sign_tasks.join("([#{root_dir}] ")} PARAMS_FILE=#{build_params}
```
mwaggett: I think you have an extra `(` here?

underscorgan:
With the exception of the extra '(' this looks great! 🎉 Big 👍 assuming the rest of the tests worked as planned

tasks/sign.rake
```diff
@@ -182,21 +192,21 @@ namespace :pl do
...
+      Pkg::Util::Net.rsync_to(root_dir, Pkg::Config.signing_server, remote_repo)
       rake_command = <<-DOC
 cd #{remote_repo} ;
 bundle_prefix= ;
 if [[ -r Gemfile ]]; then
   source /usr/local/rvm/scripts/rvm; rvm use ruby-2.3.1; bundle install --path .bundle/gems;
   bundle_prefix='bundle exec';
 fi ;
-$bundle_prefix rake #{sign_tasks.join(' ')} PARAMS_FILE=#{build_params}
+$bundle_prefix rake #{sign_tasks.map{|task| task + "[#{root_dir}]"}.join(" ")} PARAMS_FILE=#{build_params}
```
sarameisburger: I had to change this because using the .join means the last sign_task wasn't appended with the root_dir so I changed it to a map appending all of the tasks and then joining those tasks with spaces

mwaggett: rubocop is angry 😭

mwaggett: but +1 to this change!

e-gris: Just delete rubocop. Problem solved. :-)

sarameisburger:
@e-gris I added the $DEFAULT_DIRECTORY variable and tested it by running the sign task with no directory specified and it populated correctly `$bundle_prefix rake pl:sign_rpms[pkg] pl:sign_osx[pkg] pl:sign_swix[pkg] pl:sign_svr4[pkg] pl:sign_ips[pkg] pl:sign_msi[pkg]`

tasks/sign.rake
```diff
@@ -10,9 +10,12 @@ namespace :pl do
     end
   end
 
+  # If no directory to sign is specified assume "pkg"
+  $DEFAULT_DIRECTORY = "pkg"
+
```
mwaggett: Since we use this in a lot of places, perhaps it would be useful to define this in the packaging lib and call something like `Pkg::DEFAULT_DIRECTORY` here.
Definitely not suggesting you change every occurrence of 'pkg' in this repo, but there are opportunities for future refactoring if this constant is accessible more globally.

sarameisburger: Yeah that's definitely a good thought...I think in the interest of keeping PRs small/focused I'll open a different ticket for that

sarameisburger: https://tickets.puppetlabs.com/browse/RE-10481 opened!

PR #831, opened 2018-03-23
(maint) Use 'ppc' in AIX artifact paths

mwaggett:
This commit updates the `parse_platform_tag` function to return 'ppc' as the
architecture for AIX platforms. Officially, 'power' is the AIX arch, but we use
'ppc' in paths and file names, to be consistent with upstream expectations.

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -82,7 +82,6 @@ module Pkg
           artifacts.each do |artifact|
             tag = Pkg::Paths.tag_from_artifact_path(artifact)
             platform, version, arch = Pkg::Platforms.parse_platform_tag(tag)
-            arch = 'ppc' if platform == 'aix'
             package_format = Pkg::Platforms.get_attribute(tag, :package_format)
 
             # Skip this if it's an unversioned MSI. We create these to help
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -345,6 +345,9 @@ module Pkg
       architecture = platform_tag.sub(/^(#{platform}-#{version}|#{codename})-?/, '')
 
       fail unless supported_arches.include?(architecture) || architecture.empty?
+
+      # AIX uses 'ppc' as its architecture in paths and file names
+      architecture = 'ppc' if platform == 'aix'
       return [platform, version, architecture]
     rescue
       raise "Could not verify that '#{platform_tag}' is a valid tag"


ScottGarman:
I was able to test this with a build and ship of the AIX pl-boost package, and it deployed to the correct url.

mwaggett:
http://pl-build-tools.delivery.puppetlabs.net/aix/6.1/ppc/pl-boost-1.58.0-7.aix6.1.ppc.rpm was shipped with this.

lib/packaging/paths.rb
```diff
@@ -177,6 +177,7 @@ def artifacts_path(platform_tag, path_prefix = 'artifacts', nonfinal = false)
 
     case package_format
     when 'rpm'
+      architecture = 'ppc' if platform == 'aix'
```
underscorgan: I'm wondering if it would be better to have this special-casing in `Pkg::Platforms.parse_platform_tag` ... do we just need this change for the artifacts_path, or do we want the architecture reporting as ppc everywhere? I see we have this hardcoded in Pkg::Config also...

mwaggett: I think if we wanted 'ppc' everywhere, we would've changed 'power' to 'ppc' in PLATFORM_INFO, right? iirc, 'power' is the canonical architecture, but 'pcc' is the expected notation in package names?

underscorgan: I meant ppc everywhere where we're using `parse_platform_tag`. It seems prone to bugs if we need to remember to special-case everywhere we're using that.

mwaggett: What would be the difference between that and changing PLATFORM_INFO?

Or I guess, what are the places we want to see 'power' and what are the places we want to see 'ppc'?

underscorgan: We don't want to change that in PLATFORM_INFO, the architecture for AIX is 'power'. We want it to be 'ppc' in artifact names/paths though.

mwaggett: right. I guess I was wondering if changing it in `parse_platform_tag` would have unforeseen consequences and/or be effectively the same as changing it in PLATFORM_INFO. I will investigate.

mwaggett: It looks like that specific info is not used anywhere in packaging other than to set https://github.com/puppetlabs/packaging/blob/1.0.x/lib/packaging/artifactory.rb#L183 so I'm assuming it's fine.

mwaggett:
@underscorgan updated this.

PR #832, opened 2018-03-28
(PA-1958) Rename redhat-fips platform to redhatfips

ScottGarman:
Having dashes in platform names causes problems when it comes to adding
a platform to pe_repo, so the redhat-fips platform name is being changed
to redhatfips.

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -213,7 +213,7 @@ module Pkg
         },
       },
 
-      'redhat-fips' => {
+      'redhatfips' => {
         '7' => {
           architectures: ['x86_64'],
           source_architecture: 'SRPMS',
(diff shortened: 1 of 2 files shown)


ScottGarman:
This PR and the one in puppet-agent should be reviewed and merged at the same time: https://github.com/puppetlabs/puppet-agent/pull/1387

mwaggett:
Can you also update references in spec tests please?

ScottGarman:
There were no previous references to the old name in the spec tests, but I've now added one new test in platform_spec.rb to cover this. Let me know if you had something else in mind.

mwaggett:
oh @ScottGarman my bad, I was looking at the 1.0.x branch of packaging, which will definitely need this.

PR #837, opened 2018-04-16
(maint) Skip tarball signing if signature already exists

shrug:
Currently the tarball signature will fail if the external
signature file already exists. This adds logic
to skip the tarball signing if it's already signed.

### DIFF
--- a/lib/packaging/util/gpg.rb
+++ b/lib/packaging/util/gpg.rb
@@ -51,6 +51,10 @@ module Pkg::Util::Gpg
       gpg ||= Pkg::Util::Tool.find_tool('gpg')
 
       if gpg
+        if File.exist? "#{file}.asc"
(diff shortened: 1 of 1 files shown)


lib/packaging/util/gpg.rb
```diff
@@ -51,6 +51,10 @@ def sign_file(file)
       gpg ||= Pkg::Util::Tool.find_tool('gpg')
 
       if gpg
+        if File.exist? "#{file}.asc"
+          warn "Signature on #{file} exists, skipping..."
+          return TRUE
```
e-gris: Minor: use of 'TRUE' is deprecated. 'true' works in all Rubies, IIRC.

shrug:
Retargeted at master in https://github.com/puppetlabs/packaging/pull/838

PR #844, opened 2018-04-18
(maint) Backport uber_ship_lite to packaging#master

underscorgan:


tasks/jenkins.rake
```diff
@@ -249,6 +249,23 @@ namespace :pl do
       end
     end
 
+		task :uber_ship_lite => "pl:fetch" do
```
sarameisburger: i'm not sure if this spacing here is correct but approving because the logic is there

PR #845, opened 2018-04-18
Change signing to run on the signing server instead of the distribution server

underscorgan:


underscorgan:
@sarameisburger updated!

PR #848, opened 2018-04-20
Updates to remove stickler and nexus configs

shrug:
As we're migrating to Artifactory, we're deprecating nexus and stickler. This removes those configs and stops shipping to those services.


underscorgan:
fire

PR #854, opened 2018-05-03
Have Pkg::Tar archive the working directory by default

ekinanp:


mwaggett:
so after grepping around, it looks like `Pkg::Config.files` is literally only used in this one place, so even though I said maybe not to before, I think setting the default more globally would be fine.

mwaggett:
for writing tests! 😁

ekinanp:
@mwaggett Nice catch! I just confirmed the same thing. Now that I think about it, I think I might like it being at `Pkg::Tar` just b/c it is more transparent there why we are setting it to that default.

mwaggett:
@ekinanp Have you tested the behavior you're looking to get from this? Your test makes sure `@files` gets assigned properly by default, but have you tested that that value does what you want?

ekinanp:
@mwaggett Thanks for making me double check, it turns out that Pkg::Util::File#install_files_into_dir expects passed-in globs (here '*') to resolve to only files, not directories (which is not very intuitive), so had to update the PR to make it work properly.

I tested the changes here out by cloning facter, having it point to my packaging, removing the files entry from project_data, then running package:tar and it works.

One thing -- hidden files are not archived by default when archiving the working directory b/c it seems like something that TAR does by default.

lib/packaging/tar.rb
```diff
@@ -14,6 +14,10 @@ def initialize
       @files    = Pkg::Config.files
       @target   = File.join(Pkg::Config.project_root, "pkg", "#{@project}-#{@version}.tar.gz")
 
+      # If the user did not specify any files, then archive the entire working directory
+      # instead (excluding 'pkg')
+      @files ||= Dir.glob('*') - ['pkg']
```
underscorgan: this should also exclude ext/packaging, the tarballs don't include the packaging checkout

ekinanp: @underscorgan https://github.com/puppetlabs/packaging/blob/1.0.x/tasks/tar.rake#L20-L23 already excluded it, but it seems like it'd be better to just have this be in Pkg::Tar.

Updated the PR to reflect that.

lib/packaging/tar.rb
```diff
@@ -14,6 +14,10 @@ def initialize
       @files    = Pkg::Config.files
       @target   = File.join(Pkg::Config.project_root, "pkg", "#{@project}-#{@version}.tar.gz")
 
+      # If the user did not specify any files, then archive the entire working directory
+      # instead
+      @files ||= Dir.glob('*')
```
e-gris: A kitten is sacrificed every time someone uses 'glob'

underscorgan: Is there a better way to gather everything in a directory?

e-gris: Not directly. It requires stepping much further back and being directive rather than inquisitive. It would be better if the instructions were "give me the files specified in this whitelist" rather than "give me whatever you happen to have lying around".

That is often easier said than done but worthwhile when it can be pulled off.

mwaggett: fwiw, this is a fallback/default case anyway. Ideally we'd be using something like https://github.com/puppetlabs/puppet/blob/master/ext/project_data.yaml#L10.

mwaggett: though then again, with this addition, I bet we won't use ^that anymore..

ekinanp: @e-gris The purpose of this PR is to default to "whatever you have lying around" if the user does not provide a "whitelist" -- it makes it so that I don't have to specify a "files" entry in ext/project_data.yaml when all I want is an archive of the working directory.

@mwaggett Yup, that's the goal. I have a PR up https://github.com/puppetlabs/puppet/pull/6813 that will remove that entry so that all of the gem files are contained in the .gemspec (originally, package:gem also used it). Also planning on doing something similar for hiera, mcollective soon.

ekinanp: I also talked about this stuff with @McdonaldSeanp

ekinanp:
Any chance we could get this merged soon? It's blocking some additional work on simplifying the component artifact generation + shipping job for the puppet-agent package validation pipeline.

PR #860, opened 2018-06-15
(RE-10926) Check for signature before signing rpms

mwaggett:
This commit adds a check to the `sign_rpms` task to skip signing if the rpm is already signed. Previously, if we tried to sign more than once, a non-zero exit code would be returned, which caused failures.

### DIFF
--- a/lib/packaging/sign/rpm.rb
+++ b/lib/packaging/sign/rpm.rb
@@ -37,4 +37,70 @@ module Pkg::Sign::Rpm
     %x(rpm -Kv #{rpm} | grep "#{Pkg::Util::Gpg.key.downcase}" &> /dev/null)
     $?.success?
   end
+
+  def sign_all(rpm_directory)
+    # Create a hash mapping full paths to basenames.
+    # This will allow us to keep track of the different paths that may be
+    # associated with a single basename, e.g. noarch packages.
+    all_rpms = {}
+    rpms_to_sign = Dir["#{rpm_directory}/**/*.rpm"]
+    rpms_to_sign.each do |rpm_path|
+      all_rpms[rpm_path] = File.basename(rpm_path)
+    end
+    # Delete a package, both from the signing server and from the rpm array, if
+    # there are other packages with the same basename so that we only sign the
+    # package once.
+    all_rpms.each do |rpm_path, rpm_filename|
+      if rpms_to_sign.map { |rpm| File.basename(rpm) }.count(rpm_filename) > 1
+        FileUtils.rm(rpm_path)
+        rpms_to_sign.delete(rpm_path)
+      end
+    end
+
+    v3_rpms = []
+    v4_rpms = []
+    rpms_to_sign.each do |rpm|
+      if has_sig? rpm
+        puts "#{rpm} is already signed, skipping . . ."
+        next
+      end
+      platform_tag = Pkg::Paths.tag_from_artifact_path(rpm)
+      platform, version, _ = Pkg::Platforms.parse_platform_tag(platform_tag)
+
+      # We don't sign AIX rpms
+      next if platform_tag.include?('aix')
+
+      sig_type = Pkg::Platforms.signature_format_for_platform_version(platform, version)
+      case sig_type
+      when 'v3'
+        v3_rpms << rpm
+      when 'v4'
+        v4_rpms << rpm
+      else
+        fail "Cannot find signature type for package '#{rpm}'"
+      end
+    end
+
+    unless v3_rpms.empty?
+      puts "Signing old rpms..."
+      legacy_sign(v3_rpms.join(' '))
+    end
+
+    unless v4_rpms.empty?
+      puts "Signing modern rpms..."
+      sign(v4_rpms.join(' '))
+    end
+
+    # Using the map of paths to basenames, we re-hardlink the rpms we deleted.
+    all_rpms.each do |link_path, rpm_filename|
+      next if File.exist? link_path
+      FileUtils.mkdir_p(File.dirname(link_path))
+      # Find paths where the signed rpm has the same basename, but different
+      # full path, as the one we need to link.
+      paths_to_link_to = rpms_to_sign.select { |rpm| File.basename(rpm) == rpm_filename && rpm != link_path }
+      paths_to_link_to.each do |path|
+        FileUtils.ln(path, link_path, :force => true, :verbose => true)
+      end
+    end
+  end
 end
--- a/lib/packaging/sign/rpm.rb
+++ b/lib/packaging/sign/rpm.rb
@@ -70,8 +70,7 @@ module Pkg::Sign::Rpm
       # We don't sign AIX rpms
       next if platform_tag.include?('aix')
 
-      sig_type = Pkg::Platforms.signature_format_for_platform_version(platform, version)
-      case sig_type
+      case Pkg::Platforms.signature_format_for_platform_version(platform, version)
       when 'v3'
         v3_rpms << rpm
       when 'v4'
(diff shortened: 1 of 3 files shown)


tasks/sign.rake
```diff
@@ -61,6 +61,7 @@ namespace :pl do
     v3_rpms = []
     v4_rpms = []
     rpms_to_sign.each do |rpm|
+      next if Pkg::Sign::Rpm.has_sig? rpm
```
demophoon: I'd rather have something like this not in the rake task so that we can better test it.

mwaggett: I thought about adding this to the `Pkg::Sign::Rpm.sign` method, but based on learnings from @e-gris, it seems preferable for checks like this to happen in whatever is consuming that method, not in the method itself.

What I *could* do is put all of this into its own function that the rake task calls. Do you think that would help your testing concerns?

shrug:
Should we be a little more verbose about what's happening?

lib/packaging/sign/rpm.rb
```diff
@@ -37,4 +37,70 @@ def has_sig?(rpm)
...
+        FileUtils.rm(rpm_path)
+        rpms_to_sign.delete(rpm_path)
+      end
+    end
+
+    v3_rpms = []
+    v4_rpms = []
+    rpms_to_sign.each do |rpm|
+      if has_sig? rpm
+        puts "#{rpm} is already signed, skipping . . ."
```
mwaggett: @shrug verbosity added

mwaggett:
@demophoon I moved the whole rake task to its own method. Do you have ideas for ways to test that don't involve stubbing literally everything?

demophoon:
I was thinking it could be something like `Pkg::Sign::Rpm.sign!` and have that be the method that throws an error when a package is already signed. 🤷‍♀️  

I've always been under the impression that your rake tasks should be as light weight as possible, especially when they mutate the state of the underlying machine they are running on. It allows us to mock things in the tests which is not really possible in rake files unfortunately.

It seems to be convention that `!` do dangerous things like raise exceptions so I think that would make sense here. Unfortunately since we already have a `sign` method that raises exceptions i'm not sure what would be the best way to approach this seeing that it would be a breaking change. Also I think that what you have here is alright since it allows us to start writing tests for these methods.

mwaggett:
totally, definitely glad to have an excuse to move stuff out of a rake task.

and yeah, that makes sense about `!` but we don't seem to be following that much in packaging as a whole. I'm hoping that what I've done is more clear wrt to method names and stuff.

I'm also open to breaking this method down further, to make testing easier, but wanted to get your thoughts first.

lib/packaging/sign/rpm.rb
```diff
@@ -37,4 +37,70 @@ def has_sig?(rpm)
...
+        next
+      end
+      platform_tag = Pkg::Paths.tag_from_artifact_path(rpm)
+      platform, version, _ = Pkg::Platforms.parse_platform_tag(platform_tag)
+
+      # We don't sign AIX rpms
+      next if platform_tag.include?('aix')
+
+      sig_type = Pkg::Platforms.signature_format_for_platform_version(platform, version)
+      case sig_type
```
e-gris: Nitpick. You could eliminate the throwaway variable 'sig_type'

```ruby
case Pkg::Platforms.signature_format_for_platform_version(platform, version)
```

lib/packaging/sign/rpm.rb
```diff
@@ -37,4 +37,70 @@ def has_sig?(rpm)
...
+      end
+    end
+
+    unless v3_rpms.empty?
+      puts "Signing old rpms..."
+      legacy_sign(v3_rpms.join(' '))
+    end
+
+    unless v4_rpms.empty?
+      puts "Signing modern rpms..."
```
e-gris: 'Modern' could change over time. I suggest making the the message something like:

puts "Signing v4 rpms"

(Similarly with the v3, above)

tasks/sign.rake
```diff
@@ -38,67 +38,8 @@ namespace :pl do
 
   desc "Sign mocked rpms, Defaults to PL Key, pass GPG_KEY to override"
   task :sign_rpms, :root_dir do |t, args|
-    rpm_dir = args.root_dir || $DEFAULT_DIRECTORY
```
e-gris: Hooray for moving code out of rakefiles!

lib/packaging/sign/rpm.rb
```diff
@@ -37,4 +37,69 @@ def has_sig?(rpm)
...
+      end
+    end
+
+    unless v3_rpms.empty?
+      puts "Signing legacy (v3) rpms..."
+      legacy_sign(v3_rpms.join(' '))
+    end
+
+    unless v4_rpms.empty?
+      puts "Signing modern (v4) rpms..."
```
mwaggett: @e-gris How's this?

lib/packaging/sign/rpm.rb
```diff
@@ -37,4 +37,69 @@ def has_sig?(rpm)
...
+        puts "#{rpm} is already signed, skipping . . ."
+        next
+      end
+      platform_tag = Pkg::Paths.tag_from_artifact_path(rpm)
+      platform, version, _ = Pkg::Platforms.parse_platform_tag(platform_tag)
+
+      # We don't sign AIX rpms
+      next if platform_tag.include?('aix')
+
+      case Pkg::Platforms.signature_format_for_platform_version(platform, version)
```
mwaggett: @e-gris updated

demophoon:
i'd also love to see tests on this, but it looks like our tests are broken at the moment :(

mwaggett:
@demophoon broken how so? 

I'll see if I can get this broken down more so we can test it.

demophoon:
Oh, nevermind about the tests being broken, it was something in my environment.

lib/packaging/sign/rpm.rb
sarameisburger: Is there a test for this?

mwaggett: I'm gonna work on adding more tests today. 😊

mwaggett: @sarameisburger alright, tests are done!

spec/lib/packaging/sign_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        end
+        expect(Pkg::Sign::Rpm).to_not receive(:legacy_sign)
+        expect(Pkg::Sign::Rpm).to_not receive(:sign)
+        Pkg::Sign::Rpm.sign_all(rpm_directory)
+      end
+
+      it 'deletes and relinks rpms with the same basename' do
+        allow(Dir).to receive(:[]).with("#{rpm_directory}/**/*.rpm").and_return(noarch_rpms)
+        allow(Pkg::Sign::Rpm).to receive(:sign)
+        expect(FileUtils).to receive(:rm).exactly(noarch_rpms.count/2).times
```
sarameisburger: what does this line do?

mwaggett: When we're signing noarch rpms, want to sign them only once, even though we ship them to both i386 and x86_64 directories, so before we sign, we delete them: https://github.com/puppetlabs/packaging/pull/860/files#diff-01f73a5a2f2ac5efbdffb935c95671e6R55
This test is ensuring the the right number of things are deleted, since that seemed easier than testing exactly what is deleted, since we don't really care exactly *which* rpm is deleted. I chose to use `noarch_rpms.count/2` instead of 3 so that if we happened to change what's in `noarch_rpms` it should Just Work™ (as long as `noarch_rpms` has two rpms for each platform).

Does that make sense?

sarameisburger: So you are testing that noarch_rpms.count is an even number by dividing is by 2? Or is that not what's happening with the /2 ?

mwaggett: no. so right now, `noarch_rpms` has 6 rpm paths in it, 2 each for el6, el7, and sles12. I'm expecting one from each pair to be deleted and then relinked, so 3 things aka 6/2 aka `noarch_rpms.count/2`.

Is that clearer? If not, I'm happy to go over this with you in person. 😊

sarameisburger: Ok, in Java %2 checks for an even number so I thought that's what was happening. And there will always be an even number of paths? I know this is a test so it doesn't matter as much I was more just curious.

PR #861, opened 2018-06-19
(maint) Bump rvm ruby version used in remote bundle install

mwaggett:
This commit changes the ruby used when bundle installing on remote hosts from
2.3.1 to 2.4.1. Due to a rubygems bug
(https://github.com/rubygems/rubygems/issues/1448), bundler was attempting to
install the packaging gem in the wrong path. I was only able to reproduce this
issue using ruby 2.3.1, so I am bumping to the next later version that is
installed on the jenkins workers.


mwaggett:
@e-gris These were my reasons for choosing 2.4.1:
1. Even though I wasn't able to reproduce with 2.5.1, the original reporter of https://github.com/rubygems/rubygems/issues/1448 claims that this affects ruby 2.5.1
2. 2.4.1 was the next greater version installed on the jenkins workers, hopefully minimizing any side affects of this change.

PR #866, opened 2018-06-28
(RE-11090) Use metadata

mwaggett:


lib/packaging/config.rb
```diff
@@ -37,6 +37,7 @@ def get_binding
       #
       def config_from_hash(data = {})
         data.each do |param, value|
+          param = 'builds_server' if param == 'host__internal-build-host'
```
mwaggett: I don't want to have to do this for (allthethings) but this is just showing that it works. I tested by making the check `param == host__rpm-builder` and running a retrieve - it was clearly attempting to retrieve from rpm-builder.d.pl.net so yay!

e-gris: :+1: Yay! Do you think making a transformer would be useful? I'm still not sure what the best approach might be...

mwaggett: transformer like.. the adapters you wrote? I'm still not sure how we'd do that without having to maintain a gross param map. but maybe that is what we'll need in the interim. 

hmm.. maybe we just want to add all the old params to `REASSIGNMENTS` https://github.com/puppetlabs/packaging/blob/1.0.x/lib/packaging/config/params.rb#L331 and slowly deprecate?

mwaggett: or add the new ones to be reassigned to the old ones?

e-gris: Yeah. Everything seems a bit icky...

mwaggett:
@underscorgan that makes sense to me. I think I'm gonna mess around with option 3 and see what that looks like!

Thanks for the input (you too, @e-gris)!

e-gris:
Certainly important, not wanting to avoid.

I agree, option 3 looks the best to try out first.

Similarly, _now_ is the time to mess around with our assumptions about key naming. Using the topic/filename as the namespace and allowing keys to be anything might be worth considering.

mwaggett:
@e-gris yeah, I was thinking about that too. Mostly I want it to be very clear what's what in our metadata, i.e. eliminate any ambiguity. I like your namespace idea, I'm thinking that could work.

mwaggett:
I was able to build and uber_ship (through the beginning of the uber_ship step, didn't actually ship anything) puppet5-nightly-release packages from https://github.com/puppetlabs/puppetlabs-release/pull/158, where build_defaults was removed.

Next steps, for a future PR, will be to only fetch from the build-data repo if metadata was not found (https://tickets.puppetlabs.com/browse/RE-11276).

PR #870, opened 2018-07-12
(PA-2016) Have package:gem noop if the project does not build a gem

ekinanp:
This makes more sense. It is also important for PA-2016 where we want
to simplify our component artifact + shipping job to not rely on jenkins:uber_build,
b/c jenkins:uber_build dynamically generates several jobs to do the work
that makes it difficult to diagnose failures.

Our simplified job will just run all of the possible packaging tasks in
sequence. If the specific packaging task does not apply to the given
component, the task will noop and pass-thru. We ship all of the
generated artifacts at the end.

### DIFF
--- a/tasks/gem.rake
+++ b/tasks/gem.rake
@@ -1,146 +1,149 @@
-if Pkg::Config.build_gem
-  require 'rubygems/package_task'
+require 'rubygems/package_task'
 
-  def glob_gem_files
-    gem_files = []
-    gem_excludes_file_list = []
-    gem_excludes_raw = Pkg::Config.gem_excludes.nil? ? [] : Pkg::Config.gem_excludes.split(' ')
-    gem_excludes_raw << 'ext/packaging'
-    gem_excludes_raw << 'pkg'
-    gem_excludes_raw.each do |exclude|
-      if File.directory?(exclude)
-        gem_excludes_file_list += FileList["#{exclude}/**/*"]
-      else
-        gem_excludes_file_list << exclude
-      end
-    end
-    files = FileList[Pkg::Config.gem_files.split(' ')]
-    files.each do |file|
-      if File.directory?(file)
-        gem_files += FileList["#{file}/**/*"]
-      else
-        gem_files << file
-      end
-    end
-    gem_files -= gem_excludes_file_list
-  end
-
-  def add_gem_dependency(opts = {})
-    spec = opts[:spec]
-    version = opts[:version]
-    type = opts[:type].to_s
-    gem = opts[:gem].to_s
-    if opts[:version].nil? or opts[:version].empty?
-      spec.send("add_#{type}_dependency".to_sym, gem)
+def glob_gem_files
+  gem_files = []
+  gem_excludes_file_list = []
+  gem_excludes_raw = Pkg::Config.gem_excludes.nil? ? [] : Pkg::Config.gem_excludes.split(' ')
+  gem_excludes_raw << 'ext/packaging'
+  gem_excludes_raw << 'pkg'
+  gem_excludes_raw.each do |exclude|
+    if File.directory?(exclude)
+      gem_excludes_file_list += FileList["#{exclude}/**/*"]
     else
-      spec.send("add_#{type}_dependency".to_sym, gem, version)
+      gem_excludes_file_list << exclude
     end
-    spec
+  end
+  files = FileList[Pkg::Config.gem_files.split(' ')]
+  files.each do |file|
+    if File.directory?(file)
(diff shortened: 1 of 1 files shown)


tasks/gem.rake
```diff
@@ -137,10 +137,22 @@ if Pkg::Config.build_gem
...
   namespace :pl do
     task :gem => "package:gem"
   end
+else
+  namespace :package do
+    # Here, we have an agent component that does not build a gem. We will have package:maybe_gem
+    # noop.
+    task :maybe_gem do
+      puts("The #{Pkg::Config.project} project does not build a gem! Skipping package:maybe_gem ...")
+    end
```
underscorgan: I have a pretty strong preference to just moving the `if Pkg::Config.build_gem` check into the `gem` task. There should be no reason that needs to be at the top level in this file, and we've made efforts in the past to move those into individual tasks as we've come to them.

ekinanp: Cool, I'll go ahead and do that. Is it safe for the task to no-op if Pkg::Config.build_gem is false? I was worried about breaking existing stuff, that's one of the reasons why I created a separate task too.

mwaggett: +1

underscorgan: yes, the task noop-ing should be fine

mwaggett:
since you're adding a bunch of new methods anyway, do you think you could put all that stuff in lib/packaging/gem.rb? We're trying to move our logic out of rake tasks and into the packaging lib.

ekinanp:
Working on it :)

ekinanp:
Ok, testing's done.

https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_PA_2016_test_puppet-agent-component-packaging-shipping_test-pa-2016/ is the job I tested this on.

You can see from the console outputs that both tasks executed in sequence. Comparing
http://builds.delivery.puppetlabs.net/facter/3.11.2/artifacts/
http://builds.delivery.puppetlabs.net/facter/b060b034cc5df492bb9ddab93f5c43dace968ca2/artifacts/

shows the right artifacts were shipped for Facter and comparing

http://builds.delivery.puppetlabs.net/puppet/5.5.1/artifacts/
http://builds.delivery.puppetlabs.net/puppet/c734ac63ddff3f7b3a28d5cf1ace5395432c7f1e/artifacts/

shows the same thing for Puppet.

Notice for Facter, package:gem no-oped.

I still would like to hold off on merging this until my team can review the actual Jenkins job that will use this PR (see if they have any issues with the debug output and what-not).

tasks/gem.rake
```diff
@@ -1,146 +1,153 @@
...
-      create_default_gem
-      if Pkg::Config.gem_platform_dependencies
-        create_platform_specific_gems
-      end
+namespace :package do
+  desc "Build a gem - All gems if platform specific"
+  task :gem => ["clean"] do
+    # This debug message is useful when we sequence this task with other
+    # packaging tasks, like package:tar.
+    puts("\nBuilding the gems for the #{Pkg::Config.project} project ...")
```
underscorgan: This seems like it would be better past the `unless Pkg::Config.build_gem` block, so it's only printed when you're actually building gems

mwaggett: +1

ekinanp: No that was on purpose -- our component artifact job needs to get some info. of when the gem task is invoked. I don't want to use package:gem either b/c pl:gem aliases to it, so it might be confusing. Also, it currently reads as
```
12:03:07 Building the gems for the facter project ...
12:03:07 The facter project does not build any gems! Passing through ...
```

which is not that bad.

underscorgan: I don't understand what you mean when you say `our component artifact job needs to get some info. of when the gem task is invoked`

ekinanp: When you invoke
```
rake task1 task2
```
rake runs the two tasks in sequence. It does not print to the output which task is run. It prints something like:
```
<task1_output>
<task2_output>
```

but we want
```
"Invoking task1"
<task1_output>
"Invoking task2"
<task2_output>
```

Hence why I have the print statements at the top. If there's a rakey way to get something like what I want, I'm all for it.

underscorgan: You'll want to run with `--trace`

```
% rake task1 task2 --trace
** Invoke task1 (first_time)
** Invoke print (first_time)
** Execute print
calling all prints!
** Execute task1
task1
** Invoke task2 (first_time)
** Invoke print
** Execute task2
task2
```

ekinanp: Nice! Updated the PR.

PR #874, opened 2018-07-18
(RE-11106) Promote artifacts to enterprise directory in artifactory

sarameisburger:
Based on user input in the jenkins promote job copy packages from developement to enterprise in artifactory.
1. Determines if user provided a SHA, SNAPSHOT, or plain 'ol tag
2. Search for the artifact(s)
3. Copy artifacts to enterprise directory

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -1,3 +1,5 @@
+require 'uri'
+
 module Pkg
 
   # The Artifactory class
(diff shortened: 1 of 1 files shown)


sarameisburger:
Calling promotion task in enterprise-dist: https://github.com/puppetlabs/enterprise-dist/pull/3377

lib/packaging/artifactory.rb
```diff
@@ -230,6 +232,66 @@ def package_name(platform_data, platform_tag)
...
+    # Depending on if it's an RPM or Deb package promote accordingly
+    # 'promote' by copying the package(s) to the enterprise directory on artifactory
+    #
+    # @param pkg [String] the package name ex. puppet-agent
+    # @param git_ref [String] full git SHA or tag of the package(s) to be promoted
+    # @param pe_version [String] enterprise version promoting to (XX.YY)
+    # @param platform_tag [String] the platform tag of the artifact
+    #   ex. el-7-x86_64, ubuntu-18.04-amd64
+    def promote_package(pkg, git_ref, pe_version, platform_tag)
+      package_ref = sha_or_tag(ref)
```
underscorgan: missed some renames here

lib/packaging/artifactory.rb
```diff
@@ -230,6 +232,66 @@ def package_name(platform_data, platform_tag)
...
+      if platform == "ubuntu"
+        codename = Pkg::Platforms.codename_for_platform_version(platform, version)
+        search_name = "#{pkg}_#{package_ref}*#{codename}_#{arch}"
+        noarch_search_name = "#{pkg}_#{package_ref}*#{codename}_all"
+      end
+      # search for the artifact, if we can't find it, try searching for the noarch version
+      artifacts_to_promote = Artifactory::Resource::Artifact.search(name: search_name, :artifactory_uri => @artifactory_uri)
+      if artifacts_to_promote.empty?
+        artifacts_to_promote = Artifactory::Resource::Artifact.search(name: noarch_search_name, :artifactory_uri => @artifactory_uri)
+      end
```
underscorgan: should we have a second check for whether or not `artifacts_to_promote` is empty and print a message saying nothing was found to promote? Not sure if that should be a failure case

lib/packaging/artifactory.rb
```diff
@@ -230,6 +232,66 @@ def package_name(platform_data, platform_tag)
...
+        # ex. puppet-agent-*01a4311-1.el7.x86_64
+        return "*" + package_ref[0..6] + "-1"
+      end
+      if package_ref.include?('SNAPSHOT')
+        # ex. puppetserver-5.3.4-0.1SNAPSHOT.2018.07.10T2051.el7.x86_64
+        return package_ref.sub(/\.SNAPSHOT/, "-0.1SNAPSHOT")
+      end
+      # ex. puppet-agent-5.3.7-1.el7.x86_64
+      return package_ref + "-1"
+    end
```
underscorgan: oh boy. just thought of another edge case.

http://builds.puppetlabs.lan/puppet-enterprise-vanagon/

lib/packaging/artifactory.rb
e-gris: This needs to be updated to agree with the last set of changes.

ref -> git_ref
sha_or_tag -> git_sha_or_tag

lib/packaging/artifactory.rb
mwaggett: I'm wondering if we can make use of <ref>.yaml for this, instead of trying to guess at package names.

lib/packaging/artifactory.rb
mwaggett: I'm thinking failing in that case would be good - we don't want to assume things promoted, since nothing failed, but then be missing packages.

mwaggett:
Tests?

lib/packaging/artifactory.rb
sarameisburger: Because this task is run by iterating through all of the master platforms I can either just put a message in here every time something isn't found to promote or I can keep track of a return/counter value or something in the enterprise-dist promote.rake and if nothing from all platforms was found I can have it fail there. Which would we prefer?

mwaggett: I think we'd still want to fail immediately if something is missing. It seems not ideal if we managed to promote for some platforms and not others?

lib/packaging/artifactory.rb
sarameisburger: I'm not too familiar with yaml related things, is there an example you can point me to?

lib/packaging/artifactory.rb
sarameisburger: When promoting will we always be promoting for all platforms? If so I'm good with doing this I just wasn't sure.

sarameisburger:
@mwaggett tested with the following edge cases:
SHA: `puppet-agent-5.5.1.356.g01a4311-1.el7.x86_64.rpm`, `puppet-agent_1.10.12.17.g0d1ac95-1xenial_amd64.deb`
Tag: `puppet-agent-5.3.7-1.el7.x86_64.rpm`
Snapshot (noarch): `puppetserver-5.3.4-0.1SNAPSHOT.2018.07.10T2051.el7.noarch.rpm`
Snapshot (all): `puppetserver_5.3.4-0.1SNAPSHOT.2018.07.16T1356xenial_all.deb`

You can see them here: https://artifactory.delivery.puppetlabs.net/artifactory/rpm__local/enterprise/2018.1/repos/el-7-x86_64/
and here: https://artifactory.delivery.puppetlabs.net/artifactory/debian__local/pool/enterprise/2018.1/repos/ubuntu-16.04-amd64/

lib/packaging/artifactory.rb
mwaggett: If we don't always promote for all platforms, we could end up with different versions of the same package in the same version of PE, which seems.. not great to me, especially if we were intending to promote everything and something secretly failed. ¯\_(ツ)_/¯

mwaggett:
We should probably also unit test this.

lib/packaging/artifactory.rb
sarameisburger: I'm not sure I understand what you mean, I'm looking at the yaml file but not seeing anything related to snapshots and git shas but if you think it could be used to parse the package_ref then maybe we can pair on it next week?

mwaggett: here's perhaps a better example, from a non-final version of something: http://builds.delivery.puppetlabs.net/puppetserver/6.0.0.master.SNAPSHOT.2018.07.18T2223/artifacts/6.0.0.master.SNAPSHOT.2018.07.18T2223.yaml

I'm saying that instead of trying to construct the name of the package using `git_sha_or_tag`, we can do something along the lines of `File.basename(platform_data['ubuntu-16.04-amd64'][:artifact]` to get the package name. Is that clearer? I'm happy to pair if you're still confused!

sarameisburger: Hmm I'm not sure I understand. Let's table it for Monday.

lib/packaging/artifactory.rb
shrug: Not every package is promoted for every platform (razor, for example)

sarameisburger:
Gonna close this for now because it'll probably change quite a bit when we use AQL

PR #876, opened 2018-07-27
(RE-11090) Use metadata: Take 2

mwaggett:

### DIFF
--- a/lib/packaging/config.rb
+++ b/lib/packaging/config.rb
@@ -214,15 +214,12 @@ module Pkg
           warn "Failed to retrieve metadata: #{e}"
         end
 
-        if got_config
-          return self.config
-        else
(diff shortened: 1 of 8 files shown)


lib/packaging/config.rb
```diff
@@ -194,13 +203,21 @@ def load_default_configs
...
+          self.config_from_metadata
+          got_config = true
+        rescue => e
+          warn "Failed to retrieve metadata: #{e}"
+        end
+
         if got_config
-          self.config
+          return self.config
         else
```
e-gris: SillyPetPeeve[tm]: 'else' not needed after 'return'

mwaggett:
I was able to properly bundle install via the puppet-agent repo, which doesn't have a specified dependency on 'releng_metadata'. I can see that the gem is properly getting pulled in via packaging now.
I also was able to kick off a rake task and see that the metadata was properly fetched (got all those `Warning - No build data parameter found for 'aix_6.1_architectures'. Perhaps you have an erroneous entry in your yaml file?` warnings).

e-gris:
W00t!

PR #880, opened 2018-08-01
(maint) Add helpful error messages when build targets are missing

mwaggett:


lib/packaging/config.rb
```diff
@@ -353,8 +353,10 @@ def mock_to_dist_version_arch(mock)
 
       def deb_build_targets
         if self.vanagon_project
+          fail "ERROR: Could not find any deb targets. Try adding `deb_targets` to your build_defaults.yaml." unless self.deb_targets
```
shrug: I think this might break things like pe-razor-server where we only build/promote RPMs. Maybe only fail if they're both empty? Not sure, but we definitely have some cases where we only promote for a subset of PE masters.

mwaggett: ooh right. hmm. I think failing if they're both empty is gonna be a more-involved change..

I could just change these to warnings..

shrug: That's probably the best idea for now, at least have some useful output when it explodes :)

mwaggett: hmm @shrug now that I'm thinking about this.. This was already failing in the same scenario, i.e. it would try to split nil targets and fail there. so we must have something in place to prevent that for pe-razor-server? I can dig.

mwaggett: ahh, we just set deb_targets to an empty string: https://github.com/puppetlabs/razor-vanagon/blob/master/ext/build_defaults.yaml#L12

shrug: ah that makes sense. So yeah this works then.

demophoon: Would it make sense to make this nor hard fail still? It seems that us setting `deb_targets` to an empty string is just a work around for this behavior.

mwaggett: It definitely is a work-around.. but I'm not sure what's better.
I think adding `deb_targets` as an empty string is not a *bad* way to be intentional about not shipping debs. Not having `deb_targets` at all, in your scenario, could be intentional or it could be a mistake.

I can definitely change this, but I want us to be aware of potential consequences. What do you think?

demophoon: I personally think that it should be a warning because the way that the failure reads makes it seem like you cannot build anything without specifying at least one deb target which isn't true since we intend to work around it with something not apparent looking at the build data.

That is just my opinion though.

mwaggett: That makes sense.

Do you prefer warning over having the failure specifically state that you should set `deb_targets` to an empty string?

demophoon: Maybe just change the error since it is what was already there and this will just codify the behavior.

lib/packaging/config.rb
```diff
@@ -353,8 +353,10 @@ def mock_to_dist_version_arch(mock)
 
       def deb_build_targets
         if self.vanagon_project
+          fail "ERROR: Could not find any deb targets. Try adding `deb_targets` to your build_defaults.yaml. If you don't want to build any debs, set this to an empty string." unless self.deb_targets
```
mwaggett: @demophoon better?

PR #881, opened 2018-08-03
(maint) Make capture3 work with windows C:/Program Files

McdonaldSeanp:
This commit adds functionality to work around the possible inclusion of
C:/Program Files in the path to an executable. It does so using the
%ProgramFiles% environment var usually available on windows.


McdonaldSeanp:
Bad approach, closing

PR #884, opened 2018-08-14
RE-11145 Update Changelog for 0.99.10

e-gris:


mwaggett:
I would do a merge-up from master once that PR goes in, so that the 0.6.8 changelog gets included in this.

mwaggett:
also, is that Distelli check new?? am I crazy?? wherever it came from, it needs to use a later ruby..

e-gris:
"Merge-up from master" -- not sure what this means exactly. Do you mean merging master to 1.0.x before merging this PR?

mwaggett:
@e-gris yes

PR #887, opened 2018-08-20
(RE-10177) Remove bootstrap functionality

mwaggett:
This commit removes all invocations of `rake package:bootstrap` and replaces
them with `bundle install`s. As we are moving towards everything using the
packaging gem, package:bootstrap is no longer needed.

### DIFF
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -359,12 +359,12 @@ module Pkg::Util::Net
 #{tar} -zxvf /tmp/#{tarball_name}.tar.gz -C /tmp/ ;
 git clone --recursive /tmp/#{tarball_name} /tmp/#{Pkg::Config.project}-#{appendix} ;
 cd /tmp/#{Pkg::Config.project}-#{appendix} ;
-bundle_prefix= ;
 if [[ -r Gemfile ]]; then
-  source /usr/local/rvm/scripts/rvm; rvm use ruby-2.4.1; bundle install --path .bundle/gems ;
-  bundle_prefix='bundle exec' ;
+  source /usr/local/rvm/scripts/rvm; rvm use ruby-2.4.1; bundle install --path .bundle/gems --binstubs .bundle/bin ;
+else
+  echo "ERROR: Couldn't read Gemfile, can't bundle install."
+  exit 1
 fi ;
-$bundle_prefix rake package:bootstrap
 DOC
       Pkg::Util::Net.remote_ssh_cmd(host, command)
       "/tmp/#{Pkg::Config.project}-#{appendix}"
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -53,13 +53,13 @@
 #     pushd git_repo
 #
 #     ### Clone the packaging repo
-#     rake package:bootstrap
+#     bundle install --path .bundle/gems --binstubs .bundle/bin --retry 3
 #
 #     ### Perform the build
-#     rake pl:build_from_params PARAMS_FILE=$WORKSPACE/BUILD_PROPERTIES
+#     bundle exec rake pl:build_from_params PARAMS_FILE=$WORKSPACE/BUILD_PROPERTIES
 #
 #     ### Send the results
-#     rake pl:jenkins:ship["artifacts"] PARAMS_FILE=$WORKSPACE/BUILD_PROPERTIES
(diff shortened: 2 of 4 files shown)


lib/packaging/util/net.rb
```diff
@@ -362,9 +362,9 @@ def remote_bootstrap(host, treeish, tar_cmd = nil, tarball = nil)
 bundle_prefix= ;
 if [[ -r Gemfile ]]; then
   source /usr/local/rvm/scripts/rvm; rvm use ruby-2.4.1; bundle install --path .bundle/gems ;
-  bundle_prefix='bundle exec' ;
+else
+  echo "ERROR: Couldn't read Gemfile, can't bundle install."
```
underscorgan: I think this is going to break ezbake projects until things are cleared up with ezbake using packaging as a gem / all ezbake projects are updated.

Also probably want to actually raise an error in this case, not let the build continue

lib/packaging/util/net.rb
```diff
@@ -362,9 +362,9 @@ def remote_bootstrap(host, treeish, tar_cmd = nil, tarball = nil)
 bundle_prefix= ;
```
underscorgan: bundle_prefix is no longer used

lib/packaging/util/net.rb
mwaggett: oops, forgot the `exit 1` I was gonna put in there.

and yeah, this will break ezbake projects at the moment, but I figure we can either sit on merging this or just not do a packaging release 'til everything is updated. does that seem doable?

underscorgan: not doing a packaging release won't help here since ezbake builds just pull from 1.0.x

mwaggett: right, until ezbake gets updated.

templates/repo.xml.erb
```diff
@@ -71,16 +71,16 @@ if [ $PACKAGE_BUILD_RESULT -eq 0 ] ; then
     pushd git_repo
 
       ### Clone the packaging repo
-      rake package:bootstrap --trace
+      bundle install --path .bundle/gems --bin .bundle/bin --retry 3
```
underscorgan: `--bin` isn't a valid param for `bundle install`

templates/packaging.xml.erb
```diff
@@ -146,13 +146,13 @@ pushd project
   pushd git_repo
 
     ### Clone the packaging repo
-    rake package:bootstrap
+    bundle install --path .bundle/gems --bin .bundle/bin --retry 3
```
underscorgan: `--bin` isn't a valid param for `bundle install`

tasks/jenkins.rake
```diff
@@ -53,13 +53,13 @@
 #     pushd git_repo
 #
 #     ### Clone the packaging repo
-#     rake package:bootstrap
+#     bundle install --path .bundle/gems --bin .bundle/bin --retry 3
```
underscorgan: `--bin` isn't a valid param for `bundle install`

templates/repo.xml.erb
mwaggett: oh shit, I really thought I had taken those out, but I guess I took them out of somewhere else 😝

mwaggett: @underscorgan Do you think we should use `--binstubs` or not worry about bin at all? In places where we were already bundle installing, like for remote execution, we only specify `--path`.

underscorgan: `--binstubs` seems like a good thing to add in

mwaggett:
@underscorgan I used this branch for the Puppet Platform release last week (with some minor tweaks, since facter didn't have a Gemfile at the time).

underscorgan:
@mwaggett do you have a diff between this and what you used to ship? If not, can you go through at least signing in a test build to make sure it doesn't 💥 ? This had bad bundler tags in it, just want to make sure there weren't other typos that got fixed there but not here.

mwaggett:
@underscorgan I don't have a diff, but I think what I had was the first commit of this PR and then in the place where I'm now failing if there's no Gemfile, I added an install of the packaging gem so that I could ship facter. I will do the beginning of an uber_ship rn to make sure things are okay.

mwaggett:
okay @underscorgan I was able to get through the retrieve and sign_all steps of an uber_ship from puppet-agent#5.5.6 with PACKAGING_LOCATION set to my local repo.

mwaggett:
I'm gonna close this - I don't think we're ready to start failing when Gemfiles are missing and the additional work that was in this PR was merged in https://github.com/puppetlabs/packaging/pull/890.

PR #888, opened 2018-08-22
(PA-2001) Add support for SLES 15

ScottGarman:


ScottGarman:
I'm assuming the additions to the rake tasks are okay to add now and won't cause errors if we try to ship before SLES 15 is released. Lemme know if that's not correct and I can remove them.

ScottGarman:
Closing this, targeted the wrong branch.

PR #897, opened 2018-09-12
(maint) Don't check if AIX packages have signatures

underscorgan:
Running a few tests here.


underscorgan:
Running through a mock uber_ship I was able to replicate the failure and confirm this change fixes the failure.

PR #898, opened 2018-09-17
(packaging) reference gem path when looking for PackageInfo.plist

ccaviness:
PackageInfo.plist is part of the packaging gem, which is no longer vendored inside of puppet during the build process. This change uses the installed package gem path as location for the file.

### DIFF
--- a/tasks/apple.rake
+++ b/tasks/apple.rake
@@ -44,6 +44,7 @@ end
 #               in the ext/osx directory.
 def make_directory_tree
   project_tmp    = "#{Pkg::Util::File.mktemp}/#{@package_name}"
+  packaging_dir  = Gem::Specification.find_by_name('packaging').gem_dir
   @scratch       = "#{project_tmp}/#{@title}"
   @working_tree  = {
      'scripts'   => "#{@scratch}/scripts",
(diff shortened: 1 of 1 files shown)


tasks/apple.rake
```diff
@@ -44,6 +44,7 @@ end
 #               in the ext/osx directory.
 def make_directory_tree
   project_tmp    = "#{Pkg::Util::File.mktemp}/#{@package_name}"
+  packaging_dir  = Gem::Specification.find_by_name('packaging').gem_dir
```
mwaggett: In the past, in order to use the packaging repo, projects needed to include a `package:bootstrap` rake tasks that would clone the packaging repo into <project_root>/ext/packaging. 
We are moving away from this method, in favor of using packaging as a gem, but I wonder what happens here if we still use the bootstrap method, i.e. the packaging gem is not installed.

mwaggett:
@ccaviness Could you rebase this onto 1.0.x? I'm trying to test with your branch and it appears to be out of date.

ccaviness:
@mwaggett I think I've rebased successfully? I only vaguely understand git.

tasks/apple.rake
ccaviness: See [PUP-9148](https://tickets.puppetlabs.com/browse/PUP-9148)

I'm using a build method based on [this](https://groups.google.com/forum/#!msg/puppet-users/_0nePXOvcEE/RdSz3t7LgmUJ) 5-year-old post, and that method no longer works because the `Rakefile` changes to use `packaging` as a gem. This is my workaround.

If you'd like, I could have `packaging_dir` be the `gem_dir` if it exists, or the old path if it doesn't?

mwaggett:
@ccaviness thanks, rebase looks good to me!

You said this works for you locally? Can you tell me what you're running and where? e.g. are you running from the puppet repo or somewhere else? which branch? what command exactly are you running? what's the output? how have you ensured that your branch of packaging is being used?

ccaviness:
Let me break apart our bizarro bespoke build system and see if I can get steps to reproduce for you.

shrug:
@ccaviness Have you had a chance to look into this yet?

ccaviness:
Not yet - work got busy and I'm currently traveling out of country. I hope to work on this later next week.

mwaggett:
@ccaviness Do you plan to return to this work?

ccaviness:
I apologize, I haven't had time to work on this. I am planning to work on updating to 6.x in the near future, but at this point I don't think I can commit to any specific timeline. If you'd like, you can close this pull request and I'll come back with a new one once I have made the move to 6.x.

PR #900, opened 2018-09-25
(maint) Add quotes around PACKAGING_LOCATION value when exporting

mwaggett:
This commit adds single quotes around the value of ENV['PACKAGING_LOCATION']
when it is exported as part of the remote bundle install process. Previously,
if the value had spaces in it (e.g. git://github.com/mwaggett/packaging.git
--branch=foo), everything after the space would get left out of the export and
thing would start failing, since the remainder is not a valid command.


lib/packaging/util/net.rb
```diff
@@ -372,7 +372,7 @@ def remote_bootstrap(host, treeish, tar_cmd = nil, tarball = nil)
 
     def remote_bundle_install_command
       export_packaging_location = ''
-      export_packaging_location = "export PACKAGING_LOCATION=#{ENV['PACKAGING_LOCATION']};" if ENV['PACKAGING_LOCATION'] && !ENV['PACKAGING_LOCATION'].empty?
+      export_packaging_location = "export PACKAGING_LOCATION='#{ENV['PACKAGING_LOCATION']}';" if ENV['PACKAGING_LOCATION'] && !ENV['PACKAGING_LOCATION'].empty?
       command = "source /usr/local/rvm/scripts/rvm; rvm use ruby-2.4.1; #{export_packaging_location} bundle install --path .bundle/gems ;"
```
e-gris: LGTM. FWIW, (feel free to ignore) I would probably do it this way:

```ruby
command = "source /usr/local/rvm/scripts/rvm; rvm use ruby-2.4.1; "
if ENV['PACKAGING_LOCATION'] && !ENV['PACKAGING_LOCATION'].empty?
  command += "env PACKAGING_LOCATION='#{ENV['PACKAGING_LOCATION']}' "
end
command += "bundle install --path .bundle/gems;"
```

I'm assuming that PACKAGING_LOCATION unset and PACKAGING_LOCATION == '' are two different conditions and you want to preserve that.

I like using "env FOO=bar BAZ=blech <command>" rather than "export", the interface is a bit nicer especially when multiple variables are involved.

mwaggett:
@shrug travis is finished, if you click 'Details' - not sure why it's been in the pending state here for so long..

PR #901, opened 2018-09-26
Added platform support for Ubuntu 18.10 Cosmic

sebastian-miclea:
Added support for Ubuntu 18.10 cosmic cuttlefish.
I need this change in order be able to build pl-build-tools-release package.

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -294,6 +294,14 @@ module Pkg
           source_package_formats: DEBIAN_SOURCE_FORMATS,
           repo: true,
         },
+        '18.10' => {
+          codename: 'cosmic',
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -20,3 +20,7 @@ def reset_env(keys)
     ENV[key] = nil
   end
 end
+
+RSpec.configure do |config|
+  config.color = true
```
e-gris: This should be in the user's "~/.rspec" file rather in the spec_helper.rb

Having it here can make make output ugly on Jenkins & terminals not expecting it

sebastian-miclea: Updated. Thanks!

PR #907, opened 2018-10-09
(PE-16465) Update logic to create the "all the agents" tarball

shrug:
When the surrounding logic changed to use the yaml config for
generating the list of platforms (it used to be hardcoded in the job),
the existing code that generates the tarball containing all of the
agents broke. This moves the bulk of the logic out of the rake task
and into the library, and updates to work with the newer layout.

### DIFF
--- a/lib/packaging/repo.rb
+++ b/lib/packaging/repo.rb
@@ -33,6 +33,34 @@ module Pkg::Repo
       end
     end
 
+    def create_combined_repo_archive(project, versioning)
+      tar = Pkg::Util::Tool.check_tool('tar')
+      platforms = Pkg::Config.platform_repos
+      platforms.each do |platform|
+        archive_name = "#{project}-#{platform['name']}"
+        Dir.chdir("pkg") do
+          if versioning == 'ref'
+            local_target = File.join(Pkg::Config.project, Pkg::Config.ref, "repos")
+          elsif versioning == 'version'
+            local_target = File.join(Pkg::Config.project, Pkg::Util::Version.dot_version, "repos")
+          end
+
+          Dir.chdir(local_target) do
+            if !Pkg::Util::File.exist?("#{archive_name}.tar.gz")
+              warn "Skipping #{archive_name} because it (#{archive_name}.tar.gz) has no files"
+            else
+              if File.exist?("#{Pkg::Config.project}-all.tar")
+                tar_cmd = "--update"
+              else
+                tar_cmd = "--create"
+              end
+              Pkg::Util::Execution.ex("#{tar} --owner=0 --group=0 #{tar_cmd} --file #{Pkg::Config.project}-all.tar #{archive_name}.tar.gz")
+            end
+          end
+        end
+      end
+    end
+
     def directories_that_contain_packages(artifact_directory, pkg_ext)
       cmd = "[ -d #{artifact_directory} ] || exit 1 ; "
       cmd << "pushd #{artifact_directory} > /dev/null && "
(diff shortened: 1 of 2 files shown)


lib/packaging/repo.rb
```diff
@@ -33,6 +33,34 @@ def create_all_repo_archives(project, versioning)
...
+          if versioning == 'ref'
+            local_target = File.join(Pkg::Config.project, Pkg::Config.ref, "repos")
+          elsif versioning == 'version'
+            local_target = File.join(Pkg::Config.project, Pkg::Util::Version.dot_version, "repos")
+          end
+
+          Dir.chdir(local_target) do
+            if !Pkg::Util::File.exist?("#{archive_name}.tar.gz")
+              warn "Skipping #{archive_name} because it (#{archive_name}.tar.gz) has no files"
+            else
```
e-gris: Style suggestion: 

  * put main logic flow outside of "else" statements.
  * use set/override rather than if/else
  * Refactor abbreviations


```ruby
if !Pkg::Util::File.exist?("#{archive_name}.tar.gz")
  warn "Skipping ..."
  break
end
tar_sub_command = "--create"
tar_sub_command = "--update" if File.exist?("#{Pkg::Config.project}-all.tar")
```

tasks/nightly_repos.rake
```diff
@@ -79,32 +79,10 @@ DOC
...
-            if File.exist?("#{Pkg::Config.project}-all.tar")
-              tar_cmd = "--update"
-            else
-              tar_cmd = "--create"
-            end
-            Pkg::Util::Execution.ex("#{tar} --owner=0 --group=0 #{tar_cmd} --file #{Pkg::Config.project}-all.tar #{name_of_archive}.tar.gz")
-          end
-        end
-      end
+      Pkg.Repo.create_combined_repo_archive(name_of_archive, versioning)
```
mwaggett: this should be `Pkg::Repo.<method_name>` (i.e. `::` instead of `.`)

lib/packaging/repo.rb
```diff
@@ -33,6 +33,34 @@ def create_all_repo_archives(project, versioning)
...
+      tar = Pkg::Util::Tool.check_tool('tar')
+      platforms = Pkg::Config.platform_repos
+      platforms.each do |platform|
+        archive_name = "#{project}-#{platform['name']}"
+        Dir.chdir("pkg") do
+          if versioning == 'ref'
+            local_target = File.join(Pkg::Config.project, Pkg::Config.ref, "repos")
+          elsif versioning == 'version'
+            local_target = File.join(Pkg::Config.project, Pkg::Util::Version.dot_version, "repos")
+          end
```
mwaggett: I think `Pkg::Config.ref` should always give us what we want? unless there's some context I'm missing..

lib/packaging/repo.rb
```diff
@@ -33,6 +33,34 @@ def create_all_repo_archives(project, versioning)
...
+
+          Dir.chdir(local_target) do
+            if !Pkg::Util::File.exist?("#{archive_name}.tar.gz")
+              warn "Skipping #{archive_name} because it (#{archive_name}.tar.gz) has no files"
```
mwaggett: This warning is confusing to me.. Is this a case where we already have what we want or a case where something went wrong?

lib/packaging/repo.rb
underscorgan: I think we need `dot_version` here for getting the version directories to be `<tag>.<commits since last tag>`

mwaggett: oh yeah, I forgot that's what the directories on agent-downloads look like. I was just thinking of builds, where everything is either a git sha or a tag.

sarameisburger:
@shrug do you wanna keep this open? Or close until we can resume work on it

PR #909, opened 2018-10-12
(maint) Add method for determining release package link paths

mwaggett:
This commit adds a `release_package_link_path` method that takes a platform tag
and constructs the the target path for the top-level release package symlink
associated with that platform, taking into account whether this is a final or
nonfinal (i.e. nightly) release package.
Currently, updating these symlinks is a manual process. This is the first step
in an effort to automate these updates.

### DIFF
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -265,4 +265,21 @@ module Pkg::Paths
       raise "Not sure what to do with a package format of '#{package_format}'"
     end
   end
+
+  def release_package_link_path(platform_tag, nonfinal = false)
+    platform, version, arch = Pkg::Platforms.parse_platform_tag(platform_tag)
+    package_format = Pkg::Platforms.package_format_for_tag(platform_tag)
+    case package_format
+    when 'rpm'
+      base_path = nonfinal ? Pkg::Config.nonfinal_yum_repo_path : Pkg::Config.yum_repo_path
+      return File.join(base_path, "#{repo_name(nonfinal)}-release-#{platform}-#{version}.noarch.rpm")
+    when 'deb'
+      base_path = nonfinal ? Pkg::Config.nonfinal_apt_repo_path : Pkg::Config.apt_repo_path
+      codename = Pkg::Platforms.codename_for_platform_version(platform, version)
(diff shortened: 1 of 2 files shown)


lib/packaging/paths.rb
```diff
@@ -268,4 +268,21 @@ def repo_config_path(platform_tag)
...
+    when 'rpm'
+      base_path = nonfinal ? Pkg::Config.nonfinal_yum_repo_path : Pkg::Config.yum_repo_path
+      return File.join(base_path, repo_name(nonfinal), "#{repo_name(nonfinal)}-release-#{platform}-#{version}.noarch.rpm")
+    when 'deb'
+      base_path = nonfinal ? Pkg::Config.nonfinal_apt_repo_path : Pkg::Config.apt_repo_path
+      codename = Pkg::Platforms.codename_for_platform_version(platform, version)
+      return File.join(base_path, "#{repo_name(nonfinal)}-release-#{codename}.deb")
+    else
+      warn "No release packages for package format '#{package_format}', skipping . . ."
+      return nil
```
underscorgan: Right now it looks like the nightly repos are real inconsistent about where the release packages are linked to ... yum has `puppet-nightly` release packages at the top level (`Pkg::Config.nonfinal_yum_repo_path`) and under the repo name directory, puppet5-nightly and puppet6 nightly are only under the repo name directory.
For apt, all nightly release packages are both at the top level and under the repo name directory.

These changes look like they will only create links under the repo name for yum and under the root level for apt. Are there plans to clean up the other release packages? What if anything are we going to break if that happens? Is there going to be a different method for creating the nightly links for the main repos (yum.puppetlabs.com/puppet-nightly/<package>)

underscorgan: No, there aren't specific changes I want to see here. I was trying to make sure the existing release package locations are the intended ones, and that this isn't intended to cover the other cases. Mostly, with all the inconsistencies it's hard to tell what's intentional and what ended up somewhere by mistake.

mwaggett: That's totally fair. Maybe we should have a larger conversation about where we want these links to live?

mwaggett:
alright, this has been updated to reflect the conversation we had this morning - only top-level links.

PR #917, opened 2018-11-27
PA-2232 Added Fedora 29 to platform list

sebastian-miclea:
I've added Fedora 29 to the list of platforms.

Can anyone explain why does Fedora have each version twice?


lib/packaging/platforms.rb
```diff
@@ -161,6 +161,14 @@ module Platforms
...
           repo: true,
         },
+        'f29' => {
+          architectures: ['x86_64'],
+          source_architecture: 'SRPMS',
+          package_format: 'rpm',
+          source_package_formats: ['src.rpm'],
+          signature_format: 'v4',
+          repo: true,
+        },
```
mwaggett: I don't think we need this. We used to version fedora packages with an 'f' in the version string, e.g. 'f29'. We decided that was silly and switched to numerical version, since that's what everyone else does. That transition was not as smooth as we hoped, so we included both f-prefixed and no-f versions here so that both patterns would be recognized while we transitioned. The puppet-agent 1.10.x and 5.3.x streams still use f-prefixed versions, but since we're only adding new platforms to 5.5.x and later, we don't need new f-prefixed versions anymore. Does that answer your question?

sebastian-miclea: Thank you for your detailed reply.
I've updated the PR to have the platform list without f29.

PR #919, opened 2018-12-13
(PA-2326) Addded RHEL8 to platforms.rb

sebastian-miclea:
I've added RHEL 8 to the platforms list, but I need someones confirmation that I've added it to the correct place. It's a bit confusing that there is only one entry for RHEL, (and that is 'redhatfips'). 
The current version is a common x86_64 arch, how come redhatfips is separate from the others?

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -118,6 +118,14 @@ module Pkg
           signature_format: 'v4',
           repo: true,
         },
+        '8' => {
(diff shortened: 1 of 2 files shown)


ScottGarman:
redhatfips is considered a separate platform because that package, while built on el, is compiled differently in order to comply with the FIPS requirements. For now we can ignore supporting fips for el8, as that will need to be a separate decision made by product management.

PR #924, opened 2018-12-21
(RE-11741) Rogue puppet-latest links

mwaggett:

### DIFF
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -293,7 +293,7 @@ module Pkg::Util::Net
     #   @option :excludes [Array] Strings you want to exclude from your search,
     #     e.g. 'agent' if only searching for 'puppet'.
     def remote_create_latest_symlink(package_name, dir, platform_ext, options = {})
-      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep #{package_name} "
+      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep -P #{package_name}-\d "
 
       # store this in a separate var to avoid side affects
       full_package_name = String.new(package_name)
--- a/lib/packaging/util/ship.rb
+++ b/lib/packaging/util/ship.rb
@@ -70,7 +70,7 @@ module Pkg::Util::Ship
     # First find the packages to be shipped. We must find them before moving
     # to our temporary staging directory
     local_packages = collect_packages(pkg_exts, options[:excludes])
-    return if local_packages.empty?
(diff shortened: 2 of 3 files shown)


mwaggett:
not totally sure how to test without actually shipping puppet..

lib/packaging/util/net.rb
```diff
@@ -293,7 +293,7 @@ def remote_set_immutable(host, files)
     #   @option :excludes [Array] Strings you want to exclude from your search,
     #     e.g. 'agent' if only searching for 'puppet'.
     def remote_create_latest_symlink(package_name, dir, platform_ext, options = {})
-      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep #{package_name} "
+      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep -P #{package_name}-\d "
```
shrug: I haven't checked, can we expect that -P is available everywhere we need it?

mwaggett: We only need it on weth, right? Since that's where things are staged? It definitely works there..

lib/packaging/util/ship.rb
```diff
@@ -149,12 +152,14 @@ def ship_swix(local_staging_directory, remote_path, opts = {})
   end
 
   def ship_msi(local_staging_directory, remote_path, opts = {})
-    ship_pkgs(["#{local_staging_directory}/**/*.msi"], Pkg::Config.msi_staging_server, remote_path, opts)
+    packages_have_shipped = ship_pkgs(["#{local_staging_directory}/**/*.msi"], Pkg::Config.msi_staging_server, remote_path, opts)
 
     create_rolling_repo_link(Pkg::Platforms.generic_platform_tag('windows'), Pkg::Config.msi_staging_server, remote_path, opts[:nonfinal])
```
shrug: Should this also be moved into the conditional?

mwaggett: I wasn't totally sure, but my thought process is this:
Let's say we're about to ship puppet7, so the rolling repo will be updated. If we put this in the conditional, then if the first thing we ship in puppet7 doesn't have msis or dmgs, those puppet directories will still point to puppet6, while the rpm and deb puppet repos will point to puppet7, which seems undesirable.

lib/packaging/util/net.rb
e-gris: Man page for [grep 3.0](https://www.gnu.org/software/grep/manual/grep.html) indicates -P as experimental. We're running 2.20 on weth. I've no objections but just a note to proceed with caution.

PR #926, opened 2019-01-23
(RE-11984) Use quotes when grepping a regex

mwaggett:
This commit escapes the '\d' and adds quotes around the <package_name>-\<version>
regex we use for finding the latest version of a project and symlinking to it.
Previously, without the escape or quotes, we ended up searching for the literal
string '-d' after the package name (e.g. 'puppet-bolt-d'), which resulted in no
new -latest symlinks when we shipped a new version.

### DIFF
--- a/lib/packaging/util/net.rb
+++ b/lib/packaging/util/net.rb
@@ -293,7 +293,7 @@ module Pkg::Util::Net
     #   @option :excludes [Array] Strings you want to exclude from your search,
     #     e.g. 'agent' if only searching for 'puppet'.
     def remote_create_latest_symlink(package_name, dir, platform_ext, options = {})
-      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep -P #{package_name}-\d "
+      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep -P '#{package_name}-\\d' "
 
       # store this in a separate var to avoid side affects
       full_package_name = String.new(package_name)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+## [0.99.22] - 2019-01-22
+### Changed
+- (RE-11984) Add quotes and escape '\d' in the regex used to find the latest
+  version of a particular project.
+
 ## [0.99.21] - 2019-01-08
 ### Changed
 - (RE-11741) Search for '<project_name>-\d' in order to prevent aggressive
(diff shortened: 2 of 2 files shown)


mwaggett:
without quotes:
```
molly.waggett@weth:/opt/downloads/mac/puppet6/10.13/x86_64$ ls | grep -P puppet-bolt-\d
```
with quotes:
```
molly.waggett@weth:/opt/downloads/mac/puppet6/10.13/x86_64$ ls | grep -P 'puppet-bolt-\d'
puppet-bolt-0.23.0-1.osx10.13.dmg
puppet-bolt-0.24.0-1.osx10.13.dmg
puppet-bolt-0.25.0-1.osx10.13.dmg
puppet-bolt-1.0.0-1.osx10.13.dmg
puppet-bolt-1.10.0-1.osx10.13.dmg
puppet-bolt-1.1.0-1.osx10.13.dmg
puppet-bolt-1.2.0-1.osx10.13.dmg
puppet-bolt-1.3.0-1.osx10.13.dmg
puppet-bolt-1.4.0-1.osx10.13.dmg
puppet-bolt-1.5.0-1.osx10.13.dmg
puppet-bolt-1.6.0-1.osx10.13.dmg
puppet-bolt-1.7.0-1.osx10.13.dmg
puppet-bolt-1.8.0-1.osx10.13.dmg
puppet-bolt-1.8.1-1.osx10.13.dmg
puppet-bolt-1.9.0-1.osx10.13.dmg
```

lib/packaging/util/net.rb
```diff
@@ -293,7 +293,7 @@ def remote_set_immutable(host, files)
     #   @option :excludes [Array] Strings you want to exclude from your search,
     #     e.g. 'agent' if only searching for 'puppet'.
     def remote_create_latest_symlink(package_name, dir, platform_ext, options = {})
-      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep -P #{package_name}-\d "
+      ls_cmd = "ls -1 *.#{platform_ext} | grep -v latest | grep -v rc | grep -P '#{package_name}-\d' "
```
donoghuc: This could be an Ubuntu thing but do you need to escape the `\d`? Below is the output of an irb session on my ubuntu laptop. For me I only get a match with `grep -P '#{package_name}-\\d'`
```
irb(main):021:0> package_name='puppet-bolt'
=> "puppet-bolt"
irb(main):023:0> Open3.capture3("echo puppet-bolt-0.23.0-1.osx10.13.dmg | grep -P '#{package_name}-\\d'")
=> ["puppet-bolt-0.23.0-1.osx10.13.dmg\n", "", #<Process::Status: pid 13734 exit 0>]
irb(main):024:0> Open3.capture3("echo puppet-bolt-0.23.0-1.osx10.13.dmg | grep -P '#{package_name}-\d'")
=> ["", "", #<Process::Status: pid 13741 exit 1>]
```

mwaggett: hmm. possibly.. I can reproduce what you're seeing both locally and on the host where this will ultimately get run..

PR #928, opened 2019-01-28
(maint) Update changelog for 0.99.23 release

mwaggett:


CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+## [0.99.23] - 2019-01-28
+### Changed
```
underscorgan: I usually only use `changed` to indicate some sort of backwards-incompatible change. Adding a check feels more like an `Added`.

mwaggett: ooh, then I've been using `Changed` wrong for a while.. I've been using it whenever there's any behavior change, as opposed to something purely additive.

underscorgan: Yeah, the way I tend to think about it is do users need to know/care that this changed?

I don't know that my interpretation is correct, but as a consumer when I see something called out as 'changed' in the changelog I tend to look a little more carefully to make sure my stuff isn't going to break 🤷‍♀️

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
...
+- Check if a gem has already been shipped to rubygems.org before attempting to
```
sarameisburger: nitpick: might be helpful to note it looks for a specific gem + version

CHANGELOG.md
mwaggett: @shrug @sarameisburger thoughts on `Changed` vs. `Added`?

sarameisburger: I'm leaning towards `Added`

PR #933, opened 2019-03-14
(RE-11105) artifactory promotion

underscorgan:
Changes needed to promote enterprise builds into artifactory

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -206,13 +206,15 @@ module Pkg
       raise "Attempt to upload '#{package}' to #{File.join(@artifactory_uri, data[:full_artifactory_path])} failed"
     end
 
-    # @param directory [String] optional, The directory where the yaml file will
-    #   be downloaded
-    # @return [String] The path to the downloaded file
-    def retrieve_yaml_data_file(directory = nil)
-      directory ||= Dir.mktmpdir
-      retrieve_package(DEFAULT_REPO_TYPE, "#{@project_version}.yaml", directory)
-      File.join(directory, "#{@project_version}.yaml")
+    # @param pkg The package to download YAML for
+    # @param ref The git ref (sha or tag) we want the YAML for
+    #
+    # @return [String] The contents of the YAML file
+    def retrieve_yaml_data(pkg, ref)
+      yaml_url = "#{@artifactory_uri}/#{DEFAULT_REPO_TYPE}/#{DEFAULT_REPO_BASE}/#{pkg}/#{ref}/#{ref}.yaml"
+      open(yaml_url) { |f| f.read }
+    rescue
+      raise "Failed to load YAML data for #{pkg} at #{ref} from #{yaml_url}!"
     end
 
     # @param platform_data [Hash] The has of the platform data that needs to be
@@ -268,11 +270,7 @@ module Pkg
     # @param platform_tags [Array[String], String] optional, either a string, or
     #   an array of strings. These are the platform or platforms that we will
     #   download packages for.
-    # @param package [String] optional, the name of the package to be
-    #   retrieved. If the user does not know this information, we can derive it
-    #   from the yaml data. This ignores everything but the package name. Any
-    #   customization for where the user wants to fetch the package is via the
-    #   download_path parameter.
+    # @param package [String] the name of the package to be retrieved
     # @param download_path [String] Optional, an optional path set to where
     #   the user wants the retrieved package to end up. If no path is specified
     #   this defaults to the pkg directory.
@@ -280,27 +278,16 @@ module Pkg
 
       if platform_tags.nil? && !package.nil?
         platform_tags = Pkg::Paths.tag_from_artifact_path(package) || DEFAULT_REPO_TYPE
-      elsif platform_tags.nil? && package.nil?
-        yaml_file = retrieve_yaml_data_file(download_path)
-        yaml_data = Pkg::Config.config_from_yaml(yaml_file)
-        platform_data = yaml_data[:platform_data]
-        platform_tags = platform_data.keys
       end
 
       Array(platform_tags).each do |platform_tag|
         puts "fetching package for #{platform_tag}"
         data = platform_specific_data(platform_tag)
-        if package.nil?
-          package_for_tag = package_name(platform_data, platform_tag)
-          puts "package name is #{package_for_tag}"
-        else
-          package_for_tag = package
-        end
         download_path_for_tag = download_path || data[:repo_subdirectories].sub(@repo_base, 'pkg')
 
         check_authorization
         artifact = Artifactory::Resource::Artifact.new(
-          download_uri: File.join(@artifactory_uri, data[:full_artifactory_path], File.basename(package_for_tag))
+          download_uri: File.join(@artifactory_uri, data[:full_artifactory_path], File.basename(package))
         )
         artifact.download(download_path_for_tag)
       end
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -273,8 +273,16 @@ module Pkg
           puts "promoting #{artifact_name} to #{path}"
           artifact_to_promote[0].copy(path)
         end
-      rescue
-        puts "Skipping promotion of #{artifact_name}; it has already been promoted"
+      rescue Artifactory::Error::HTTPError => e
+        if e.message =~ /destination and source are the same/i
+          puts "Skipping promotion of #{artifact_name}; it has already been promoted"
+        else
+          puts "#{e.level}: #{e.message}"
+          raise e
+        end
+      rescue => e
+        puts "Something went wrong promoting #{artifact_name}!"
+        raise e
       end
     end
 
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -256,7 +256,7 @@ module Pkg
       artifact_to_promote = Artifactory::Resource::Artifact.search(name: artifact_name, :artifactory_uri => @artifactory_uri)
 
       if artifact_to_promote.empty?
-        puts "Error: could not find PKG=#{pkg} at REF=#{git_ref} for #{platform_tag}"
+        raise "Error: could not find PKG=#{pkg} at REF=#{git_ref} for #{platform_tag}"
       end
 
       # This makes an assumption that we're using some consistent repo names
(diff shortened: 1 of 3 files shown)


lib/packaging/artifactory.rb
```diff
@@ -230,42 +235,69 @@ def package_name(platform_data, platform_tag)
...
+      if File.extname(artifact_name) == '.rpm'
+        promotion_paths = Array(repositories).compact.map { |repo| "rpm_#{repo}/#{platform_tag}/#{artifact_name}" }
+      else
+        promotion_paths = Array(repositories).compact.map { |repo| "debian_#{repo}/#{platform_tag}/#{artifact_name}" }
+      end
+
+      begin
+        promotion_paths.each do |path|
+          puts "promoting #{artifact_name} to #{path}"
+          artifact_to_promote[0].copy(path)
```
e-gris: Should this be ```artifact_to_promote[0].move(path)``` ?

lib/packaging/artifactory.rb
```diff
@@ -230,42 +235,69 @@ def package_name(platform_data, platform_tag)
...
+    # 'promote' by copying the package(s) to the enterprise directory on artifactory
+    #
+    # @param pkg [String] the package name ex. puppet-agent
+    # @param ref [String] tag or SHA of package(s) to be promoted
+    # @param platform_tag [String] the platform tag of the artifact
+    #   ex. el-7-x86_64, ubuntu-18.04-amd64
+    # @param repositories [Array(String)] the repositories to promote
+    #   the artifact to. Will prepend 'rpm_' or 'debian_' to the repositories
+    #   depending on package type
+    def promote_package(pkg, ref, platform_tag, repositories)
```
e-gris: Here, using ```pkg``` and ```ref``` but ```retrieve_yaml_data``` using ```package``` and ```git_ref``` seems inconsistent.

lib/packaging/artifactory.rb
```diff
@@ -230,42 +235,69 @@ def package_name(platform_data, platform_tag)
...
+      # load package metadata
+      yaml_content = retrieve_yaml_data(pkg, ref)
+      yaml_data = YAML::load(yaml_content)
+
+      # get the artifact name
+      artifact_name = package_name(yaml_data[:platform_data], platform_tag)
+      artifact_to_promote = Artifactory::Resource::Artifact.search(name: artifact_name, :artifactory_uri => @artifactory_uri)
+
+      if artifact_to_promote.empty?
+        raise "Error: could not find PKG=#{pkg} at REF=#{git_ref} for #{platform_tag}"
```
e-gris: I like using the "Error: " tag a lot in failure messages. Suggest adding it to the failure message at line 271 as well.

lib/packaging/artifactory.rb
```diff
@@ -203,13 +206,15 @@ def deploy_package(package)
       raise "Attempt to upload '#{package}' to #{File.join(@artifactory_uri, data[:full_artifactory_path])} failed"
```
e-gris: __Minor (and I understand not part of this PR)_ : ```File.join``` on URIs, if my vague recollection is not failing me, can result in undesirable effects.

Feel free to skip it but it just caught my eye.

lib/packaging/artifactory.rb
mwaggett: @e-gris I think we want to copy, not move, so we can still find artifacts in the non-promoted place.

lib/packaging/artifactory.rb
```diff
@@ -203,13 +206,15 @@ def deploy_package(package)
...
-      directory ||= Dir.mktmpdir
-      retrieve_package(DEFAULT_REPO_TYPE, "#{@project_version}.yaml", directory)
-      File.join(directory, "#{@project_version}.yaml")
+    # @param package The package to download YAML for
+    # @param git_ref The git ref (sha or tag) we want the YAML for
+    #
+    # @return [String] The contents of the YAML file
+    def retrieve_yaml_data(package, git_ref)
+      yaml_url = "#{@artifactory_uri}/#{DEFAULT_REPO_TYPE}/#{DEFAULT_REPO_BASE}/#{package}/#{git_ref}/#{git_ref}.yaml"
+      open(yaml_url){|f| f.read}
```
mwaggett: rubocop is mad about the formatting around the `{}`

lib/packaging/artifactory.rb
underscorgan: Yep, we want to copy.

lib/packaging/artifactory.rb
```diff
@@ -230,42 +235,69 @@ def package_name(platform_data, platform_tag)
...
+
     # @param platform_tags [Array[String], String] optional, either a string, or
     #   an array of strings. These are the platform or platforms that we will
     #   download packages for.
-    # @param package [String] optional, the name of the package to be
-    #   retrieved. If the user does not know this information, we can derive it
-    #   from the yaml data. This ignores everything but the package name. Any
-    #   customization for where the user wants to fetch the package is via the
-    #   download_path parameter.
+    # @param package [String] the name of the package to be retrieved
```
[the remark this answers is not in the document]
underscorgan: Hmm, I might actually just remove this method. It isn't in use anywhere and I'm not convinced that it is (or was previously) in working condition.  I suspect we may need to rewrite it when we get to the point of actually using the functionality.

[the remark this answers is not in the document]
mwaggett: I'm super down to remove and rewrite if it's not in use.

lib/packaging/artifactory.rb
underscorgan: I wasn't using the 'Error' tag on the other message because it's not actually raising an error, it's more 'info' level.

underscorgan:
updated!

lib/packaging/artifactory.rb
```diff
@@ -203,13 +206,15 @@ def deploy_package(package)
...
 
-    # @param directory [String] optional, The directory where the yaml file will
-    #   be downloaded
-    # @return [String] The path to the downloaded file
-    def retrieve_yaml_data_file(directory = nil)
-      directory ||= Dir.mktmpdir
-      retrieve_package(DEFAULT_REPO_TYPE, "#{@project_version}.yaml", directory)
-      File.join(directory, "#{@project_version}.yaml")
+    # @param pkg The package to download YAML for
+    # @param ref The git ref (sha or tag) we want the YAML for
```
mwaggett: We should probably note that these params are strings, to be consistent with the rest of this file.
Also, perhaps include an example for `pkg` - when I see `pkg` or `package`, I'm never sure if that means the project name, like 'puppet-agent', or the package file name.

lib/packaging/artifactory.rb
```diff
@@ -230,47 +235,46 @@ def package_name(platform_data, platform_tag)
...
+    # @param ref [String] tag or SHA of package(s) to be promoted
+    # @param platform_tag [String] the platform tag of the artifact
+    #   ex. el-7-x86_64, ubuntu-18.04-amd64
+    # @param repositories [Array(String)] the repositories to promote
+    #   the artifact to. Will prepend 'rpm_' or 'debian_' to the repositories
+    #   depending on package type
+    def promote_package(pkg, ref, platform_tag, repositories)
+      # load package metadata
+      yaml_content = retrieve_yaml_data(pkg, ref)
+      yaml_data = YAML::load(yaml_content)
```
mwaggett: Any reason for this not to just live in the `retrieve_yaml_data` method? Is there any scenario where we'd just want the unparsed file content?

lib/packaging/artifactory.rb
```diff
@@ -230,47 +235,46 @@ def package_name(platform_data, platform_tag)
...
-          puts "package name is #{package_for_tag}"
-        else
-          package_for_tag = package
-        end
-        download_path_for_tag = download_path || data[:repo_subdirectories].sub(@repo_base, 'pkg')
+      # This makes an assumption that we're using some consistent repo names
+      # but need to either prepend 'rpm_' or 'debian_' based on package type
+      if File.extname(artifact_name) == '.rpm'
+        promotion_paths = Array(repositories).compact.map { |repo| "rpm_#{repo}/#{platform_tag}/#{artifact_name}" }
+      else
```
mwaggett: Should we maybe check for `File.extname(artifact_name) == '.deb'` and then fail if it's something other than '.rpm' or '.deb'?

lib/packaging/artifactory.rb
```diff
@@ -230,47 +235,46 @@ def package_name(platform_data, platform_tag)
...
-          download_uri: File.join(@artifactory_uri, data[:full_artifactory_path], File.basename(package_for_tag))
-        )
-        artifact.download(download_path_for_tag)
+      begin
+        promotion_paths.each do |path|
+          puts "promoting #{artifact_name} to #{path}"
+          artifact_to_promote[0].copy(path)
+        end
+      rescue
+        puts "Skipping promotion of #{artifact_name}; it has already been promoted"
```
mwaggett: Is this the only case where promotion could fail? Or do we want a more generic message?

lib/packaging/artifactory.rb
underscorgan: I was thinking we may not need to parse it if we're trying to download the file maybe? I'm not sure though, since this is the only place we're using that method 🤷‍♀️

mwaggett: I'm not sure I know what you mean by 'download the file'. Like, download the yaml file? I'm still not seeing how an unparsed string of yaml would ever be useful.

underscorgan: I was thinking if we just needed to write the output to a file we wouldn't need to parse it, like

```
file = File.open('tmp.yml')
file.puts yaml_content
file.close
```

mwaggett: ahh, I see. well I'll leave it up to you whether to keep this here or move it. I can't imagine when we'd want to rewrite the yaml, but I'm fine leaving it.

lib/packaging/artifactory.rb
```diff
@@ -230,47 +236,56 @@ def package_name(platform_data, platform_tag)
...
-    def retrieve_package(platform_tags = nil, package = nil, download_path = nil)
-
-      if platform_tags.nil? && !package.nil?
-        platform_tags = Pkg::Paths.tag_from_artifact_path(package) || DEFAULT_REPO_TYPE
-      elsif platform_tags.nil? && package.nil?
-        yaml_file = retrieve_yaml_data_file(download_path)
-        yaml_data = Pkg::Config.config_from_yaml(yaml_file)
-        platform_data = yaml_data[:platform_data]
-        platform_tags = platform_data.keys
+    # Promotes a build based on build SHA or tag
```
sarameisburger: Would it be useful to specifically mention SNAPSHOTS here as well?

underscorgan: updated!

PR #935, opened 2019-04-09
(RE-12205) Add retry + other time servers to msi signing

denis-susan:
This pr solves latency issues of timeservers by:
* adding multiple time servers
* adding a retry mechanism of 5 tries per server.

### DIFF
--- a/lib/packaging/sign/msi.rb
+++ b/lib/packaging/sign/msi.rb
@@ -66,23 +66,42 @@ module Pkg::Sign::Msi
       if "/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" verify -in "C:/#{work_dir}/$msi" ; then
         echo "$msi is already signed, skipping . . ." ;
       else
-        "/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" sign \
-          -n "Puppet" -i "http://www.puppet.com" \
-          -h sha1 \
-          -pkcs12 "#{Pkg::Config.msi_signing_cert}" \
-          -pass "#{Pkg::Config.msi_signing_cert_pw}" \
-          -t "http://timestamp.verisign.com/scripts/timstamp.dll" \
-          -in "C:/#{work_dir}/$msi" \
-          -out "C:/#{work_dir}/signed-$msi"
-        "/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" sign \
-          -n "Puppet" -i "http://www.puppet.com" \
-          -nest -h sha256 \
-          -pkcs12 "#{Pkg::Config.msi_signing_cert}" \
-          -pass "#{Pkg::Config.msi_signing_cert_pw}" \
-          -ts "http://sha256timestamp.ws.symantec.com/sha256/timestamp" \
-          -in "C:/#{work_dir}/signed-$msi" \
-          -out "C:/#{work_dir}/$msi"
-        rm "C:/#{work_dir}/signed-$msi" ;
+          tries=5
+          sha1Servers=(http://timestamp.verisign.com/scripts/timstamp.dll
+            http://timestamp.globalsign.com/scripts/timstamp.dll
+            http://www.startssl.com/timestamp)
+              for timeserver in "${sha1Servers[@]}"; do
+                for ((try=1; try<=$tries; try++)) do
+                  ret=$(/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe sign \
+                            -n "Puppet" -i "http://www.puppet.com" \
+                            -h sha1 \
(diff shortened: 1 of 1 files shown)


melissa:
@denis-susan thanks for submitting this! Any reason you're only adding the retries to the sha1 signature and not the sha256 signature?

lib/packaging/sign/msi.rb
```diff
@@ -66,14 +66,24 @@ def sign(target_dir = 'pkg')
...
+                    -h sha1 \
+                    -pkcs12 "#{Pkg::Config.msi_signing_cert}" \
+                    -pass "#{Pkg::Config.msi_signing_cert_pw}" \
+                    -t "$timeserver" \
+                    -in "C:/#{work_dir}/$msi" \
+                    -out "C:/#{work_dir}/signed-$msi)
+          if [ $ret == "Succeeded" ]; then break; fi
+          done;
+          if [ $ret == "Succeeded" ]; then break; fi
+          done;
```
mwaggett: I assume this duplication was a copy+paste mistake?

mwaggett:
seconding @melissa: since timeouts appear to happen for both signature types, I think we should probably add retries to both, unless you know of a reason not to.

mwaggett:
also @denis-susan could you update the commit message to include the relevant ticket title? https://tickets.puppetlabs.com/browse/RE-12205

lib/packaging/sign/msi.rb
```diff
@@ -66,14 +66,24 @@ def sign(target_dir = 'pkg')
...
-          -t "http://timestamp.verisign.com/scripts/timstamp.dll" \
-          -in "C:/#{work_dir}/$msi" \
-          -out "C:/#{work_dir}/signed-$msi"
+        timeservers=(http://timestamp.verisign.com/scripts/timstamp.dll
+          http://timestamp.globalsign.com/scripts/timstamp.dll
+          http://www.startssl.com/timestamp)
+          tries=5
+          for timeserver in "${timeservers[@]}"; do
+          for ((try=1; try<=$tries; try++)) do
+          ret=$(/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" sign \
```
mwaggett: I think you've got an errant `"` here

mwaggett:
Has this been tested? I created https://confluence.puppetlabs.com/display/SRE/How+to+test+%28certain%29+changes+to+packaging to explain how to test changes to packaging. 
For this, you'll probably want to build a new windows package (`bundle exec build puppet-agent windows2012r2-x64`) and then run `bundle exec sign` to test these changes. 
I *think* you should be able to run that without a problem since you're in the 'release' group and windows signing doesn't require a passphrase (this is why you need to build new packages - if you retrieved any rpms or debs, you'd be prompted for a passphrase when trying to sign those), but if you run into trouble, we can test it out for you.

lib/packaging/sign/msi.rb
denis-susan: No, this is to break the other loop (there's 2 loops, one to retry the current server and another to try different servers, that loop just breaks out of the loops on success

denis-susan:
> @denis-susan thanks for submitting this! Any reason you're only adding the retries to the sha1 signature and not the sha256 signature?

currently only the sha1 timestamp server has issues. If we need to add the same mechanism for sha256 it should be trivial.

denis-susan:
> Has this been tested? I created https://confluence.puppetlabs.com/display/SRE/How+to+test+%28certain%29+changes+to+packaging to explain how to test changes to packaging.
> For this, you'll probably want to build a new windows package (`bundle exec build puppet-agent windows2012r2-x64`) and then run `bundle exec sign` to test these changes.
> I _think_ you should be able to run that without a problem since you're in the 'release' group and windows signing doesn't require a passphrase (this is why you need to build new packages - if you retrieved any rpms or debs, you'd be prompted for a passphrase when trying to sign those), but if you run into trouble, we can test it out for you.

I cant test, it asks me for a password to mozart

lib/packaging/sign/msi.rb
mwaggett: ohhhh, I see. the lack of indentation threw me off.

mwaggett:
gahh, @denis-susan I forgot that since this is just signing, you can test with the vanagon generic builder, since we sign at build-time now.
Looks like syntax is off somewhere: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/337/

denis-susan:
Passing build: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/vanagon-generic-builder/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/339/

PR #940, opened 2019-04-11
(maint) Fix (soft) failure in artifactory promotions

underscorgan:


e-gris:
Assuming you don't want to squash but don't want to merge incorrectly. I'll let you do whichever merge seems right.

PR #943, opened 2019-04-15
(maint) Add ability to promote debian packages into a component

underscorgan:
We need this ability to be able to use the enterprise repos we're
promoting into.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -266,10 +266,12 @@ module Pkg
     # @param ref [String] tag or SHA of package(s) to be promoted
     # @param platform_tag [String] the platform tag of the artifact
     #   ex. el-7-x86_64, ubuntu-18.04-amd64
-    # @param repositories [Array(String)] the repositories to promote
+    # @param repository [String] the repository to promote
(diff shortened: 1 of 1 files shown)


mwaggett:
Can you provide a bit more context in the commit message? I'm not sure what you mean by promoting debian packages "into a component"..

underscorgan:
Components are parts of debian repositories that let you break up repositories into separate parts that can be individually enabled or disabled. With the way we have our enterprise debian repos configured in artifactory, the component includes the PE version to be pulling packages from, and also whether or not to be pulling from release or feature branch. See https://github.com/puppetlabs/enterprise-dist/pull/3477 for example usage. I didn't include more information in the commit message since components are just part of debian repositories.

mwaggett:
¯\_(ツ)_/¯  the commit message could include that we're setting the component to the PE version to ensure we're getting the appropriate packages

underscorgan:
That's in the enterprise-dist commit comment, since this is just adding the generic ability to specify a component when promoting debian packages and the enterprise-dist commit has the changes that are tied specifically to our repository setup.

PR #950, opened 2019-05-20
(CPR-698) Update apt archive automation

mwaggett:


mwaggett:
tested the `stage_apt_archives` method by running `bundle exec rake pl:remote:stage_archives\['cisco-wrlinux','main','mac/10.10'\]` from the katamari repo with `STAGING_SERVER=weth-dev.delivery.puppetlabs.net`.
I *think* it did what I want, but it was hard to test because there isn't a good way to put things back after moving them and I already did a bunch of testing on weth-dev during the first iteration of this work.

lib/packaging/archive.rb
```diff
@@ -8,7 +8,7 @@ def base_paths
 
   # Array of paths for temporarily staging artifacts before syncing to release-archives on s3
   def archive_paths
-    [Pkg::Config.yum_archive_path, Pkg::Config.apt_archive_path, Pkg::Config.freight_archive_path, Pkg::Config.downloads_archive_path].compact.freeze
+    [Pkg::Config.yum_archive_path, Pkg::Config.apt_archive_path, Pkg::Config.freight_archive_path, Pkg::Config.downloads_archive_path, '/opt/tmp-apt'].compact.freeze
```
e-gris: '/opt/tmp-apt' is reused several times. Is there any convenient place to define it as a constant?

mwaggett: yeeeeah, I thought about adding something to `lib/packaging/config/params.rb`, like I did for `apt_archive_repo_command` and the other archive paths, but I decided it wasn't worth it since this path is literally just for moving packages out of the way and then deleting them. like, we won't even sync from this path, since we'll be regenerating these files with the `update_archive_apt_repo`. I _could_ add it as a constant within this module, if you think it's worth it.

e-gris: If you considered it and dismissed it, that works for me.

e-gris:
Leaving un-merged in case you want to squash.

PR #954, opened 2019-05-30
(maint) Don't delete packages from rsync servers

mwaggett:
This commit removes the `--delete` flag from the `deploy_to_rsync_server`
process. As we plan to archive EOL packages, we want people using our rsync
servers to be able continue to do so with minimal effort on our part, so,
rather than maintaining a separate rsync server for archives, we'll just keep
packages available on the existing ones FOREVER.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -240,7 +240,8 @@ namespace :pl do
         Pkg::Util::Execution.retry_on_fail(:times => 3) do
           Pkg::Config.rsync_servers.each do |rsync_server|
             ['apt', 'yum'].each do |repo|
-              command = "sudo su - rsync --command 'rsync --verbose -a --exclude '*.html' --delete /opt/repo-s3-stage/repositories/#{repo}.puppetlabs.com/ rsync@#{rsync_server}:/opt/repository/#{repo}'"
+              # Don't --delete so that folks using archived packages can continue to do so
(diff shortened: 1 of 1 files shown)


tasks/ship.rake
```diff
@@ -240,7 +240,7 @@ namespace :pl do
         Pkg::Util::Execution.retry_on_fail(:times => 3) do
           Pkg::Config.rsync_servers.each do |rsync_server|
             ['apt', 'yum'].each do |repo|
-              command = "sudo su - rsync --command 'rsync --verbose -a --exclude '*.html' --delete /opt/repo-s3-stage/repositories/#{repo}.puppetlabs.com/ rsync@#{rsync_server}:/opt/repository/#{repo}'"
+              command = "sudo su - rsync --command 'rsync --verbose -a --exclude '*.html' /opt/repo-s3-stage/repositories/#{repo}.puppetlabs.com/ rsync@#{rsync_server}:/opt/repository/#{repo}'"
```
underscorgan: might be helpful to comment about why we aren't `--delete`ing here just for the sake of future us?

mwaggett: ahh yeah, I figured the commit message could provide that info, but I can add a comment.

mwaggett: @underscorgan added!

e-gris:
@mwaggett I'll hold off the merge in case you want to squash or not.

PR #956, opened 2019-06-05
 (maint) Update regex for error when repromoting a package

underscorgan:
We recently found anonymous deletions were enabled on the repo we were
targeting. When we removed that permission, the error message for
promoting a package that already exists changed. As a result of this, we
also found we were using an invalid field in our error messages.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -305,10 +305,10 @@ module Pkg
             promoted_artifact.properties(properties)
           end
         rescue Artifactory::Error::HTTPError => e
-          if e.message =~ /destination and source are the same/i
+          if e.message =~ /(destination and source are the same|user doesn't have permissions to override)/i
             puts "Skipping promotion of #{artifact_name}; it has already been promoted"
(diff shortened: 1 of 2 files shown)


lib/packaging/artifactory.rb
```diff
@@ -305,10 +305,10 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
             promoted_artifact.properties(properties)
           end
         rescue Artifactory::Error::HTTPError => e
-          if e.message =~ /destination and source are the same/i
+          if e.message =~ /(destination and source are the same|user doesn't have permissions to override)/i
```
mwaggett: Is it possible to get the permissions error in other cases? I want to make sure we don't obscure a real permissions issue.

underscorgan: I think this error is just when you're trying to write a file that already exists and you don't have delete permissions.

mwaggett: ahh okay, 'override', that makes sense

PR #957, opened 2019-06-07
(maint) Don't use `set -e` for windows signing

underscorgan:
We need to be able to loop through timestamp servers, and if the first
signing fails `set -e` will cause the code to exit. Instead, after each
signature if it was unable to sign successfully, explicitly `exit 1`.

### DIFF
--- a/lib/packaging/sign/msi.rb
+++ b/lib/packaging/sign/msi.rb
@@ -62,50 +62,54 @@ module Pkg::Sign::Msi
     #
     # Once we no longer support Windows 8/Windows Vista, we can remove the
     # first Sha1 signature.
-    Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, %Q(for msi in #{msis.map { |d| File.basename(d) }.join(" ")}; do
-      if "/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" verify -in "C:/#{work_dir}/$msi" ; then
-        echo "$msi is already signed, skipping . . ." ;
-      else
-          tries=5
-          sha1Servers=(http://timestamp.verisign.com/scripts/timstamp.dll
-            http://timestamp.globalsign.com/scripts/timstamp.dll
-            http://www.startssl.com/timestamp)
-              for timeserver in "${sha1Servers[@]}"; do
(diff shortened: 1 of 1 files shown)


underscorgan:
I'm running some testing locally and want to make sure that works before this gets merged in

underscorgan:
also definitely recommend reviewing with whitespace changes hidden, it was complicated looping with bad whitespacing and I lined it up to make sure I was putting changes in the right place

underscorgan:
Using this packaging I was able to confirm that I can sign windows packages
Screen Shot of msi digital signatures

I also tested this with an explicit `exit 1` in the ssh command to make sure it would in fact fail with a signature failure
```
commit 061882df87e4b930fb3ca46187e0c4ae1ae1333f (HEAD -> test, origin/test)
Author: Morgan Rhodes <morgan@puppet.com>
Date:   Fri Jun 7 15:04:43 2019 -0700

    (maint) force a failure

diff --git a/lib/packaging/sign/msi.rb b/lib/packaging/sign/msi.rb
index 2be0bc4..22b9221 100644
--- a/lib/packaging/sign/msi.rb
+++ b/lib/packaging/sign/msi.rb
@@ -63,6 +63,7 @@ module Pkg::Sign::Msi
     # Once we no longer support Windows 8/Windows Vista, we can remove the
     # first Sha1 signature.
     Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, %Q(for msi in #{msis.map { |d| File.basename(d) }.join(" ")}; do
+      exit 1
       if "/cygdrive/c/tools/osslsigncode-fork/osslsigncode.exe" verify -in "C:/#{work_dir}/$msi" ; then
         echo "$msi is already signed, skipping . . ." ;
       else
```

and as expected the sign command failed.

underscorgan:
Difference between 2 commits without whitespace: https://github.com/underscorgan/packaging/compare/6aa7f2efa47d1c2adf0d3ece2eedd4da0898e008...64178976674e294883dec3f17ef0a441b82cb416?w=1

I tested this signing unsigned packages, signed packages, and forced an `exit 1` to make sure this all still works.

PR #958, opened 2019-06-10
(maint) Make the Pkg::Deb::Repo #sign_repos spec allow for the 'keychain' command.

e-gris:
In an attempt into understanding the SIGTERM issues we've been
occasionally spotting in packaging, I noticed that rspec didn't run on
my machine. Turns out I had 'keychain' installed and the spec tests
didn't allow for that. rspec was failing on my box.

Update the spec test to allow for this possibility.
Also sneak in a minor bit of code refactoring in Pkg::Deb::Repo.sign_repos

### DIFF
--- a/spec/lib/packaging/deb/repo_spec.rb
+++ b/spec/lib/packaging/deb/repo_spec.rb
@@ -128,12 +128,21 @@ describe "Pkg::Deb::Repo" do
     end
 
     it "makes a repo for each dist" do
+      # stub out start_keychain to prevent actual keychain starts.
+      Pkg::Util::Gpg.stub(:start_keychain)
+
       dists = cows.reject { |cow| cow.empty? }
       distfiles = {}
-      reprepro = "/bin/reprepro"
       Pkg::Util::File.should_receive(:directories).with("repos/apt").and_return(dists)
-      Pkg::Util::Tool.should_receive(:check_tool).with("reprepro").and_return(reprepro)
-      Pkg::Util::Execution.should_receive(:capture3).at_least(1).times.with("#{reprepro} -vvv --confdir ./conf --dbdir ./db --basedir ./ export")
+
+      # Let the keychain command happen if find_tool('keychain') finds it
+      keychain_command = "/usr/bin/keychain -k mine"
+      allow(Pkg::Util::Execution).to receive(:capture3) { keychain_command }
+
+      # Enforce reprepro
+      reprepro_command = "/bin/reprepro"
+      Pkg::Util::Tool.should_receive(:check_tool).with("reprepro").and_return(reprepro_command)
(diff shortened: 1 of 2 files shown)


mwaggett:
Looks like the `Pkg::Deb::Repo.sign_repos` method is only used here: https://github.com/puppetlabs/packaging/blob/1.0.x/tasks/nightly_repos.rake#L40
which is only used here: https://github.com/puppetlabs/packaging/blob/1.0.x/tasks/nightly_repos.rake#L27
which is only used here: https://github.com/puppetlabs/packaging/blob/1.0.x/tasks/nightly_repos.rake#L263
which is used in ci-job-configs, e.g. https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/puppet-agent-internal-promotion.sh#L111

I think if you run https://github.com/puppetlabs/ci-job-configs/blob/master/resources/scripts/puppet-agent-internal-promotion.sh#L109-L111 from the puppet-agent repo at an already-build version that should be good?

e-gris:
I tested the minor refactor. I didn't break anything.

spec/lib/packaging/deb/repo_spec.rb
```diff
@@ -128,12 +128,21 @@
...
       dists = cows.reject { |cow| cow.empty? }
       distfiles = {}
-      reprepro = "/bin/reprepro"
       Pkg::Util::File.should_receive(:directories).with("repos/apt").and_return(dists)
-      Pkg::Util::Tool.should_receive(:check_tool).with("reprepro").and_return(reprepro)
-      Pkg::Util::Execution.should_receive(:capture3).at_least(1).times.with("#{reprepro} -vvv --confdir ./conf --dbdir ./db --basedir ./ export")
+
+      # Let the keychain command happen if find_tool('keychain') finds it
+      keychain_command = "/usr/bin/keychain -k mine"
+      allow(Pkg::Util::Execution).to receive(:capture3) { keychain_command }
```
mwaggett: It looks like keychain is only loaded if `Pkg::Util::Tool.find_tool('keychain')` is true. I'm wondering if it would make more sense to just `allow(Pkg::Util::Tool).to receive(:find_tool).with('keychain').and_return(false)` (or whatever the exact syntax is) and move on from there?

e-gris: Seems to me minimally testing the case of keychain existing on the current machine is richer than coercing the environment into insisting that keychain isn't there.

mwaggett: In that case, the keychain case should be a separate test case. Right now, you're just allowing the current test to pass through if keychain exists.

e-gris: If keychain is there, I'm testing that it has the correct form. If it isn't there, we carry on.
This mimics the design of the actual method.

The method itself could be refactored to handle keychain and reprepro seperately but that looks like unnecessary work.

PR #961, opened 2019-06-28
(RE-12520) Don't try to ship gems with unset params

mwaggett:
tldr: allow skipping the `ship_gem_to_downloads` task if `gem_host` is unset so that projects can keep using `uber_ship_lite` (which includes `ship_gem`, which includes `ship_gem_to_downloads`), but fail if `gem_path` is unset so that we don't accidentally ship to a weird path.

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -324,6 +324,7 @@ namespace :pl do
   task :ship_gem_to_downloads => 'pl:fetch' do
     unless Pkg::Config.gem_host
       warn 'Value `Pkg::Config.gem_host` not defined; skipping shipping to public Download server'
+      exit
     end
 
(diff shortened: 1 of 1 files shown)


mwaggett:
Tested by:
1. checkout bolt-vanagon 1.25.0
2. `bundle install`
3. `bundle exec rake pl:jenkins:retrieve`
4. `bundle exec rake pl:ship_gem` - confirmed that without this change, the task fails
5. set `PACKAGING_LOCATION` to this branch
6. `bundle update`
7. `bundle exec rake pl:ship_gem` - confirmed that the task exits cleanly with a warning after this change
8. set `GEM_HOST` to weth
9. `bundle exec rake pl:ship_gem` - confirmed that the task now fails with `gem_path` unset

PR #962, opened 2019-07-09
(RE-12520) 0.99.36 release preparation.

e-gris:


sarameisburger:
You'll wanna update https://github.com/puppetlabs/packaging/pull/962/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR428 as well

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+## [0.99.36] - 2019-07-09
+### Updated
```
mwaggett: (nitpick, also not specific to this PR, BUT) do we want to keep `Updated` as a heading? Looks like the 0.99.34 release is the first time we used it. In the past, we've just used 'Added', 'Changed', 'Fixed', and 'Removed'. ¯\_(ツ)_/¯

e-gris:
D'oh!  I miss that _every_ damn time.

CHANGELOG.md
e-gris: Let's fix that now. Don't need 'Updated' with 'Changed' already established.

PR #964, opened 2019-08-09
(RE-12499) Add artifactory tasks to download and ship PE packages

sarameisburger:
`download_packages` downloads specified packages based on md5sum
`ship_pe_tarballs` grabs tarballs from specified directory and ships them to specified artifactory repo + paths

Relates to: https://github.com/puppetlabs/enterprise-dist/pull/3536

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -1,5 +1,6 @@
 require 'uri'
 require 'open-uri'
+require 'digest'
 
 module Pkg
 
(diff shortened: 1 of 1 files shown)


lib/packaging/artifactory.rb
```diff
@@ -318,6 +318,47 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
       end
     end
 
+    # Using the manifest provided by enterprise-dist, grab the appropropriate packages from artifactory based on md5sum
+    # staging_directory String location to download packages to
+    # manifest File containing information about what packages to download and the corresponding md5sums
```
mwaggett: Can these follow the `@param <name> [<Type>] <description>` format that's already used in this file?

lib/packaging/artifactory.rb
```diff
@@ -318,6 +318,47 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
...
+    def download_packages(staging_directory, manifest)
+      check_authorization
+      require 'digest'
```
mwaggett: I feel like things should be required at the top of the file, though I'm not totally sure of the pros/cons..

lib/packaging/artifactory.rb
```diff
@@ -318,6 +318,47 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
...
+            artifact_to_download.download("#{staging_directory}/#{dist}", filename: "#{info["filename"]}")
+          end
+        end
+      end
+    end
+
+    # Ship PE tarballs to specified artifactory repo and paths
+    # tarball_path String the path of the tarballs to ship
+    # target_repo String the artifactory repo to ship the tarballs
+    # ship_paths [String] the artifactory path(s) to ship the tarballs to within the target_repo
```
mwaggett: same re: formatting

lib/packaging/artifactory.rb
```diff
@@ -318,6 +318,47 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
...
+            puts "downloading #{artifact_to_download.download_uri}"
+            artifact_to_download.download("#{staging_directory}/#{dist}", filename: "#{info["filename"]}")
+          end
+        end
+      end
+    end
+
+    # Ship PE tarballs to specified artifactory repo and paths
+    # tarball_path String the path of the tarballs to ship
+    # target_repo String the artifactory repo to ship the tarballs
```
mwaggett:
```suggestion
    # target_repo String the artifactory repo to ship the tarballs to
```

sarameisburger:
@mwaggett updated based on your comments. There aren't any meaningful tests I can think of for these methods that don't require an artifactory connection

lib/packaging/artifactory.rb
```diff
@@ -318,6 +319,46 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
       end
     end
 
+    # Using the manifest provided by enterprise-dist, grab the appropropriate packages from artifactory based on md5sum
+    # @param staging_directory [String] location to download packages to
+    # @param manifest [File] manifest file containing information about what packages to download and the corresponding md5sums
```
mwaggett: Is `manifest` really of type `File`? It's being operated on like a Hash I think?

mwaggett:
okay, yeah, I guess tests wouldn't really make much sense for these. Just wanted to make sure we were thinking about it. 😜

lib/packaging/artifactory.rb
sarameisburger: as far as i can tell it's a JSON file? https://github.com/puppetlabs/enterprise-dist/blob/02470a6d422f1b6e576650d4166c26908cc483bb/tasks/utils.rake#L234

mwaggett: huh.. I didn't know you could operate on that without opening the file? ¯\_(ツ)_/¯

PR #966, opened 2019-08-13
(maint) Add test and release actions

underscorgan:


.github/workflows/release.yml
```diff
@@ -0,0 +1,17 @@
...
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: scarhand/actions-ruby@master
+      - name: Build gem
+        args: build *.gemspec
+      - name: Publish gem
+        env:
+          RUBYGEMS_AUTH_TOKEN: ${{ secrets.RUBYGEMS_AUTH_TOKEN }}
```
mwaggett: Where does this live?

underscorgan: https://github.com/puppetlabs/packaging/settings/secrets

mwaggett: cool! Didn't know about that!

.github/workflows/release.yml
```diff
@@ -0,0 +1,17 @@
...
+        args: push *.gem
```
e-gris: globs. why does it have to be globs? ;-)

PR #967, opened 2019-08-13
(RE-12499) Add function to purge tarballs from artifactory if a compose fails

sarameisburger:
Tested by shipping pe tarballs to artifactory and then deleting them. Will be called in enterprise-dist during compose if not all platform cells succeed in the compose job because we don't want partial ships

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -359,6 +359,31 @@ module Pkg
       end
     end
 
+    # Remove shipped PE tarballs from artifactory
+    # Used when compose fails, we only want the tarball shipped to artifactory if all platforms succeed
+    # Identify which packages were created and shipped based on md5sum and remove them
+    # @param tarball_path [String] the local path to the tarballs that were shipped
+    # @param pe_repo [String] the artifactory repo the tarballs were shipped to
+    def purge_copied_pe_tarballs(tarball_path, pe_repo)
+      check_authorization
+      Dir.foreach("#{tarball_path}/") do |pe_tarball|
+        unless pe_tarball == '.' || pe_tarball == ".."
+          md5 = Digest::MD5.file("#{tarball_path}/#{pe_tarball}").hexdigest
+          artifacts_to_delete = Artifactory::Resource::Artifact.checksum_search(md5: md5, repos: pe_repo)
+          unless artifacts_to_delete.nil?
+            begin
+              artifacts_to_delete.each do |artifact|
+                puts "removing...#{pe_tarball} from #{pe_repo}"
+                artifact.delete
+              end
+            rescue Artifactory::Error::HTTPError
+              STDERR.puts "Error: cannot remove #{pe_tarball}, do you have the right permissions?"
+            end
+          end
+        end
+      end
+    end
+
     private :check_authorization
   end
 end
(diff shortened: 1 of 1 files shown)


lib/packaging/artifactory.rb
```diff
@@ -359,6 +359,31 @@ def ship_pe_tarballs(tarball_path, target_repo, ship_paths)
       end
     end
 
+    # Remove shipped PE tarballs from artifactory
+    # Used when compose fails, we only want the tarball shipped to artifactory if all platforms succeed
+    # Identify which packages were created and shipped based on md5sum and remove them
+    # @param tarball_path [String] the local path to the tarballs that were shipped
+    # @param pe_repo [String] the artifactory repo the tarballs were shipped to
+    def purge_copied_pe_tarballs(tarball_path, pe_repo)
```
e-gris: This is purely a stylistic comment but I find code much easier to read if the main gist of the code isn't inside a conditional. This keeps me from having to hold onto the conditional context while I'm reading the code. In this case, by turning the `unless` statements into `next`, much of it unfolds nicely:

```ruby
Dir.foreach("#{tarball_path}/") do |pe_tarball|
  next if pe_tarball == '.' || pe_tarball == ".."
  md5 = Digest::MD5.file("#{tarball_path}/#{pe_tarball}").hexdigest
  artifacts_to_delete = Artifactory::Resource::Artifact.checksum_search(md5: md5, repos: pe_repo)
  next if artifacts_to_delete.nil?
  begin
    artifacts_to_delete.each do |artifact|
    puts "removing...#{pe_tarball} from #{pe_repo}"
    artifact.delete
  end
  rescue Artifactory::Error::HTTPError
    STDERR.puts "Error: cannot remove #{pe_tarball}, do you have the right permissions?"
  end
end
```

sarameisburger: I'm not super familiar with using `next` but we don't want to continue `if pe_tarball == '.' || pe_tarball == ".." ` so would the unless be replaced with a `break` or am I misunderstanding?

e-gris: 'break' will stop the iteration entirely.
'next' will proceed to the next item in the iteration.

If I'm reading this correctly if pe_tarball == . or .., we just want to skip it and continue onto the next item in the foreach.

sarameisburger: Ahh ok gotcha, I was thinking of it as a "continue" instead of a "move on to the next one." I'll rewrite it with the nexts and test it out

sarameisburger: Ok @e-gris I re-wrote it as suggested and tested with it! I agree that this new way is easier to read

e-gris: break / continue / next / pass / goto / last -- computer languages, gotta love em!

lib/packaging/artifactory.rb
```diff
@@ -359,6 +359,29 @@ def ship_pe_tarballs(tarball_path, target_repo, ship_paths)
...
+      check_authorization
+      Dir.foreach("#{tarball_path}/") do |pe_tarball|
+        next if pe_tarball == '.' || pe_tarball == ".."
+        md5 = Digest::MD5.file("#{tarball_path}/#{pe_tarball}").hexdigest
+        artifacts_to_delete = Artifactory::Resource::Artifact.checksum_search(md5: md5, repos: pe_repo)
+        next if artifacts_to_delete.nil?
+        begin
+          artifacts_to_delete.each do |artifact|
+            puts "removing...#{pe_tarball} from #{pe_repo}"
```
mwaggett: (nitpick) the `...` looks weird in the middle there..

PR #970, opened 2019-08-15
(RE-11765) remove symlinks in the gem

e-gris:
Be nice to Windows users (and generally a good citizen overall) by replacing
symlinks in the spec directory with actual files.

See the Jira for testing notes for this change.

### DIFF
--- a/spec/fixtures/config/ext/build_defaults.yaml
+++ b/spec/fixtures/config/ext/build_defaults.yaml
@@ -1,1 +1,2 @@
-../params.yaml
+---
+:apt_host: foo
--- a/spec/fixtures/config/ext/project_data.yaml
+++ b/spec/fixtures/config/ext/project_data.yaml
@@ -1,1 +1,2 @@
-../params.yaml
+---
+:apt_host: foo
--- a/spec/fixtures/config/params.yaml
+++ b/spec/fixtures/config/params.yaml
@@ -1,2 +1,0 @@
----
-:apt_host: foo
--- a/spec/lib/packaging/config_spec.rb
+++ b/spec/lib/packaging/config_spec.rb
@@ -357,7 +357,7 @@ describe "Pkg::Config" do
         # apt_host: is set to "foo" in the fixture
         orig = Pkg::Config.apt_host
         Pkg::Config.apt_host = "bar"
-        Pkg::Config.config_from_yaml(File.join(FIXTURES, 'config', 'params.yaml'))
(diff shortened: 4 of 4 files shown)


mwaggett:
Want to also remove the whole `spec/fixtures/config/params.yaml` file, if it's not being used anywhere anymore?

e-gris:
I had a moment where it seemed like something might indirectly be using it, but I couldn't fully confirm one way or the other. For the sake of conservatism (knowing that I'm going on PTO soon), I thought we should probably keep it for now.

I can write a ticket to delete them entirely in the short term but I would like to get Glenn unblocked in the short term.

mwaggett:
https://github.com/puppetlabs/packaging/blob/1.0.x/spec/lib/packaging/config_spec.rb#L360 is the only reference to that file that I can find. I think that could just be updated to use the build_defaults file.

If something explodes while you're on PTO, we've got your back. 😊

mwaggett:
@e-gris looks like you may have misunderstood me. I was saying you could remove the `spec/fixtures/config/ext/params.yaml` file, since you had changed the `spec/fixtures/config/ext/build_defaults.yaml` and `spec/fixtures/config/ext/project_data.yaml` files to contain their own data, rather than referencing `params.yaml`.

I don't think this change will work because tests still reference the `build_defaults.yaml` and `project_data.yaml` files. If you wanted to change those to reference `params.yaml`, that could be another approach.

e-gris:
Hmm,

https://github.com/puppetlabs/packaging/blob/c273ce180ce9aa80be1a92ab148180fc645a41c2/spec/lib/packaging/config_spec.rb#L360 still references params.yaml

Without it:

$ bundle exec rspec
Failures:

  1) Pkg::Config #config_from_yaml given a yaml file should, use it to set params
     Failure/Error: Pkg::Config.config_from_yaml(File.join(FIXTURES, 'config', 'params.yaml'))
     RuntimeError:
       There was an error loading data from /home/eric_griswold/git/packaging/spec/fixtures/config/params.yaml.


If I'm reading things correctly, then we should keep all 3 files, just not have any symlinks.

mwaggett:
@e-gris That's what I was saying in my previous comment: you could change https://github.com/puppetlabs/packaging/blob/c273ce180ce9aa80be1a92ab148180fc645a41c2/spec/lib/packaging/config_spec.rb#L360 to reference `spec/fixtures/config/ext/build_defaults.yaml` instead and then `params.yaml` could be safely removed.

e-gris:
@mwaggett OK, done.

PR #971, opened 2019-08-20
(maint) Calculate + ship md5sum of package

sarameisburger:
There are sometimes md5sum mismatch errors when shipping to artifactory, so let's calculate the md5sum of the file we want to ship to artifactory and include it when shipping to ensure accuracy. I'm hoping this will resolve the md5sum mismatch errors we've been seeing when trying to compose (mismatch md5s between packages.json and what lives in artifactory).

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -198,10 +198,13 @@ module Pkg
 
       check_authorization
       artifact = Artifactory::Resource::Artifact.new(local_path: package)
+      artifact_md5 = Digest::MD5.file(package).hexdigest
+      headers = { "X-Checksum-Md5" => artifact_md5 }
       artifact.upload(
         data[:repo_name],
         File.join(data[:alternate_subdirectories], File.basename(package)),
-        deploy_properties(platform_tag)
+        deploy_properties(platform_tag),
+        headers
       )
     rescue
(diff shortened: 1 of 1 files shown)


sarameisburger:
Tested this by shipping a pe-client-tools rpm - instead of getting a warning of `Client did not publish a checksum value.` like in -> https://artifactory.delivery.puppetlabs.net/artifactory/webapp/#/artifacts/browse/tree/General/rpm__local/development/pe-client-tools/71a79fc42e7a4f1afb3b35d863e8a29aec7b9826/el-7-x86_64/pe-client-tools-19.0.5.3.g71a79fc-1.el7.x86_64.rpm it now shows up as ` (Uploaded: Identical)` here https://artifactory.delivery.puppetlabs.net/artifactory/webapp/#/artifacts/browse/tree/General/rpm__local/development/pe-client-tools/59bd1a49e8a7926eaa23975aea314f83f7219e83/el-7-x86_64/pe-client-tools-19.1.3.20.g6b05cb0-1.el7.x86_64.rpm

sarameisburger:
Note: _most_ of the time the md5sum matches what we expect it to be but there are some instances where this is not the case. It's been really difficult to pinpoint the issue because it's so sporadic (ex. pe-client-tools md5sum mismatches in 2018.1 and 2019.1 but not 2019.0 or 2019.2) so hopefully calculating and publishing the sums during ship will resolve this problem or at least give us more insight as to where the problem is occurring

PR #973, opened 2019-08-21
(PE-16465) Implement "all" tarball of repos

e-gris:
Refactor repo.rb a bit so that 'create_all_repo_archives' also
creates a tarball of all the repos. Remove some old duplicating
code.

I've been testing with this Jenkins job: https://jenkins-master-prod-1.delivery.puppetlabs.net/job/experimental_exg_all_repos_test/

The resulting tarballs are being uploaded to Artifactory: https://artifactory.delivery.puppetlabs.net/artifactory/scratchpad__local/experimental_exg_all_repos_test

### DIFF
--- a/lib/packaging/repo.rb
+++ b/lib/packaging/repo.rb
@@ -5,12 +5,12 @@ module Pkg::Repo
     ##
     ## Construct a local_target based upon the versioning style
     ##
-    def construct_local_target_path(versioning)
+    def construct_local_target_path(project, versioning)
       case versioning
       when 'ref'
-        return File.join(Pkg::Config.project, Pkg::Config.ref)
+        return File.join(project, Pkg::Config.ref)
       when 'version'
-        return File.join(Pkg::Config.project, Pkg::Util::Version.dot_version)
+        return File.join(project, Pkg::Util::Version.dot_version)
       else
         fail "Error: Unknown versioning argument: #{versioning}"
       end
@@ -23,7 +23,7 @@ module Pkg::Repo
     def create_signed_repo_archive(repo_location, archive_name, versioning)
       tar = Pkg::Util::Tool.check_tool('tar')
 
-      local_target = construct_local_target_path(versioning)
+      local_target = construct_local_target_path(Pkg::Config.project, versioning)
 
       if Pkg::Util::File.empty_dir?(File.join('pkg', local_target, repo_location))
         if ENV['FAIL_ON_MISSING_TARGET'] == "true"
@@ -44,7 +44,7 @@ module Pkg::Repo
 
     ##
     ## Add a single repo tarball into the 'all' tarball located in
-    ## 'pkg/<local_target>/<Pkg::Config.project>-all.tar'
+    ## 'pkg/<local_target>/<project>-all.tar'
     ## Create the 'all' tarball if needed.
     ##
     def update_tarball_of_all_repos(project, platform, versioning)
@@ -52,7 +52,7 @@ module Pkg::Repo
 
       all_repos_tarball_name = "#{project}-all.tar"
       archive_name = "#{project}-#{platform['name']}"
-      local_target = construct_local_target_path(versioning)
+      local_target = construct_local_target_path(project, versioning)
       repo_tarball_name = "#{archive_name}.tar.gz"
       repo_tarball_path = File.join('repos', repo_tarball_name)
 
@@ -75,7 +75,7 @@ module Pkg::Repo
 
     ##
     ## Invoke gzip to compress the 'all' tarball located in
-    ## 'pkg/<local_target>/<Pkg::Config.project>-all.tar'
+    ## 'pkg/<local_target>/<project>-all.tar'
     ##
     def compress_tarball_of_all_repos(all_repos_tarball_name)
       gzip = Pkg::Util::Tool.check_tool('gzip')
@@ -92,7 +92,7 @@ module Pkg::Repo
     ##
     def create_all_repo_archives(project, versioning)
       platforms = Pkg::Config.platform_repos
-      local_target = construct_local_target_path(versioning)
+      local_target = construct_local_target_path(project, versioning)
       all_repos_tarball_name = "#{project}-all.tar"
 
       platforms.each do |platform|
--- a/lib/packaging/repo.rb
+++ b/lib/packaging/repo.rb
@@ -1,35 +1,110 @@
 module Pkg::Repo
 
   class << self
-    def create_signed_repo_archive(path_to_repo, name_of_archive, versioning)
-      tar = Pkg::Util::Tool.check_tool('tar')
-      Dir.chdir("pkg") do
-        if versioning == 'ref'
-          local_target = File.join(Pkg::Config.project, Pkg::Config.ref)
-        elsif versioning == 'version'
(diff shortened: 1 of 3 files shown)


lib/packaging/repo.rb
```diff
@@ -1,35 +1,110 @@
...
+
+    ##
+    ## Add a single repo tarball into the 'all' tarball located in
+    ## 'pkg/<local_target>/<Pkg::Config.project>-all.tar'
+    ## Create the 'all' tarball if needed.
+    ##
+    def update_tarball_of_all_repos(project, platform, versioning)
+      tar = Pkg::Util::Tool.check_tool('tar')
+
+      all_repos_tarball_name = "#{Pkg::Config.project}-all.tar"
```
mwaggett: If you're using `Pkg::Config.project` here, you probably don't need `project` as a parameter.. or this should just be `"#{project}-all.tar"`.

tasks/nightly_repos.rake
```diff
@@ -68,60 +68,6 @@ DOC
...
     end
 
-    # This is pretty similar to the 'pack_signed_repo' task. The difference here is that instead
-    # of creating a tarball for each repo passed, it adds each repo to a single archive, creating
-    # one 'all' tarball with all of the repos. This is useful for cutomers who have a PE master with
-    # no internet access. They can unpack the puppet-agent-all tarball into the location that
-    # pe_repo expects and use simplified agent install without needing internet access, or having to
-    # manually download each agent that they need to feed to pe_repo.
-    # This task should be invoked after prepare_signed_repos, so that there are repos to pack up.
-    task :pack_all_signed_repos, [:path_to_repo, :name_of_archive, :versioning] => ["pl:fetch"] do |t, args|
```
mwaggett: How did you confirm that it's safe to remove these tasks? I'm all for getting rid of them if they're not being used, but I want to make sure that's the case first.

lib/packaging/repo.rb
```diff
@@ -1,35 +1,110 @@
...
-              raise "ERROR: missing packages under #{path_to_repo}"
-            else
-              warn "Skipping #{name_of_archive} because #{path_to_repo} has no files"
-            end
-          else
-            puts "Archiving #{path_to_repo} as #{name_of_archive}"
-            stdout, _, _ = Pkg::Util::Execution.capture3("#{tar} --owner=0 --group=0 --create --gzip --file #{File.join('repos', "#{name_of_archive}.tar.gz")} #{path_to_repo}")
-            stdout
-          end
+        warn "Warn: Skipping #{archive_name} because #{repo_location} has no files"
```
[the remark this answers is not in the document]
e-gris: Fixed.

lib/packaging/repo.rb
```diff
@@ -1,35 +1,110 @@
...
+      end
+
+      Dir.chdir(File.join('pkg', local_target)) do
+        puts "Info: Archiving #{repo_location} as #{archive_name}"
+        target_tarball = File.join('repos', "#{archive_name}.tar.gz")
+        tar_command = "#{tar} --owner=0 --group=0 --create --gzip --file #{target_tarball} #{repo_location}"
+        puts "Info: Executing tar: #{tar_command}"
```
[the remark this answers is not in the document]
e-gris: -x won't help us here but capture3 debug would. 

IMO, it should be forced on. However, what I had was redundant with that.

lib/packaging/repo.rb
e-gris: I chose the latter.

tasks/nightly_repos.rake
e-gris: I haven't. I'm not sure if there is a way to verify it.

Maybe turn these tasks into no-ops, perhaps?

lib/packaging/repo.rb
mwaggett: This happens in a few other places still, I think.

e-gris: Found (and fixed) 1 more.

tasks/nightly_repos.rake
mwaggett: Well, what led you to delete them in the first place? I see that you've essentially moved all that functionality into methods, which is great, but I guess I would then expect these tasks to just wrap those methods?

But if these aren't being used, then let's def get rid of them. I would first check the usual culprits (i.e. heavy packaging consumers), like ci-job-configs and enterprise-dist. And if they're not there, then I guess let's just remove them? Making them no-op doesn't seem like it will help uncover if/where these are being used, imo. Just another thing to rip out later.

sarameisburger: I think these were the ones made by @shrug but never actually used anywhere. We had chatted about them on monday during our lecture session.

mwaggett: ahhh, gotcha, I missed that 👍

lib/packaging/repo.rb
```diff
@@ -1,35 +1,108 @@
...
+
+    ##
+    ## Construct a local_target based upon the versioning style
+    ##
+    def construct_local_target_path(versioning)
+      case versioning
+      when 'ref'
+        return File.join(Pkg::Config.project, Pkg::Config.ref)
+      when 'version'
+        return File.join(Pkg::Config.project, Pkg::Util::Version.dot_version)
```
mwaggett: These should probably also just be `project` (added as an additional param) if we're using a `project` param elsewhere. Even though the consumers of these will probably pass `Pkg::Config.project` as the `project` param anyway, seems less confusing and potentially more flexible to be consistent.

lib/packaging/repo.rb
```diff
@@ -1,35 +1,108 @@
...
+        target_tarball = File.join('repos', "#{archive_name}.tar.gz")
+        tar_command = "#{tar} --owner=0 --group=0 --create --gzip --file #{target_tarball} #{repo_location}"
+        stdout, _, _ = Pkg::Util::Execution.capture3(tar_command)
+        return stdout
+      end
+    end
+
+    ##
+    ## Add a single repo tarball into the 'all' tarball located in
+    ## 'pkg/<local_target>/<Pkg::Config.project>-all.tar'
```
mwaggett:
```suggestion
    ## 'pkg/<local_target>/<project>-all.tar'
```

lib/packaging/repo.rb
```diff
@@ -1,35 +1,108 @@
...
+
+        tar_command = "#{tar} --owner=0 --group=0 #{tar_action} --file #{all_repos_tarball_name} #{repo_tarball_path}"
+        stdout, _, _ = Pkg::Util::Execution.capture3(tar_command)
+        puts stdout
       end
     end
 
+    ##
+    ## Invoke gzip to compress the 'all' tarball located in
+    ## 'pkg/<local_target>/<Pkg::Config.project>-all.tar'
```
mwaggett:
```suggestion
    ## 'pkg/<local_target>/<project>-all.tar'
```

lib/packaging/repo.rb
e-gris: Updated

lib/packaging/repo.rb
e-gris: Updated

lib/packaging/repo.rb
e-gris: Updated

PR #976, opened 2019-09-03
(maint) Always ship .yaml and .json files

sarameisburger:
Because these files when built are exclusively named after the version of the package
there will be multiple files with that name (ex. 2019.2.0.yaml) so we can't treat these files like other built packages where we do not ship if a package of that name already exists on artifactory

Tested with a set of packages that already existed on artifactory, yaml was shipped but others were not

### DIFF
--- a/tasks/ship.rake
+++ b/tasks/ship.rake
@@ -545,7 +545,9 @@ namespace :pl do
 
       local_dir = args.local_dir || 'pkg'
       Dir.glob("#{local_dir}/**/*").reject { |e| File.directory? e }.each do |artifact|
-        if artifactory.package_exists_on_artifactory?(artifact)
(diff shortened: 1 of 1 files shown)


mwaggett:
If a project is re-built, I'm pretty sure the <ref>.yaml file will be regenerated, even if build artifacts aren't. I don't think we'd want to upload the new <ref>.yaml in that case, since the metadata in there won't necessarily match the metadata from when those artifacts were originally built.

underscorgan:
@mwaggett we currently regenerate and reship the ref.yaml if artifacts are rebuilt. I think not doing that breaks the actual generation of that file since it runs / reships after each individual artifact ships?

mwaggett:
ooh, maybe. I guess that's something separate to address then.

PR #977, opened 2019-09-03
(maint) Prepare for 0.99.41 release

sarameisburger:
Is this more of a "changed" or a "fixed"? because it changes the functionality but it fixes it to what we'd expect


mwaggett:
I would say "fixed", since the original way was arguably a bug? but probably doesn't really matter.

PR #979, opened 2019-09-06
(RE-12729) read LATEST file

sarameisburger:
Tested with LATEST file on artifactory

Because the artifactory gem only supports searching with artifact_name + repo but not a specific path, grab any files that match the name/repo (which we'd expect multiple LATEST files in the generic_enterprise__local repo) but then check their download uris for the provided path and only download the one that matches.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -390,6 +390,20 @@ module Pkg
       end
     end
 
+    # Download an artifact based on name, repo, and path to artifact
+    # @param artifact_name [String] name of artifact to download
+    # @param repo [String] repo the artifact lives
+    # @param path [String] path to artifact in the repo
+    def download_artifact(artifact_name, repo, path)
+      check_authorization
+      artifacts = Artifactory::Resource::Artifact.search(name: artifact_name, repos: repo)
+      artifacts.each do |artifact|
+        if artifact.download_uri.include? path
+          artifact.download('.')
+        end
+      end
+    end
+
     # Remove shipped PE tarballs from artifactory
     # Used when compose fails, we only want the tarball shipped to artifactory if all platforms succeed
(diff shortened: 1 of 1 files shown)


lib/packaging/artifactory.rb
```diff
@@ -390,6 +390,20 @@ def update_latest_file(latest_file, target_repo, path)
...
     end
 
+    # Download an artifact based on name, repo, and path to artifact
+    # @param artifact_name [String] name of artifact to download
+    # @param repo [String] repo the artifact lives
+    # @param path [String] path to artifact in the repo
+    def download_artifact(artifact_name, repo, path)
+      check_authorization
+      artifact_to_download = Artifactory::Resource::Artifact.search(name: artifact_name, repos: repo)
+      artifact_to_download.each do |artifact|
```
e-gris: I'm having difficultly understanding if this method downloads just one artifact or multiple ones.

If it only does one: I don't think the '.each' loop is required.
if it does multiple: The documentation should indicate that and variable names like ```artifact_to_download``` should be plural.

sarameisburger: When doing the initial search with name + repo there's a good chance that it will find more than one file fitting that criteria (like LATEST files in generic_enterprise__local because each PE directory will have one) but then we iterate through them and only select the one that is at the specified path. I wasn't sure if naming it plural or not would be more confusing because in the end we only want one file, but because at the time of populating that variable it could hold multiple artifacts I'll update it to make it plural

sarameisburger: Ok i renamed it to `artifacts` to avoid any confusion

e-gris: :+1: Any variable that _can_ contain a collection should be plural IMO.

PR #984, opened 2019-09-19
(RE-12734) Add functionality for downloading final pe tarballs

sarameisburger:
and also add method to copy tarballs to archives/releases

pairs with https://github.com/puppetlabs/enterprise-dist/pull/3553

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -450,6 +450,36 @@ module Pkg
       end
     end
 
+    # Download final pe tarballs to local path based on name, repo, and path on artifactory
+    # @param pe_version [String] pe final tag
+    # @param repo [String] repo the tarballs live
+    # @param remote_path [String] path to tarballs in the repo
+    # @param local_path [String] local path to download tarballs to
+    def download_final_pe_tarballs(pe_version, repo, remote_path, local_path)
+      check_authorization
+      artifacts = Artifactory::Resource::Artifact.search(name: pe_version, repos: repo)
+      artifacts.each do |artifact|
+        next unless artifact.download_uri.include? remote_path
+        next if artifact.download_uri.include? "-rc"
+        artifact.download(local_path)
+      end
+    end
+
+    # When we ship a new PE release we copy final tarballs to archives/releases
+    # @param pe_version [String] pe final tag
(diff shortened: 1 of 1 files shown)


lib/packaging/artifactory.rb
```diff
@@ -450,6 +450,38 @@ def download_artifact(artifact_name, repo, path)
...
+    # Download final pe tarballs to local path based on name, repo, and path on artifactory
+    # @param pe_version [String] pe final tag
+    # @param repo [String] repo the tarballs live
+    # @param remote_path [String] path to tarballs in the repo
+    # @param local_path [String] local path to download tarballs to
+    def download_final_pe_tarballs(pe_version, repo, remote_path, local_path)
+      check_authorization
+      artifacts = Artifactory::Resource::Artifact.search(name: pe_version, repos: repo)
+      artifacts.each do |artifact|
+        if artifact.download_uri.include? remote_path
```
e-gris: Trivial/optional: I would suggest putting the main logic outside of the 'if' statement:

```ruby

artifacts.each do |artifact|
  next unless artifact.download_uri.include? remote_path
  next if artifact.download_uri.include? "-rc"
  artifact.download(local_path)
end

File: lib/packaging/artifactory.rb [hunk 2]
```diff
@@ -450,6 +450,38 @@ def download_artifact(artifact_name, repo, path)
...
+    # When we ship a new PE release we copy final tarballs to archives/releases
+    # @param pe_version [String] pe final tag
+    # @param repo [String] repo the tarballs live
+    # @param remote_path [String] path to tarballs in the repo
+    # @param target_path [String] path copy tarballs to, assumes same repo
+    def copy_final_pe_tarballs(pe_version, repo, remote_path, target_path)
+      check_authorization
+      final_tarballs = Artifactory::Resource::Artifact.search(name: pe_version, repos: repo)
+      final_tarballs.each do |artifact|
+        if artifact.download_uri.include? remote_path
```
e-gris: Same thing as above.

lib/packaging/artifactory.rb
sarameisburger: Updated! How does that look?

PR #986, opened 2019-09-20
(RE-12765) Add pattern_search class method to Artifactory gem

e-gris:
This was done in an unfortunately awkward way. See the comments in the
artifactory.rb file.

Alternatively, we can write a method that circumvents the Artifactory
gem but that, I think, would be just as sloppy, if not more so.

At the moment, nothing is using this code but it will soon likely be useful.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -53,6 +53,7 @@ module Artifactory
       #
       repo_uri = response['repoUri']
       unless repo_uri.include?('/api/storage/')
+        # rubocop:disable Style/PercentLiteralDelimiters
         repo_uri.sub!(%r(/#{params[:repo]}$), "/api/storage/#{params[:repo]}")
       end
       response['files'].map do |file_path|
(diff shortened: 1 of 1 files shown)


e-gris:
Tested locallyh with enterprise-dist like this:

```
desc "Try out pattern stuff"
task :try_pattern_stuff do
  require 'packaging'
  module Pkg
    class ManageArtifactory
      def test_pattern_stuff
        Artifactory::Resource::Artifact.pattern_search(repo: 'scratchpad__local',
                                                     pattern: '*/*').each do |a|
          pp "Results: ", a
        end
      end
    end
  end

  artifact = Pkg::ManageArtifactory.new(Pkg::Config.project, Pkg::Config.ref)
  artifact.test_pattern_stuff
end
```

mwaggett (approved):
not sure I totally understand how/why you tested the way you did, but if it does what you want, looks good to me!

sarameisburger (approved):
Nice! Can't wait to try it out

PR #988, opened 2019-09-25
(RE-12771) clean out release repos after PE release

sarameisburger:
this is a generalized function that can be used to delete all the artifacts with any specified repo + pattern but we'll use it for clearing out release artifacts from release/ 

tested [here](https://artifactory.delivery.puppetlabs.net/artifactory/rpm_enterprise__local/2018.1/release/el-6-x86_64/) by deleting some artifacts

previously I had also written a pretty ugly method that copies release artifacts back from release/ to repos/ but after thinking about it I realize that this isn't necessary because all packages specified in the base branch's packages.json will already have been promoted to repos/ and this would just be copying over packages that either already exist there or are out of date.

enterprise-dist pr to come

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -567,6 +567,20 @@ module Pkg
       end
     end
 
+    # Remove all artifacts in repo based on pattern, used when we purge all artifacts in release/ after PE release
+    # @param repos [Array] repos that we want to search for artifacts in
+    # @param pattern [String] pattern for artifacts that should be deleted ex. `2019.1/release/*/*`
+    def teardown_repo(repos, pattern)
+      check_authorization
+      repos.each do |repo|
+        artifacts = Artifactory::Resource::Artifact.pattern_search(repo: repo, pattern: pattern)
+        artifacts.each do |artifact|
+          puts "Deleting #{artifact.download_uri}"
+          artifact.delete
+        end
+      end
+    end
+
     # Remove shipped PE tarballs from artifactory
     # Used when compose fails, we only want the tarball shipped to artifactory if all platforms succeed
     # Identify which packages were created and shipped based on md5sum and remove them


sarameisburger:
@mwaggett so I realized that when we merge the release branch back into the mainline (like [here](https://github.com/puppetlabs/enterprise-dist/pull/3522/files) if you ignore the whitespace changes) we are only merging in the commits but not any actual code changes so the packages.json file that's in the mainline remains the same so those packages would have already been promoted to the main repos/ . so the act of copying back artifacts isn't actually necessary

mwaggett:
@sarameisburger in that example yes - I think people have been good about promoting both places. but it's possible that someone could've promoted a package *just* to the release branch, in which case the packages would be missing. If we thing that's rare enough that we want to deal with it ad-hoc, I'm find with that.

sarameisburger:
@mwaggett while that could be true I believe our rule when there's a merge conflict is to favor the code that's in the mainline branch because that's the one people are developing off of.

mwaggett:
@sarameisburger If the version in the release branch is newer than the one in mainline, we have to take that one. usually the mainline version ends up being newer, if people have continued to promote, but I'm imagining a scenario where someone has to land a last-minute fix to the release branch and hasn't promoted to mainline since then. Does that make sense?

sarameisburger:
@mwaggett hmmm ya it makes sense but I think if something has been promoted to the release branch and not the mainline then we should consider that "bad practice" because we shouldn't be adding anything to a release branch that we wouldn't also add to mainline at the same time. So maybe adding something earlier in the process like during promotion to check to see if it's been promoted to mainline repos yet and if not copy it over there as well. Does that seem reasonable?

mwaggett:
I agree that that's bad practice.. hmm. I like that idea, but I'm a liiiittle wary of too much magic (i.e. it can be good to be extra explicit sometimes). What do you think of just leaving this and then adding that check if we see this scenario happening often?

sarameisburger:
@mwaggett works for me 👍

PR #990, opened 2019-09-25
(RE-12732) Add ability to download pe beta tarballs to local directory

sarameisburger:
Tested with 2018.1.0-rc10, was able to download the tarballs

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -529,6 +529,20 @@ module Pkg
       end
     end
 
+    # Download beta pe tarballs to local path based on tag, repo, and path on artifactory
+    # @param beta_tag [String] rc tag of beta release ex. 2019.2.0-rc10
+    # @param repo [String] repo the tarballs live
+    # @param remote_path [String] path to tarballs in the repo
+    # @param local_path [String] local path to download tarballs to
+    def download_beta_pe_tarballs(beta_tag, repo, remote_path, local_path)
+      check_authorization
+      pattern = "#{remote_path}/*-#{beta_tag}-*"
+      artifacts = Artifactory::Resource::Artifact.pattern_search(repo: repo, pattern: pattern)
+      artifacts.each do |artifact|
+        artifact.download(local_path)
+      end
+    end
+
     # When we ship a new PE release we copy final tarballs to archives/releases
     # @param pe_version [String] pe final tag
     # @param repo [String] repo the tarballs live


lib/packaging/artifactory.rb
```diff
@@ -529,6 +529,20 @@ def download_final_pe_tarballs(pe_version, repo, remote_path, local_path)
       end
     end
 
+    # Download beta pe tarballs to local path based on tag, repo, and path on artifactory
+    # @param beta_tag [String] rc tag of beta release ex. 2019.2.0-rc10
+    # @param repo [String] repo the tarballs live
+    # @param remote_path [String] path to tarballs in the repo
+    # @param local_path [String] local path to download tarballs to
+    def download_beta_pe_tarballs(beta_tag, repo, remote_path, local_path)
```
mwaggett: What if we refactored the existing `download_final_pe_tarballs` method to use `pattern_search` and turned it into a more generic `download_artifact_by_tag` or something? And then enterprise-dist can have more specific methods for final vs. beta. My thinking is that a lot of this is essentially doing the same thing, just with different paths and tags, which are already parameters.

lib/packaging/artifactory.rb
```diff
@@ -529,6 +529,20 @@ def download_final_pe_tarballs(pe_version, repo, remote_path, local_path)
...
+      check_authorization
+      pattern = "#{remote_path}/*-#{beta_tag}-*"
```
mwaggett: I know you're calling this with the `archives/internal` path, which shouldn't have dev builds, but I think it would be safer to match against `"#{remote_path}/*-#{beta_tag}$"` so that we can't get matches like `2018.1.10-rc1-141-g4b621dafa`.

mwaggett: oh wait, shit, there's platform info after that.. hmm..

sarameisburger: @mwaggett I was under the impression that $ aren't allowed in the pattern search

lib/packaging/artifactory.rb
sarameisburger: @mwaggett yeah that was my initial thought but when I tried re-writing it using a pattern that worked for both it got funky. so I figured two somewhat similar functions were alright but if we can figure out a pattern that would work for both of them that works for me too.

lib/packaging/artifactory.rb
mwaggett: based on what? you might be right, I haven't tried it, but wondering what gave you that impression.

mwaggett: but anyway, I guess that won't work because of the platform info after the tag

lib/packaging/artifactory.rb
mwaggett: happy to pair/brainstorm with you if you want. if it ends up being too gross, we can keep the separate methods.

lib/packaging/artifactory.rb
sarameisburger: Only `*` and `?` characters are allowed for pattern matching http://ant.apache.org/manual/dirtasks.html#patterns

mwaggett: oh huh. I know @e-gris said the thing he wrote was Ant-like, but I didn't really know what that meant or whether there was a reason - presumably an Artifactory limitation?

lib/packaging/artifactory.rb
sarameisburger: ya I wasn't able to come up with something that didn't get messy. but if you'd like to pair after lunch that works for me.

lib/packaging/artifactory.rb
e-gris: Yup. It's an Artifactory limitation. We *could* do a regex-based one but it's not directly supported by Artifactory so the strategy would be to fetch everything in the repo and filter with the regex.

This is not a terribly expensive operation because of the Artifactory database indices.

PR #992, opened 2019-09-30
(maint) Add generic method for uploading a file to Artifactory

mwaggett:
This commit replaces the `update_latest_file` method with a more generic
`upload_file` method. There was nothing in the `update_latest_file` method that
was actually specific to updating the LATEST file, so this generic method can
be used in its place. This allows for more reusability.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -461,17 +461,20 @@ module Pkg
       end
     end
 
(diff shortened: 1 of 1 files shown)


mwaggett:
@sarameisburger yeah, I have an enterprise-dist PR ready to go, was just testing some stuff before putting it up.

PR #995, opened 2019-10-02
(RE-12793) Allow rake versions >=13.0.0 to be used at runtime

e-gris:


mwaggett:
@e-gris Can you include a summary of this issue in the commit message? When trying to debug, I find it extremely helpful when commit messages include *why* a change was made.

e-gris:
Sure thing. I've been used to the Jira ticket being sufficient but happy to be more descriptive in the commit.

mwaggett:
ahh yeah. I think of the JIRA ticket as the place for *all* the details, but I know I definitely look at commit messages first (and often don't look at JIRA at all) when debugging, so having at least a brief summary with the **why** is helpful to me.

PR #997, opened 2019-10-03
(RE-12788) Set `deb.component` property when copying PE packages

mwaggett:
This commit sets the `deb.component` property on copied debian packages when
populating a new PE repo. We set this to the PE version, which allows the
versioned subdirectories in Artifactory to be treated as separate repository
components. Without this, packages won't show up in the `dists` directory and
repo metadata won't pick them up, meaning we can't download the packages via
package manager.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -580,6 +580,13 @@ module Pkg
           rescue Artifactory::Error::HTTPError
             STDERR.puts "Could not copy #{artifact_target_path}. Source and destination are the same. Skipping..."
           end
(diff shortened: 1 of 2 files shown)


mwaggett:
Tested this by merging up the `packages.json` from 2019.2.x to 9999.9.x and running from the 9999.9.x branch. Packages were added with the `9999.9` component and `9999.9` shows up under `dists` now, e.g. https://artifactory.delivery.puppetlabs.net/artifactory/webapp/#/artifacts/browse/tree/Properties/debian_enterprise__local/dists/bionic/9999.9

PR #998, opened 2019-10-07
(RE-11802) Update the release spreadsheet when shipping

mwaggett:
This commit adds methods for updating the Google Sheet we use to track all our
software releases. These methods get called as part of `uber_ship_lite` so that
we can use the already-defined Pkg::Config.project and Pkg::Config.ref params.

### DIFF
--- a/lib/packaging/util/date.rb
+++ b/lib/packaging/util/date.rb
@@ -11,5 +11,10 @@ module Pkg::Util::Date
       end
       Time.now.strftime(format)
     end
+
+    def today
+      format = "%m/%d/%Y"
+      Time.now.strftime(format)
+    end
   end
 end
--- a/lib/packaging.rb
+++ b/lib/packaging.rb
@@ -19,6 +19,7 @@ module Pkg
   require 'packaging/retrieve'
   require 'packaging/sign'
   require 'packaging/archive'
+  require 'packaging/metrics'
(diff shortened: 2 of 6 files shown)


mwaggett:
this is my test spreadsheet: https://docs.google.com/spreadsheets/d/1Kvz3lJ_xymk-H4DsyAApOeT6NDjSArH7KYukWegx9-A/edit?usp=sharing

lib/packaging/util/date.rb
```diff
@@ -11,5 +11,10 @@ def timestamp(separator = nil)
       end
       Time.now.strftime(format)
     end
+
+    def today
```
e-gris: IMO, it would be nice to use an ISO 8601 date format to keep this from being US-centric.

I *think* it's possible to have the existing ones converted so there's not a sudden change.

mwaggett: I get that, but I think there are some formulas in the spreadsheet that expect this format. If this is something we want to do, I think it should be a separate effort, but seeing as how wishy-washy the need for this spreadsheet is anyway, I'm hesitant to make changes.

mwaggett:
welp there is no Google account for jenkins - the plan is now to just put all the spreadsheet data into a csv file in a git repo and update it there.

PR #1000, opened 2019-10-09
(RE-11802) Add task for updating release-metrics

mwaggett:
This commit adds an `update_release_metrics` task for adding a release to the release-metrics repo (previously the release spreadsheet). This task gets called during `uber_ship_lite` in order to retain the context of what was just shipped.

### DIFF
--- a/tasks/jenkins.rake
+++ b/tasks/jenkins.rake
@@ -267,6 +267,12 @@ namespace :pl do
       end
       # mark the build as successfully shipped
       Rake::Task["pl:jenkins:ship"].invoke("shipped")
+      # add the release to release-metrics
+      begin
+        Rake::Task["pl:update_release_metrics"].invoke
+      rescue => e
+        puts "Uh oh! Something went wrong updating release-metrics:\n#{e}\nPlease add this release manually. Proceeding..."
+      end
     end
 
     task :stage_nightlies => "pl:fetch" do
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,9 @@ sudo: false
 notifications:
   email: false
 rvm:
-  - 2.0.0
-  - 2.1.9
   - 2.3.8
   - 2.4.1
+  - 2.6.5
 
 env:
   - "CHECK='rspec spec --color --format documentation --order random'"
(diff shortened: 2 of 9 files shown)


mwaggett:
Tested this by running the `pl:update_release_metrics` task from the bolt-vanagon repo to get this: https://github.com/puppetlabs/release-metrics/commit/58af80f0a9ee21d21777e479a35fcce34d547aaa

tasks/jenkins.rake
```diff
@@ -267,6 +267,12 @@ namespace :pl do
       end
       # mark the build as successfully shipped
       Rake::Task["pl:jenkins:ship"].invoke("shipped")
+      # add the release to release-metrics
+      begin
+        Rake::Task["pl:update_release_metrics"].invoke
+      rescue => e
+        puts "Uh oh! Something went wrong updating release-metrics:\n#{e}\nPlease add this release manually. Proceeding..."
```
sarameisburger: Suggestion: using `fail` instead of `puts`, that way the user can immediately know that this failed and they can look into the cause, otherwise I think they'd either have to check the release-metrics repo after every release to make sure it succeeded or review the log

mwaggett: I eventually want to do that, but because this is called during `uber_ship_lite`, which is run in the stage-artifacts job, if this fails, the whole ship pipeline stops. I don't want updating this data to get in the way of shipping until we've had a few successful runs and are more confident in this automation. Does that seem reasonable?

mwaggett: I eventually want to do that, but because this is called in `uber_ship_lite`, if this fails, the whole ship pipeline stops. I don't want updating this data to get in the way of shipping, so I was gonna leave it like this until we've had a few successful runs and are more confident in this automation. Does that seem reasonable?

sarameisburger: ya that's just, just wanted to make sure it was intentional.

sarameisburger:
did you want to update the changelog?

mwaggett:
ahh yes, thank you for the reminder!

mwaggett:
@sarameisburger changelog has been updated!

PR #1006, opened 2019-10-22
(maint) Set debian component properly when populating repos

underscorgan:


underscorgan:
I tested the new definition in irb and it behaved as expected

PR #1007, opened 2019-11-07
(RE-12874) Re-do the Artifactory 'cleanup.skip' logic

e-gris:
This is a *breaking change* and will need to be coordinated with
changes to enterprise-dist.

This changes the way we handle the Artifactory 'cleanup.skip' logic.
Rather than try to calculate it at the time we upload the tarballs,
we set the 'cleanup.skip' property when we upload the 'LATEST' file.

The former approach had the problem that the version string for the tarballs
that we wanted to keep was unavailable. This approach simply uses
the string in the LATEST file as a pattern to keep.

It starts by setting 'cleanup.skip' to *false* on all the files in the target
Artifactory directory, then sets 'cleanup.skip' to *true* for all tarballs
containing the string in the LATEST file.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -432,28 +432,14 @@ module Pkg
     def ship_pe_tarballs(local_tarball_directory, target_repo, ship_paths)
       check_authorization
       ship_paths.each do |path|
-        unset_cleanup_skip_on_artifacts(target_repo, path)
         Dir.foreach(local_tarball_directory) do |pe_tarball|
           next if pe_tarball == '.' || pe_tarball == ".."
           begin
             puts "Uploading #{pe_tarball} to #{target_repo}/#{path}#{pe_tarball}"
             artifact = Artifactory::Resource::Artifact.new(
               local_path: "#{local_tarball_directory}/#{pe_tarball}")
-            uploaded_artifact = artifact.upload(target_repo, "#{path}#{pe_tarball}")
-            # The Artifactory gem property setter only works when '/api/storage' is used in
-            # the 'uri' field.
-            # Strangely, the above Artifactory::Resource::Artifact.new gives us the raw URI.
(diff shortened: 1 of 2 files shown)


shrug:
lgtm, prevent_artifact_cleanup will get called from enterprise-dist when the LATEST file gets shipped?

e-gris:
@shrug Yes. I'll that that PR up in a moment.

lib/packaging/artifactory.rb
```diff
@@ -478,29 +464,43 @@ def upload_file(local_path, target_repo, target_path)
...
-        ARTIFACTORY_CLEANUP_SKIP_PROPERTY => true,
-        "repos" => repo
+    # @param pe_build_version [String] Set 'cleanup.skip' property on artifacts that
+    #   contain this string in their file inside the directory.
+    def prevent_artifact_cleanup(repo, directory, pe_build_version)
+      # Clean up any trailing slashes on directory, just in case
+      directory.sub!(/(\/)+$/, '')
+
+      all_artifacts_pattern = "#{directory}/*"
+      latest_artifacts_pattern = "#{directory}/*#{pe_build_version}*"
```
mwaggett: What happens when the latest version is a tagged version? e.g. `2019.2.0` will match all the rcs and stuff prior to the tag.
I guess this doesn't really matter, since the consequence would be setting `cleanup.skip` on _more_ stuff, but wanted to point it out just in case there's something I'm not thinking of.

mwaggett:
Could you please squash your commits if they're not in meaningful chunks? and if they are, could you update the commit messages to describe the separate parts of the change?

lib/packaging/artifactory.rb
e-gris: That's exactly the scenario I had to work through. Overtagging 'cleanup.skip' here is a conservative action.

PR #1008, opened 2019-11-08
(RE-12869) Refactor `ship_gem_to_downloads` task

mwaggett:
This commit reorganizes the `ship_gem_to_downloads` task to be clearer about
the cases where we do and don't want to actually attemp shipping to downloads.
Previously, the `exit` that was called to prevent the upload in the case of
unset params was preventing downstream tasks from getting run. For example,
because Bolt ships gems to rubygems, but not downloads, they have `build_gem`
set to true but no `gem_host` or `gem_path`. When shipping, none of the
`uber_ship_lite` tasks after `ship_gem_to_downloads` were getting run.


mwaggett:
I was able to reproduce the issue by running `uber_ship_lite` from bolt-vanagon and then confirmed that this fixes it.

PR #1011, opened 2019-11-15
(maint) Include filename when copying in Artifactory

mwaggett:
This commit updates the `copy_final_pe_tarballs` method to include the filename
in the target path when copying to archives/releases. Previously, when trying to
copy, instead of creating a directory to copy into, it would create a single
file with the desired directory name.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -557,7 +557,11 @@ module Pkg
       final_tarballs.each do |artifact|
         next unless artifact.download_uri.include? remote_path
         next if artifact.download_uri.include? "-rc"
-        artifact.copy("#{repo}/#{target_path}")
+        filename = File.basename(artifact.download_uri)
+        # Artifactory does NOT like when you use `File.join`, so let's concatenate!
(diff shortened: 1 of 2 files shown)


mwaggett:
ahh shit forgot the changelog, one sec

lib/packaging/artifactory.rb
```diff
@@ -557,7 +557,11 @@ def copy_final_pe_tarballs(pe_version, repo, remote_path, target_path)
       final_tarballs.each do |artifact|
         next unless artifact.download_uri.include? remote_path
         next if artifact.download_uri.include? "-rc"
-        artifact.copy("#{repo}/#{target_path}")
+        filename = File.basename(artifact.download_uri)
+        # Artifactory does NOT like when you use `File.join`, so let's concatenate!
```
underscorgan: weirdddddddd

e-gris: Imma going to try to hunt that down to understand *why*. It may result in a ticket to JFrog. Not today, though.

PR #1016, opened 2019-12-05
(PA-2987) Handle puppet nightly gems

gimmyxd:
Scope of this pull-request is to add possibility build nightly gem and to create `#{gem_name}-latest-#{platform}`symlinks.

Usage: 
```
package:nightly_gem
pl:remote:link_nightly_shipped_gems_to_latest
```

Eg. result :
```
puppet-6.11.0.79.g61d0522-x86-mingw32.gem
puppet-6.11.0.79.g61d0522-x86-mingw32.gem -> puppet-6.11.0.73-x64-mingw32.gem
```

Having `-latest` version would let puppet nightly gem consumers to download the latest nightly gem with minimal effort.

I'm thinking on structuring the gems in nightlies.puppet.com like this:
```
puppet 5:
   /downloads/gems/puppet5-nightly/puppet-latest.gem
   /downloads/gems/puppet5-nightly/puppet-x64-mingw32.gem
etc...

puppet6: 
  /downloads/gems/puppet6-nightly/puppet-latest.gem
etc...
```

### DIFF
--- a/tasks/gem.rake
+++ b/tasks/gem.rake
@@ -127,23 +127,33 @@ def create_platform_specific_gems
   end
 end
 
+def package_gem
+  unless Pkg::Config.build_gem
+    puts "The #{Pkg::Config.project} project does not build any gems! Passing through ..."
+    return
+  end
+
+  mkdir_p File.join(Pkg::Config.project_root, "pkg")
+  create_default_gem
+  if Pkg::Config.gem_platform_dependencies
+    create_platform_specific_gems
+  end
+end
+
 namespace :package do
   desc "Build a gem - All gems if platform specific"
   task :gem => ["clean"] do
-    unless Pkg::Config.build_gem
-      puts("The #{Pkg::Config.project} project does not build any gems! Passing through ...")
-      next
-    end
+    package_gem
+  end
 
-    mkdir_p File.join(Pkg::Config.project_root, "pkg")
-    create_default_gem
-    if Pkg::Config.gem_platform_dependencies
-      create_platform_specific_gems
-    end
+  task :nightly_gem => ["clean"] do
+    Pkg::Config.gemversion = Pkg::Util::Version.extended_dot_version
+    package_gem
   end
 end
 
 # An alias task to simplify our remote logic in jenkins.rake
 namespace :pl do
   task :gem => "package:gem"
+  task :nightly_gem => "package:nightly_gem"
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
 - Add windowsfips versionless symlinks
+- Add rake task to build nightly gem packages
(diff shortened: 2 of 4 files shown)


tasks/gem.rake
```diff
@@ -141,6 +149,13 @@ namespace :package do
       create_platform_specific_gems
     end
   end
+
+  namespace :gem do
+    desc "Creates symlinks to gem_name-latest-platform"
+    task :create_latest_symlinks do
```
e-gris: Minor: Is there a reason why the task is named 'create_latest_symlinks' but the method is called 'create_gem_latest_symlinks'? Parity would seem to be desirable but not sure if there's other factors involved.

gimmyxd: Not really, I just put up some quick code to gather feedback on the direction. I can rename the method

e-gris: Certainly a minor point but desirable to do, IMO.

gimmyxd:
Moved the task to `ship.rake` and will be called after `ship_nightly_gem` task.
The rake task will create symlinks for the gems that were shipped to `-latest`.
Eg:

```
 puppet-6.11.0.72.dirty.gem
 puppet-6.11.0.72.dirty-universal-darwin.gem
 puppet-6.11.0.72.dirty-x64-mingw32.gem
 puppet-6.11.0.72.dirty-x86-mingw32.gem
 puppet-latest.gem -> /opt/repository-nightlies/downloads/gems/puppet-6.11.0.72.dirty.gem
 puppet-latest-universal-darwin.gem -> /opt/repository-nightlies/downloads/gems/puppet-6.11.0.72.dirty-universal-darwin.gem
 puppet-latest-x64-mingw32.gem -> /opt/repository-nightlies/downloads/gems/puppet-6.11.0.72.dirty-x64-mingw32.gem
 puppet-latest-x86-mingw32.gem -> /opt/repository-nightlies/downloads/gems/puppet-6.11.0.72.dirty-x86-mingw32.gem
```

gimmyxd:
@underscorgan sorry, i reworked the PR since I missed the versioning for nightly gems
To summarise this:
 - `package:nightly_gem` - builds nightly gems containing the short ref in name: `puppet-6.11.0.79.g61d0522-x86-mingw32.gem`
- `pl:remote:link_nightly_shipped_gems_to_latest` symlinks the latest built gems to `puppet-latest` corespondents.

tasks/gem.rake
```diff
@@ -127,23 +127,32 @@ def create_platform_specific_gems
   end
 end
 
+def package_gem
+  unless Pkg::Config.build_gem
+    abort("The #{Pkg::Config.project} project does not build any gems! Passing through ...")
```
underscorgan: This is a change in behavior, since this will cause the process to exit with exit code `1` whereas it looks like previously it wouldn't return errors. I think you maybe want
```ruby
puts "The #{Pkg::Config.project} project does not build any gems! Passing through ..."
return
```

instead of the abort

mihaibuzgau: updated the code

PR #1034, opened 2020-02-14
(maint) Add `search_with_path` artifactory method

underscorgan:
- [x] docs
- [x] changelog


underscorgan:
@sarameisburger I think leaving that to whatever's calling the method should be fine, my changes to `download_artifact` were just moving the search logic into `search_with_path` and we weren't previously printing that.

PR #1037, opened 2020-03-03
(RE-13163) Update code signing cert for solaris signing

underscorgan:
New cert unpacked on rama as signing_key_2020.pem and
signing_cert_2020.pem.


sarameisburger:
💯

PR #1039, opened 2020-03-06
(maint) Update `download_packages` to optionally add a search path

underscorgan:

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -408,17 +408,22 @@ module Pkg
     # Using the manifest provided by enterprise-dist, grab the appropropriate packages from artifactory based on md5sum
     # @param staging_directory [String] location to download packages to
     # @param manifest [File] JSON file containing information about what packages to download and the corresponding md5sums
-    # @param path [String] path matching for where the package should be downloaded from. Defaults to '' (no custom path for searching)
-    def download_packages(staging_directory, manifest, path = '')
+    # @param remote_path [String] Optional partial path on the remote host containing packages
+    #        Used to specify which subdirectories packages will be downloaded from.
+    def download_packages(staging_directory, manifest, remote_path = '')
       check_authorization
       manifest.each do |dist, packages|
         puts "Grabbing the #{dist} packages from artifactory"
         packages.each do |name, info|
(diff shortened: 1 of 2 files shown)


lib/packaging/artifactory.rb
```diff
@@ -408,15 +408,17 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
...
       manifest.each do |dist, packages|
         puts "Grabbing the #{dist} packages from artifactory"
         packages.each do |name, info|
-          artifact_to_download = Artifactory::Resource::Artifact.checksum_search(md5: "#{info["md5"]}", repos: ["rpm_enterprise__local", "debian_enterprise__local"]).first
+          artifacts = Artifactory::Resource::Artifact.checksum_search(md5: "#{info["md5"]}", repos: ["rpm_enterprise__local", "debian_enterprise__local"])
+          artifact_to_download = artifacts.select { |artifact| artifact.download_uri.include? path }.first
           if artifact_to_download.nil?
             filename = info["filename"] || info["name"]
-            raise "Error: what the hell, could not find package #{filename} with md5sum #{info["md5"]}"
+            raise "Error: what the hell, could not find package #{filename} with md5sum #{info["md5"]} in #{path}"
```
sarameisburger: if we don't set path will this message say "with md5sum <sum> in "? not sure if it really matters but might confuse when debugging

lib/packaging/artifactory.rb
```diff
@@ -408,15 +408,17 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
     # Using the manifest provided by enterprise-dist, grab the appropropriate packages from artifactory based on md5sum
     # @param staging_directory [String] location to download packages to
     # @param manifest [File] JSON file containing information about what packages to download and the corresponding md5sums
-    def download_packages(staging_directory, manifest)
+    # @param path [String] path matching for where the package should be downloaded from. Defaults to '' (no custom path for searching)
```
e-gris: Parsing the doc is a bit tough for me. Maybe something like:
```
# @param remote_path [String] Optional file path on remote machine containing packages.
#  Used for <reason to use optional path>
```

underscorgan:
updated!

PR #1040, opened 2020-03-10
[RFC] Moar artifactory

underscorgan:
This builds on #1039 and adds logic so that the compose jobs can be set to be self-healing (if they find packages but they aren't in the correct subdir on artifactory, copy it be in the correct subdir).

Mostly, when working on https://github.com/puppetlabs/enterprise-dist/pull/3657 I was updating repos with all the missing packages and that was tedious and felt like something that was automatable. 

Need to do changelog first, but wanted to validate that we were good with this sort of change before hand

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -417,6 +417,17 @@ module Pkg
         packages.each do |name, info|
           artifacts = Artifactory::Resource::Artifact.checksum_search(md5: "#{info["md5"]}", repos: ["rpm_enterprise__local", "debian_enterprise__local"])
           artifact_to_download = artifacts.select { |artifact| artifact.download_uri.include? remote_path }.first
+          # If we found matching artifacts, but not in the correct path, copy the artifact to the correct path
+          # This should help us keep repos up to date with the packages we are expecting to be there
+          # while helping us avoid 'what the hell, could not find package' errors
+          if artifact_to_download.nil? && !artifacts.empty?
+            artifact_to_copy = artifacts.first
+            filename = info['filename'] || File.basename(artifact_to_copy.download_uri)
+            copy_artifact(artifact_to_copy, artifact_to_copy.repo, "#{remote_path}/#{dist}/#{filename}")
+            artifacts = Artifactory::Resource::Artifact.checksum_search(md5: "#{info["md5"]}", repos: ["rpm_enterprise__local", "debian_enterprise__local"])
+            artifact_to_download = artifacts.select { |artifact| artifact.download_uri.include? remote_path }.first
+          end
+
           if artifact_to_download.nil?
             filename = info["filename"] || name
             message = "Error: what the hell, could not find package #{filename} with md5sum #{info["md5"]}"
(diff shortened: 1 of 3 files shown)


lib/packaging/paths.rb
```diff
@@ -303,10 +303,17 @@ def release_package_link_path(platform_tag, nonfinal = false)
     end
   end
 
-  def debian_component_from_path(path)
-    matches = path.match(/(\d+\.\d+)\/(\w+)/)
+  def debian_component_from_path(path, regex_for_substitution: nil, substitute_with: '_')
```
e-gris: I'm wondering if the regex and replacement can't be hard-coded since the pattern, I think, should be fixed (and prescriptively so).

Is there a scenario where regex_for_substitution is legitimately nil when this is called?

lib/packaging/artifactory.rb
```diff
@@ -410,13 +410,24 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
     # @param manifest [File] JSON file containing information about what packages to download and the corresponding md5sums
     # @param remote_path [String] Optional partial path on the remote host containing packages
     #        Used to specify which subdirectories packages will be downloaded from.
-    def download_packages(staging_directory, manifest, remote_path = '')
+    # @param fix_errors [Boolean]
+    def download_packages(staging_directory, manifest, remote_path = '', fix_errors: false)
```
e-gris: I want to think the fix_errors flag isn't really needed (why shouldn't we fix errors when we find them? :0)

However, I'm guessing that artifact_to_download.nil? && !artifacts.empty? could evaluate to true but we don't want to fix it?

lib/packaging/paths.rb
underscorgan: I think I was maybe aiming for too much future-proofing / flexibility given how very specific the rest of this method is, will update :)

lib/packaging/artifactory.rb
underscorgan: Fair point.

sarameisburger: Maybe add a little more info here on what will happen if fix_errors == true?

lib/packaging/artifactory.rb
```diff
@@ -410,13 +410,24 @@ def promote_package(pkg, ref, platform_tag, repository, debian_component = nil)
...
       check_authorization
       manifest.each do |dist, packages|
         puts "Grabbing the #{dist} packages from artifactory"
         packages.each do |name, info|
           artifacts = Artifactory::Resource::Artifact.checksum_search(md5: "#{info["md5"]}", repos: ["rpm_enterprise__local", "debian_enterprise__local"])
           artifact_to_download = artifacts.select { |artifact| artifact.download_uri.include? remote_path }.first
+          # This is only going to actually work if remote_path is set, but not checking for that here
+          # since it's not possible for remote_path to not be set, artifact_to_download to be nil, and artifacts to not be empty
+          if artifact_to_download.nil? && !artifacts.empty? && fix_errors
```
sarameisburger: or add that info here

lib/packaging/artifactory.rb
```diff
@@ -417,6 +417,17 @@ def download_packages(staging_directory, manifest, remote_path = '')
         packages.each do |name, info|
           artifacts = Artifactory::Resource::Artifact.checksum_search(md5: "#{info["md5"]}", repos: ["rpm_enterprise__local", "debian_enterprise__local"])
           artifact_to_download = artifacts.select { |artifact| artifact.download_uri.include? remote_path }.first
+          # If we found matching artifacts, but not in the correct path, copy the artifact to the correct path
+          # This should help us keep repos up to date with the packages we are expecting to be there
+          # while helping us avoid 'what the hell, could not find package' errors
```
underscorgan: How's this @sarameisburger ?

underscorgan:
Just need to update changelog then this should be ready to go i think!

lib/packaging/artifactory.rb
sarameisburger: perf 💯

underscorgan:
@sarameisburger want to give the changelog quick 👀 ?

PR #1043, opened 2020-03-18
(RE-13225) Allow property and header settings in Pkg::ManageArtifactory#upload_file

e-gris:
Two parameters, `properties` and `headers` are supported by the Resource::Artifact class of
the Artifactory gem.

Add this support to our `upload_file` method, mostly for the purpose of RE-13225, which
is to allow 'cleanup.skip' to be set on uploaded artifacts.

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -477,14 +477,19 @@ module Pkg
     # @param local_path [String] local path to file to upload
     # @param target_repo [String] repo on artifactory to upload to
     # @param target_path [String] path within target_repo to upload to
-    def upload_file(local_path, target_repo, target_path)
+    # @param properties [Hash] Optional property names and values to assign the uploaded file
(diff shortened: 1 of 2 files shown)


lib/packaging/artifactory.rb
```diff
@@ -472,14 +472,17 @@ def ship_pe_tarballs(local_tarball_directory, target_repo, ship_paths)
     # @param local_path [String] local path to file to upload
     # @param target_repo [String] repo on artifactory to upload to
     # @param target_path [String] path within target_repo to upload to
-    def upload_file(local_path, target_repo, target_path)
+    # @param properties [Hash] Optional property names and values to assign the uploaded file
```
sarameisburger: maybe include an example of what something in this hash might look like?

e-gris: @sarameisburger Updated!

e-gris:
... Needed rebasing

PR #1045, opened 2020-03-20
(RE-13177)(RE-13180)(RE-13183)(RE-13186) Remove old platforms + update spec

sarameisburger:
removed:
      - cumulus-2.2-amd64
      - debian-7-amd64
      - debian-7-i386
      - eos-4-i386
      - fedora-f25-i386
      - fedora-f25-x86_64
      - fedora-f26-x86_64
      - fedora-25-i386
      - fedora-25-x86_64
      - fedora-26-x86_64
      - fedora-27-x86_64
      - fedora-28-x86_64
      - fedora-29-x86_64
      - osx-10.10-x86_64
      - osx-10.11-x86_64
      - osx-10.12-x86_64
      - osx-10.13-x86_64
      - sles-11-i386
      - sles-11-x86_64
      - ubuntu-14.04-amd64
      - ubuntu-14.04-i386
    removed s390x from platforms as well

and 
clean up spec tests based on platform removals, we removed some platforms which will cause spec tests to fail, cleaned up some tests to use existing platforms.

should the changelog include all platforms we removed or just say "removed eol platforms"? ^

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 ### Fixed
 - (RE-9687) Set `deb.architecture` correctly in artifactory for noarch packages.
 
+### Removed
+- (RE-13177)(RE-13180)(RE-13183)(RE-13186) Removed EOL platforms and updated spec tests
+  accordingly.
+
 ## [0.99.59] - 2020-03-11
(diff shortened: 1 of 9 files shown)


mihaibuzgau:
We still support the following platforms on the Puppet Agent side the following OSes:
* SLES 11 (only s390x was removed)
* Ubuntu 14 (both i386 and amd64)

PR #1050, opened 2020-04-10
Hack: refactor from case/loop invariant code

e-gris:
After RE-13273 I noticed that this could use some cleanup of invariant code inside the cases and loops.

This seems more readable to me. Curious if everyone else does.

### DIFF
--- a/lib/packaging/paths.rb
+++ b/lib/packaging/paths.rb
@@ -101,27 +101,81 @@ module Pkg::Paths
     return Pkg::Config.repo_link_target
   end
 
-  def artifacts_base_path_and_link_path(platform_tag, path_prefix = 'artifacts', nonfinal = false)
-    platform, version, architecture = Pkg::Platforms.parse_platform_tag(platform_tag)
-    package_format = Pkg::Platforms.package_format_for_tag(platform_tag)
+  # Construct a platform-dependent symlink target path.
+  def construct_base_path(path_data)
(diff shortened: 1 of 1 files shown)


lib/packaging/paths.rb
```diff
@@ -101,29 +101,85 @@ def link_name(nonfinal = false)
...
     else
-      raise "Not sure where to find packages with a package format of '#{package_format}'"
+      raise "Error: Unknown package format '#{package_format}'"
     end
   end
 
+  # Given platform information, create symlink target (base_path) and link path in the
+  # form of a 2-element array
+  def artifacts_base_path_and_link_path(platform_tag, prefix = 'artifacts', is_nonfinal = false)
+    platform_name, version, architecture = Pkg::Platforms.parse_platform_tag(platform_tag)
```
underscorgan, line 163: [nitpick] It doesn't look like we use the version or architecture from `path_data` in either of the methods, may make sense to get rid of those and change this line to
```suggestion
    platform_name, _ = Pkg::Platforms.parse_platform_tag(platform_tag)
```

Feel free to ignore :)

e-gris: Great catch. Updated.

e-gris:
Given that this is totally internal with no user changes, I'm going to forgo a Changelog entry. If anyone else things I need it, I'll patch one in.

PR #1051, opened 2020-04-17
(RE-12065)(RE-12069) Remove s390x for sles and el

sarameisburger:
update spec tests accordingly - I swapped s390x for i386 in the tests, if that's a bad idea I can change it or just remove it altogether

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -103,7 +103,7 @@ module Pkg
           repo: true,
         },
         '6' => {
-          architectures: ['x86_64', 'i386', 's390x'],
+          architectures: ['x86_64', 'i386'],
           source_architecture: 'SRPMS',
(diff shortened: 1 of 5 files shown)


e-gris:
Did we just leave off i386 before because it would be over-testing?
If that's the case, it seems fine to me to replace s390x with i386.

underscorgan:
@sarameisburger are any projecs/pipelines still building for s390x? If so we should get the platform removed there before it's removed from packaging since that will break shipping.

sarameisburger:
@underscorgan From what I can tell there aren't any? But I don't know how to extensively check. PE and Puppet Agent have EOL'd it and I'm not seeing it in recent builds of other projects like puppetserver and puppetdb

underscorgan:
I just did a quick scan in ci-job-configs and I'm not seeing anything, so 👍

PR #1052, opened 2020-04-23
Revert "Revert "Merge pull request #1045 from sarameisburger/remove-p…

sarameisburger:
…latforms""

We previously had removed a handful of platforms but had to revert due to some unforeseen failures. This reverts that revert but also no longer removes eos-4/swix because it seems that we are still shipping that for 5.5.x http://builds.delivery.puppetlabs.net/puppet-agent/5.5.19/shipped/eos/4/puppet5/i386/?C=M&O=D

I used https://github.com/puppetlabs/ci-job-configs/blob/master/jenkii/platform/projects/puppet-agent.yaml#L74 to determine the platforms

This reverts commit cf49b4bf74ae3f7556a230e414c6ab76a8f56792.

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -47,26 +47,7 @@ module Pkg
         },
       },
 
-      'cumulus' => {
-        '2.2' => {
-          codename: 'cumulus',
-          architectures: ['amd64'],
-          source_architecture: 'source',
-          package_format: 'deb',
-          source_package_formats: DEBIAN_SOURCE_FORMATS,
-          repo: true,
-        },
-      },
-
       'debian' => {
-        '7' => {
-          codename: 'wheezy',
-          architectures: ['amd64', 'i386'],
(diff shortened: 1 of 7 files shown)


sarameisburger:
@underscorgan I think this is all that needs to be done? I couldn't find anything else obvious on ci-job-configs to update but entirely possible I missed something

e-gris:
Let me see if I understand: the _original_ removal had problems because of eos-4/swix issues so we are re-issuing the removal but exempting eos-4/swix, correct?

sarameisburger:
@e-gris yup pretty much. The platform removal caused failures for the puppet-agent team because it was looking for a platform that wasn't there. As far as I can tell eos-4 is the only one that was still declared in ci-job-configs but removed from packaging but @underscorgan understands/remembers the failure better than I do so I wanna make sure I'm doing the right thing here

sarameisburger:
ok merging!

PR #1053, opened 2020-04-23
(RE-13386) Don't ship deb packages to a 'pool' subdirectory by default

underscorgan:


sarameisburger:
Is this gonna break anything if we release with this change/do we need to announce it?

underscorgan:
Yeah, I'm not seeing anything obvious that should break, +1 to failing forward with any issues that arise :)

PR #1060, opened 2020-06-30
(RE-13303) Update rolling repo link creation

underscorgan:


sarameisburger:
not sure if there's anywhere we need to note about the tasks being separated? just remember to run them when puppet7 goes out?

underscorgan:
I'll find somewhere on confluence to include it!

PR #1066, opened 2020-08-10
(RE-13629) Do not use exact file match when downloading final PE tarballs

underscorgan:


underscorgan:
Going to test this with the s3 ship job, will merge and release once that goes in

PR #1076, opened 2020-10-20
Add new apt pathing for puppet7

e-gris:
Update specs to puppet 6 pathing; prepratory work for puppet 7.


e-gris:
Also, new repo tests can't be allowed to fall-through. Need to positively identify each case and error, rather than default.

underscorgan (approved):

PR #1077, opened 2020-10-21
(maint) make curl less noisy

e-gris:
When curl is fetching <gem>.json from rubygems, it prints download
information that is unhelpful during batch jobs. This turns that off
by adding '--silent' to the curl call.

There's also, in a seperate commit, some code formatting and output
message cleanups.


underscorgan (approved):

underscorgan:
Maybe add a changelog note?

PR #1094, opened 2021-01-29
(RE-11821) Add config validation after pl:fetch

e-gris:
Introduce a hook that allows for data validation after a pl:fetch.

Create an initial simple validation per the suggestion in RE-11821
that checks if 'project' is non-empty.


underscorgan (approved):

underscorgan:
I like this! Seems like an elegant solution :)

PR #1096, opened 2021-02-19
(RE-12465) Refactor remote_ssh_cmd

e-gris:
Create a new method, #remote_execute and leave #remote_ssh_cmd as a
shim.

This allows for the #remote_execute method to be called with more
clarity and simplicity.


e-gris:
Needs some more testing, still.

lib/packaging/sign/dmg.rb
```diff
@@ -15,10 +15,10 @@ def sign(target_dir = 'pkg')
     work_dir  = "/tmp/#{Pkg::Util.rand_string}"
     mount     = File.join(work_dir, "mount")
     signed    = File.join(work_dir, "signed")
-    Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, "mkdir -p #{mount} #{signed}")
+    Pkg::Util::Net.remote_execute(ssh_host_string, "mkdir -p #{mount} #{signed}")
     dmgs = Dir.glob("#{target_dir}/apple/**/*.dmg")
     Pkg::Util::Net.rsync_to(dmgs.join(" "), rsync_host_string, work_dir)
-    Pkg::Util::Net.remote_ssh_cmd(ssh_host_string, %Q[for dmg in #{dmgs.map { |d| File.basename(d, ".dmg") }.join(" ")}; do
+    Pkg::Util::Net.remote_execute(ssh_host_string, %Q[for dmg in #{dmgs.map { |d| File.basename(d, ".dmg") }.join(" ")}; do
```
underscorgan, line 21: I know this isn't directly related to your change, but maybe this command would make more sense in a HEREDOC variable?

tasks/ship.rake
```diff
@@ -547,7 +555,8 @@ namespace :pl do
       local_dir = args.local_dir || 'pkg'
       Dir.glob("#{local_dir}/**/*").reject { |e| File.directory? e }.each do |artifact|
         if File.extname(artifact) == ".yaml" || File.extname(artifact) == ".json"
-          artifactory.deploy_package(artifact)
+          puts "EXGDEBUG::: Passing on #{artifact}"
```
underscorgan, line 558: this doesn't look related to this change

underscorgan (approved):
From my reading of this it looks like this is a backwards compatible change / doesn't change default behavior, is that accurate?

e-gris:
One of the experimental changes, eliminating the Kernel#system / Pkg::Util::Execution#capture3 split in the remote_execute_method broke calls to the gpg CLI. I'll need to put it back.

e-gris:
Made a number of changes for making sure the Kernel#system case was covered.

Tested with puppet-enterprise-vanagon (vangon build & vanagon sign) with pe-java, pe-nginx

tasks/ship.rake
e-gris: Yea, that's not right.

underscorgan (approved):

PR #1100, opened 2021-03-17
Update gem dependencies and cope with the Rubocop results

e-gris:
I got a bee in my bonnet about this being rather out-of-date, so I
updated it. Not sure if it's a worthwhile change but submitted for
discussion.

I updated the dependency versions in the gemspec.

I updated rspec examples from deprecated spec syntax by replacing old 'should' and 'stub' rspec syntax with updated 'expect' and 'allow'

Inside the spec examples, I got more conventional about using single
quotes when making fixed strings.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ AllCops:
     - 'vendor/**/*'
 
 
-Lint/ConditionPosition:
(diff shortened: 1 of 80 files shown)


e-gris:
Needs ruby >= 2.4.0 and a rubocop config update.

PR #1103, opened 2021-04-27
(RE-13941) Add shim to support new apt shipping

e-gris:
(RE-13941) Ship to new puppet-version based apt repos
    
    - Broke some of the :ship tasks into smaller bits and moved them out of Rake.
    
    - Introduced two new tasks, `pl:stage_nightly_debs` and `pl:stage_stable_debs`
      for sending debs to a staging area for repos based on puppet major version.
    
    - The above new tasks use a required set of shims from the `apt_stage_repos` gem.
    
    - Code cleanup for my own readability and positioning for future planned development

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,121 +1,85 @@
+---
+
 AllCops:
   Include:
-    - 'lib/**/*.rb'
-    - 'tasks/**/*'
+  - lib/**/*.rb
+  - tasks/**/*
   Exclude:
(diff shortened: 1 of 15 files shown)


e-gris:
Still need to sort out the specific entry for apt.repos.puppet.com publishing. Right now it is too broad. It expects the published project to be either 'puppet6' or 'puppet7' repo and many (pe-bolt, for example) are not aware of where they go.

PR #1104, opened 2021-04-29
(RE-13647) Refactoring ship code

e-gris:
Currently a WIP. Looking for possible fubars. Main concern is whether the artifact bundle receives multiple updates over multiple calls. I'm freezing it out once the first one occurs.


underscorgan:
> Currently a WIP. Looking for possible fubars. Main concern is whether the artifact bundle receives multiple updates over multiple calls. I'm freezing it out once the first one occurs.

Ah, yeah, at least for vanagon builds the artifact bundle is updated for each platform, since the data is updated in the rake tasks called in `vanagon ship` which executes on each build machine

PR #1116, opened 2021-07-27
(maint) don't upload checksums to artifactory before uploading artifacts

underscorgan:

### DIFF
--- a/lib/packaging/artifactory.rb
+++ b/lib/packaging/artifactory.rb
@@ -224,8 +224,6 @@ module Pkg
         deploy_properties(platform_tag, File.basename(package)),
         headers
       )
-    rescue
-      raise "Attempt to upload '#{package}' to #{File.join(@artifactory_uri, data[:full_artifactory_path])} failed"
     end
 
     # @param pkg [String] The package to download YAML for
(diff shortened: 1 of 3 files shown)


tasks/ship.rake
```diff
@@ -553,7 +553,10 @@ namespace :pl do
       artifactory = Pkg::ManageArtifactory.new(Pkg::Config.project, Pkg::Config.ref)
 
       local_dir = args.local_dir || 'pkg'
-      Dir.glob("#{local_dir}/**/*").reject { |e| File.directory? e }.each do |artifact|
+      files = Dir.glob("#{local_dir}/**/*").reject { |e| File.directory? e }
+      sorted_files = files.sort_by { |file| File.extname(file) =~ /^(md5|sha[d+])$/ ? 1 : 0 }
```
jackie-kinsler, line 557: Do you want to call them `artifacts` and `sorted_artifacts` instead of `files` for consistency w local `artifact` variable?

jackie-kinsler (approved):

lib/packaging/artifactory.rb
```diff
@@ -224,8 +224,9 @@ def deploy_package(package)
         deploy_properties(platform_tag, File.basename(package)),
         headers
       )
-    rescue
-      raise "Attempt to upload '#{package}' to #{File.join(@artifactory_uri, data[:full_artifactory_path])} failed"
+    rescue => e
+      raise e
+      #raise "Attempt to upload '#{package}' to #{File.join(@artifactory_uri, data[:full_artifactory_path])} failed"
```
joshcooper, line 229: Could drop the rescue so you don’t need to reraise?

PR #1132, opened 2021-11-16
(VANAGON-165) Created methods to replace the need to invoke packaging rake tasks by vanagon.

isaac-hammes:


e-gris (commented):
Looks great overall. Rubocop errors need to be cleaned.

Next step AFTER this would be to make the existing rake tasks call the new methods.

PR #1144, opened 2022-02-17
(RE-14537) Add Gemfile.lock for Snyk security scan.

jackie-kinsler:


e-gris (commented):
I prefer we not do this. I think it could break a lot of stuff.

If we can find a way to make Synk run bundle install on its own, that would be better.

PR #1154, opened 2022-04-08
(maint) add nil handling

underscorgan:


lib/packaging/artifactory.rb
```diff
@@ -353,6 +353,7 @@ def download_packages(staging_directory, manifest, remote_path = '')
       check_authorization
       download_repositories = %w[rpm_enterprise__local debian_enterprise__local]
       manifest.each do |dist, packages|
+        next if packages.nil?
```
e-gris, line 356: Maybe include info/warning that the package list was nil for 'dist', so we're skipping it.

e-gris (approved):

PR #1159, opened 2022-04-28
(RE-14611) Change msi sign method to use the gcp msi signer

isaac-hammes:


lib/packaging/sign/msi.rb
```diff
@@ -2,123 +2,95 @@ module Pkg::Sign::Msi
...
+    require 'google/cloud/storage'
+    require 'googleauth'
+    require 'json'
+    require 'net/http'
+    require 'uri'
 
-    ssh_host_string = "#{use_identity} Administrator@#{Pkg::Config.msi_signing_server}"
-    rsync_host_string = "-e 'ssh #{use_identity}' Administrator@#{Pkg::Config.msi_signing_server}"
+    service_account_credentials = '/var/lib/jenkins/.creds/gcp-service-account-credentials-windows-signer.json'
+    service_url = 'https://serverless-sign-4k7em2ejkq-uc.a.run.app/signMSI'
```
underscorgan, line 12: Is this url something that will be stable? If it's something that will change we should see if we can get something more stable that can always just point to the latest version. May want to have this in Pkg::Config / loaded from build_data?

isaac-hammes: I know I mentioned a while back that it wasn't stable, but I was wrong I think. I deleted the service and recreated it in terraform and the url stayed the same. Do you think it is better to just put it in build-data anyways?

underscorgan (approved):

PR #1178, opened 2022-12-13
(RE-15086) Clean up RPM signing

e-gris:
RPM signing works with GPG >=1 now, so stop evading it. In the old case
where /usr/bin/gpg is still gpg1, this will continue to work.

Clean up some organic code drift to isolate specific responsibility


isaac-hammes (approved):

e-gris:
I want to hold on merging this until after current Puppet FOSS ships.

underscorgan (approved):

PR #1193, opened 2023-03-01
(maint) revert call to ERB.new for the sake of Ruby 2.5

e-gris:
Named parameters to ERB.new came in at Ruby 2.6; we're still pegged at
Ruby 2.5 in some places so we need to accommodate them.

### DIFF
--- a/lib/packaging/util/file.rb
+++ b/lib/packaging/util/file.rb
@@ -58,7 +58,14 @@ module Pkg::Util::File
 
     def erb_string(erbfile, b = binding)
       template = File.read(erbfile)
-      message  = ERB.new(template, trim_mode: "-")
+
+      # We should be using this but some images are pegged at Ruby <= 2.5; this interface
+      # changed at Ruby 2.6
+      # message  = ERB.new(template, safe_level: nil, trim_mode: '-')
+
+      # rubocop:disable Lint/ErbNewArguments
+      message = ERB.new(template, nil, '-')
+      # rubocop:enable Lint/ErbNewArguments
       message.result(b)
     end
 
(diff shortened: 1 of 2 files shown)


lib/packaging/util/file.rb
```diff
@@ -58,7 +58,14 @@ def file_writable?(file, args = { :required => false })
...
     def erb_string(erbfile, b = binding)
       template = File.read(erbfile)
-      message  = ERB.new(template, trim_mode: "-")
+
+      # We should be using this but some images are pegged at Ruby <= 2.5; this interface
+      # changed at Ruby 2.6
+      # message  = ERB.new(template, safe_level: nil, trim_mode: '-')
+
+      # rubocop:disable Lint/ErbNewArguments
+      message = ERB.new(template, nil, '-')
```
donoghuc, line 67: Do we need to support both ruby versions? We could try somethin like:
```ruby
message = if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.6.0')
    ERB.new(template, nil, '-')                                    
else                                                         
    ERB.new(template, safe_level: nil, trim_mode: '-')                                
end 
```

jonathannewman (approved):

lib/packaging/util/file.rb
e-gris: We don't just yet. As of 3.2 it remains deprecated but supported. IMO, the Rubocop flag is sufficient.

Also I prefer seeing the deprecation noise, if any, than to accommodate it.

I hope to have the fpm image updated before we need to go further.

isaac-jha (approved):

PR #1199, opened 2023-03-16
(maint) Avoid race conditions pushing gem to rubygems

donoghuc:
In order to avoid re-pushing to rubygems if we have already pushed a version we check the `versions` api. There appears to be a race condition where that API has out-dated infor by the time we go to push. This commit catches the "gem already pushed" error, logs it and will *not* throw it as the intent of the action is already satisfied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+### Changed
+- (maint) Do not throw an error when pushing a gem to rubygems if it already has been pushed.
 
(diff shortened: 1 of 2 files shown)


donoghuc:
I didnt add a test because I would be mocking the response from rubygems anyway. If you want that I can, but it seems like it would provide no value.

e-gris (changes requested):
This is great. Please add an entry into CHANGELOG.md and it'll be good.

donoghuc:
Updated. Thanks so much Eric. Is this enough to push a release?

e-gris (approved):

e-gris:
Yes, I'll do a release shortly.

PR #1213, opened 2023-05-03
(RE-15217, RE-15221) Remove SLES 11 from PLATFORM_INFO list.

isaac-jha:

### DIFF
--- a/spec/lib/packaging/sign_spec.rb
+++ b/spec/lib/packaging/sign_spec.rb
@@ -63,7 +63,7 @@ describe 'Pkg::Sign' do
         %W[#{rpm_directory}/aix/7.1/PC1/ppc/puppet-agent-5.5.3-1.aix7.1.ppc.rpm]
       end
       let(:v3_rpms) do
-        %W[#{rpm_directory}/sles/11/PC1/x86_64/puppet-agent-5.5.3-1.sles11.x86_64.rpm]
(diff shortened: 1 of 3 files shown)


spec/lib/packaging/sign_spec.rb
```diff
@@ -63,7 +63,7 @@
         %W[#{rpm_directory}/aix/7.1/PC1/ppc/puppet-agent-5.5.3-1.aix7.1.ppc.rpm]
       end
       let(:v3_rpms) do
-        %W[#{rpm_directory}/sles/11/PC1/x86_64/puppet-agent-5.5.3-1.sles11.x86_64.rpm]
+        %W[#{rpm_directory}/redhatfips/7/PC1/x86_64/puppet-agent-5.5.3-1.redhatfips7.x86_64.rpm]
```
e-gris, line 66: Is this correct? Does fips still up v3 signatures?

e-gris (approved):
Good riddance SLES 11.

I had one question but still approved regardless of the answer.

spec/lib/packaging/sign_spec.rb
isaac-jha: I think so. Unless I'm mistaken, whether it is a v3 or v4 signature is specified in [lib/packaging/platforms.rb](https://github.com/puppetlabs/packaging/blob/1.0.x/lib/packaging/platforms.rb). It looks like only redhatfips 7 and 8 are v3 outside of sles 11.

PR #1215, opened 2023-05-18
(maint) Make sign command output verbose and specify digest-algo as sha512.

isaac-jha:


isaac-jha:
Something I'm unsure about is whether or not we should look into updating the v3 signing.

e-gris (approved):
Approval contingent on merging after PE patch release.

PR #1232, opened 2023-11-29
(maint) add support for Debian 12 bookworm

h0tw1r3:


e-gris (commented):
While I'm fairly certain that this will be fine, please confirm that you've tried one or two examples to confirm that it works.

h0tw1r3:
@e-gris confirmed. Built puppet-agent, bolt, pdk, and client-tools.

e-gris (approved):

PR #1237, opened 2024-01-16
(PA-5514) Update puppetlabs/packaging platform hash to include Fedora…

skyamgarp:
… 38 x86_64

(PA-5514) Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+### Added
+- (PA-5514) Add support for Fedora 38 x86_64
 
 ## [0.113.0] - 2024-01-08
 ### Added
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -109,6 +109,14 @@ module Pkg
           signature_format: 'v4',
           repo: true,
         },
+        '38' => {
+          architectures: ['x86_64'],
+          source_architecture: 'SRPMS',
+          package_format: 'rpm',
+          source_package_formats: ['src.rpm'],
+          signature_format: 'v4',
+          repo: true,
+        },
       },
 
       'osx' => {


tasks/00_utils.rake
```diff
@@ -103,9 +103,9 @@ def print_url_info(url_string)
   Pkg::Util::Net.print_url_info(url_string)
 end
 
-def retry_on_fail(args, &block)
+def retry_on_fail(...)
   Pkg::Util.deprecate("retry_on_fail", "Pkg::Util::Execution.retry_on_fail")
-  Pkg::Util::Execution.retry_on_fail(args, &block)
+  Pkg::Util::Execution.retry_on_fail(...)
```
e-gris: There may still be things in our creaky infrastructure using Ruby 2.5, which would cause this to fail.

While I feel its a good change, it does have the possibility of backfiring. 

I'm going to approve the PR but it may require a quick revert if something trips over it.

e-gris (approved):

tasks/00_utils.rake
skyamgarp: Do we want to ignore rubocop in that case?
What do you think about adding required_ruby_version as 2.7/3 in gemspec same as TargetRubyVersion in rubocop.yml? Or vice versa

e-gris: I think we should disable this particular cop for this case.

`packaging` is a terrible beast that needs always to serve the lowest common denominator. This results in holding it back in sometimes comical ways. It would be a good test to require ruby 2.7 to observe what, if anything, breaks as a result. Then be prepared to fix those things or roll back packaging as needed.

tasks/00_utils.rake
```diff
@@ -103,10 +103,14 @@ def print_url_info(url_string)
   Pkg::Util::Net.print_url_info(url_string)
 end
 
+# rubocop:disable Lint/RedundantCopDisableDirective
```
skyamgarp: To avoid below offense have to disable `RedundantCopDisableDirective` too.

```
Offenses:

tasks/00_utils.rake:106:1: W: [Correctable] Lint/RedundantCopDisableDirective: Unnecessary disabling of Style/ArgumentsForwarding.
# rubocop:disable Style/ArgumentsForwarding
```

skyamgarp:
@e-gris Please let me know if we are good to merge this PR now. Thanks.

tasks/00_utils.rake
e-gris: I strongly encourage Rubocop disables be placed in the .rubocop.yml file rather than as source code commentary in order to make consistent policy and to keep the source code tidy.

e-gris: There _may_ be the rare exception where we want to temporarily disable an otherwise desirable but this, I feel, isn't one of them.

PR #1244, opened 2024-02-16
Export GEM_SOURCE to remote machine

mhashizume:
Previously, when executing a command on a remote machine, packaging exported some environmental variables such as PACKAGING_LOCATION and VANAGON_LOCATION which allowed commands run on a remote machine to maintain the same environment.

However, GEM_SOURCE was not among those exported environmental variables. If a project was using a gem not found on Rubygems--in my case, pl_fustigit hosted on Artifactory--that gem could be installed locally by using the source defined in GEM_SOURCE, but not remotely.

This commit checks whether the GEM_SOURCE environmental variable is defined and, if it is and isn't empty, exports it to a remote machine with other environmental variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Changed
+- Pass GEM_SOURCE environmental variable to remote hosts.
+
 ## [0.115.0] - 2024-02-13
 ### Changed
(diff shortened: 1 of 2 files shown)


e-gris (changes requested):
FWIW, pl_fustigit on Artifactory is there only for a quick test. A full release to rubygems I'll likely do next week. 

Nevertheless, this is a fine change. My only request is an update to the CHANGELOG

mhashizume:
I've added my changes to the changelog

e-gris (approved):

PR #1245, opened 2024-02-22
(PA-6063) Update puppetlabs/packaging platform hash to include Red Ha…

imaqsood:
…t 9 FIPS (Intel)

### DIFF
--- a/lib/packaging/platforms.rb
+++ b/lib/packaging/platforms.rb
@@ -137,6 +137,14 @@ module Pkg
           source_package_formats: ['src.rpm'],
           signature_format: 'v3',
           repo: true,
+        },
+        '9' => {
+          architectures: ['x86_64'],
+          source_architecture: 'SRPMS',
+          package_format: 'rpm',
+          source_package_formats: ['src.rpm'],
+          signature_format: 'v4',
+          repo: true,
         }
       },
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com/).
 
 ## [Unreleased]
+### Added
+- (PA-6063) Added platform for Red Hat 9 FIPS for x86_64
 
 ## [0.112.0]
 ### Added


lib/packaging/platforms.rb
```diff
@@ -137,6 +137,14 @@ module Platforms
           source_package_formats: ['src.rpm'],
           signature_format: 'v3',
           repo: true,
+        },
+        '9' => {
+          architectures: ['x86_64'],
+          source_architecture: 'SRPMS',
+          package_format: 'rpm',
+          source_package_formats: ['src.rpm'],
+          signature_format: 'v3',
```
joshcooper: Is v3 required for FIPS? Or should we use v4?

joshcooper:
Could you add an entry to the Unreleased section like https://github.com/puppetlabs/packaging/commit/fc555c844a795319e15edc94062eadbbb1117759#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed It helps later when we go to release the packaging gem.

lib/packaging/platforms.rb
imaqsood: From my research on various online sources, RPM version 4 appears to be the preferable choice over version 3 for Red Hat 9 FIPS compliance, unless there are specific reasons necessitating the use of RPM version 3. One significant disparity lies in the immutable header region in Signature and Header. RPM version 4 employs header-only signatures and hashes on this immutable region, with file paths stored in a "compressed" format (dirname, dirindex, basename).

joshcooper: Ah thanks @imaqsood!

@e-gris just wanted to confirm it's ok to use v4 for RHEL9 FIPS? For some reason RHEL 7 & 8 FIPS are still using v3, while non-FIPS is using v4.

e-gris: v4 seems right to me but we can always change back to v3 if there are any issues.

Conservatively, v4 should work (without breaking changes) with any rpm version after 4.14.0 (released Oct 12 2017)

It'd be a fun experiment if we can move everything in packaging to v4. Unfortunately, I don't see time for me to do so any time soon.

imaqsood: Using v4 I build puppet-agent and it ran successfully 
[Build](https://jenkins-platform.delivery.puppetlabs.net/job/platform_vanagon-generic-builder_vanagon-packaging_generic-builder/BUILD_TARGET=redhatfips-9-x86_64,SLAVE_LABEL=k8s-worker/2678/)
[artifacts](https://builds.delivery.puppetlabs.net/puppet-agent/16f8f8232d419a71115ba7f1739ce82d507df463/artifacts/)

joshcooper (approved):

e-gris (approved):

PR #1249, opened 2024-03-25
(PE-33254) Move Solaris signing to use GPG .asc files

e-gris:


tasks/sign.rake
```diff
@@ -17,22 +17,22 @@ namespace :pl do
...
-    unless packages.empty?
-      Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
-      packages.each do |swix_package|
-        Pkg::Util::Gpg.sign_file swix_package
-      end
+    next if packages.empty?
+
+    Pkg::Util::Gpg.load_keychain if Pkg::Util::Tool.find_tool('keychain')
+    packages.each do |swix_package|
+      Pkg::Util::Gpg.sign_file swix_package
```
shaigy, line 24: Should the Gpg.sign_file also be within the if conditional like load_keychain?

e-gris: It shouldn't. keychain is a device only on debian systems. This conditional is saying 'if we have the "keychain" utility use it'. 

However, it's not available on rpm-based systems.

The whole Pkg::Util::Gpg module could stand to be less coy.

shaigy (approved):

PR #1277, opened 2026-07-13
(maint) Accept ARTIFACTORY_ACCESS_TOKEN in check_authorization

isaac-jha:
## Summary

- `check_authorization` in `lib/packaging/artifactory.rb` now accepts `ARTIFACTORY_ACCESS_TOKEN` alongside the existing `ARTIFACTORY_USERNAME`/`ARTIFACTORY_PASSWORD` and `ARTIFACTORY_API_KEY` variants.
- Error message updated to guide callers toward `ARTIFACTORY_ACCESS_TOKEN` and note that `ARTIFACTORY_API_KEY` is deprecated by JFrog.
- Spec extended with positive tests for both the new access-token path and the existing (deprecated) api-key path, with an `around` hook that isolates env-var state per example.

## Why

JFrog deprecated API-key authentication and blocked new API-key creation starting with **Artifactory 7.98**. The bundled `puppet-artifactory-client` SDK already accepts an access token via `ARTIFACTORY_ACCESS_TOKEN` (see `lib/artifactory/defaults.rb` / `lib/artifactory/client.rb` in that repo — the auth precedence there is username+password → access_token → api_key). But this gem's `check_authorization` only recognized the old set, so any caller that migrated to a scoped access token or reference token would hit the "Unable to determine credentials" error even though the underlying SDK would authenticate fine.

This is a prerequisite for the on-prem Artifactory upgrade to 7.146.x — we need callers to be able to run against a new access token without changing this gem's semantics for existing callers.

## Test plan

- [x] `bundle exec rspec spec/lib/packaging/artifactory_spec.rb` — 66 examples, 0 failures locally.
- [ ] CI passes.

## Notes

- `ARTIFACTORY_API_KEY` remains fully supported for backwards compatibility until existing callers migrate. No behavior change for them.
- Priority order between the three auth surfaces is delegated to the SDK, which currently prefers basic auth → access token → api key.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This changelog adheres to [Keep a CHANGELOG](http://keepachangelog.com).
 
 ## [Unreleased]
 ### Added
+- (maint) Accept `ARTIFACTORY_ACCESS_TOKEN` in `check_authorization` so callers can authenticate
+  with a scoped access token or reference token. `ARTIFACTORY_API_KEY` is still accepted for
+  backwards compatibility, but JFrog has deprecated API-key auth (blocked new keys in Artifactory
+  7.98) and callers should migrate.
 - (maint) Provide a mechanism, by way of a new environment variable `PACKAGING_GITREF_REPLACEMENT`,
   which allows us to not require a release tag before building product.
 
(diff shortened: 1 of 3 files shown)


isaac-jha:
Closing — team no longer uses this repo; equivalent fix will land in packaging-private only.

